PR #35, opened 2026-04-23
Use persisted monotonic counter for nogood IDs

### LINKED ISSUE #26: Nogood IDs assume append-only list, deletion causes collisions
## Problem

In `network.py`, `add_nogood` derives nogood IDs from `len(self.nogoods) + 1`. If a nogood is ever deleted from the list, subsequent calls will generate IDs that collide with existing entries, corrupting the contradiction history.

## Impact

This root defect gates **3 derived beliefs** in the knowledge base, including:
- `nogood-resolution-maintains-consistent-ids`
- `belief-revision-is-fully-reliable`

## Suggested fix

Use a monotonically increasing counter (e.g., `self._next_nogood_id`) that is persisted alongside the nogoods, or use UUIDs, so that IDs remain unique regardless of deletions.

## Source

Belief: `nogood-ids-assume-append-only`
Entry: `entries/2026/04/23/reasons_lib-network-add_nogood.md`

benthomasson:
## Summary

- Replaces `len(self.nogoods) + 1` with a `_next_nogood_id` counter that only increments, ensuring deleted nogood IDs are never reused
- Persists the counter in a new `network_meta` SQLite table
- Updates the counter correctly when importing nogoods from JSON or beliefs.md
- Gracefully handles old databases that lack the `network_meta` table

Closes #26

## Test plan

- [x] 8 new tests covering monotonic IDs, persistence round-trips, old DB compat, and both import paths
- [x] Full test suite passes (425 tests)

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -922,6 +922,7 @@ def import_json(json_file: str, db_path: str = DEFAULT_DB) -> dict:
             remaining = next_remaining
 
         # Import nogoods
+        import re
         from . import Nogood
         nogoods_imported = 0
         for ng_data in data.get("nogoods", []):
@@ -932,6 +933,9 @@ def import_json(json_file: str, db_path: str = DEFAULT_DB) -> dict:
                 resolution=ng_data.get("resolution", ""),
             )
             net.nogoods.append(nogood)
+            m = re.fullmatch(r"nogood-(\d+)", nogood.id)
+            if m:
+                net._next_nogood_id = max(net._next_nogood_id, int(m.group(1)) + 1)
             nogoods_imported += 1
 
         # Import repos
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -7,6 +7,7 @@ or argparse. Each function opens the database, operates, saves, and closes.
 All functions return dicts suitable for JSON serialization.
 """
 
+import re
 from pathlib import Path
 
 from . import Justification
@@ -922,7 +923,7 @@ def import_json(json_file: str, db_path: str = DEFAULT_DB) -> dict:
             remaining = next_remaining
 
         # Import nogoods
-        import re
+
         from . import Nogood
         nogoods_imported = 0
         for ng_data in data.get("nogoods", []):
--- a/reasons_lib/import_beliefs.py
+++ b/reasons_lib/import_beliefs.py
@@ -239,6 +239,11 @@ def import_into_network(
                     resolution=ng["resolution"],
                 )
                 network.nogoods.append(nogood)
+                m = re.fullmatch(r"nogood-(\d+)", nogood.id)
+                if m:
+                    network._next_nogood_id = max(
+                        network._next_nogood_id, int(m.group(1)) + 1
+                    )
                 nogoods_imported += 1
 
     return {
--- a/reasons_lib/network.py
+++ b/reasons_lib/network.py
@@ -20,6 +20,7 @@ class Network:
     def __init__(self):
         self.nodes: dict[str, Node] = {}
         self.nogoods: list[Nogood] = []
+        self._next_nogood_id: int = 1
         self.repos: dict[str, str] = {}  # name → path mapping
         self.log: list[dict] = []  # propagation audit trail
 
@@ -299,7 +300,8 @@ class Network:
             if nid not in self.nodes:
                 raise KeyError(f"Node '{nid}' not found")
 
-        nogood_id = f"nogood-{len(self.nogoods) + 1:03d}"
+        nogood_id = f"nogood-{self._next_nogood_id:03d}"
+        self._next_nogood_id += 1
         nogood = Nogood(
             id=nogood_id,
             nodes=list(node_ids),
--- a/reasons_lib/storage.py
+++ b/reasons_lib/storage.py
@@ -52,6 +52,11 @@ CREATE TABLE IF NOT EXISTS propagation_log (
     value TEXT NOT NULL
 );
 
+CREATE TABLE IF NOT EXISTS network_meta (
+    key TEXT PRIMARY KEY,
+    value TEXT NOT NULL
+);
+
 CREATE VIRTUAL TABLE IF NOT EXISTS nodes_fts USING fts5(id, text);
 """
 
@@ -80,6 +85,7 @@ class Storage:
             self.conn.execute("DELETE FROM nogoods")
             self.conn.execute("DELETE FROM repos")
             self.conn.execute("DELETE FROM propagation_log")
+            self.conn.execute("DELETE FROM network_meta")
 
             for node in network.nodes.values():
                 self.conn.execute(
@@ -112,6 +118,11 @@ class Storage:
                     "VALUES (?, ?, ?, ?)",
                     (nogood.id, json.dumps(nogood.nodes), nogood.discovered, nogood.resolution),
                 )
+
+            self.conn.execute(
+                "INSERT INTO network_meta (key, value) VALUES (?, ?)",
+                ("next_nogood_id", str(network._next_nogood_id)),
+            )
 
             for name, path in network.repos.items():
                 self.conn.execute(
@@ -179,6 +190,16 @@ class Storage:
                 discovered=discovered,
                 resolution=resolution,
             ))
+
+        # Load network metadata
+        try:
+            row = self.conn.execute(
+                "SELECT value FROM network_meta WHERE key = 'next_nogood_id'"
+            ).fetchone()
+            if row:
+                network._next_nogood_id = int(row[0])
+        except Exception:
+            pass  # network_meta table may not exist in old databases
 
         # Load repos
         try:
--- a/reasons_lib/storage.py
+++ b/reasons_lib/storage.py
@@ -198,6 +198,15 @@ class Storage:
             ).fetchone()
             if row:
                 network._next_nogood_id = int(row[0])
+            elif network.nogoods:
+                # Old database without persisted counter — derive from existing nogoods
+                import re
+                max_id = 0
+                for ng in network.nogoods:
+                    m = re.fullmatch(r"nogood-(\d+)", ng.id)
+                    if m:
+                        max_id = max(max_id, int(m.group(1)))
+                network._next_nogood_id = max_id + 1
         except Exception:
             pass  # network_meta table may not exist in old databases
 
--- a/reasons_lib/storage.py
+++ b/reasons_lib/storage.py
@@ -191,24 +191,26 @@ class Storage:
                 resolution=resolution,
             ))
 
-        # Load network metadata
+        # Load network metadata — persisted counter takes priority,
+        # otherwise derive from existing nogoods to avoid ID collisions
+        loaded_counter = False
         try:
             row = self.conn.execute(
                 "SELECT value FROM network_meta WHERE key = 'next_nogood_id'"
             ).fetchone()
             if row:
                 network._next_nogood_id = int(row[0])
-            elif network.nogoods:
-                # Old database without persisted counter — derive from existing nogoods
-                import re
-                max_id = 0
-                for ng in network.nogoods:
-                    m = re.fullmatch(r"nogood-(\d+)", ng.id)
-                    if m:
-                        max_id = max(max_id, int(m.group(1)))
-                network._next_nogood_id = max_id + 1
+                loaded_counter = True
         except Exception:
             pass  # network_meta table may not exist in old databases
+        if not loaded_counter and network.nogoods:
+            import re
+            max_id = 0
+            for ng in network.nogoods:
+                m = re.fullmatch(r"nogood-(\d+)", ng.id)
+                if m:
+                    max_id = max(max_id, int(m.group(1)))
+            network._next_nogood_id = max_id + 1
 
         # Load repos
         try:
--- a/tests/test_nogood_id.py
+++ b/tests/test_nogood_id.py
@@ -114,6 +114,30 @@ class TestPersistence:
         assert loaded._next_nogood_id == 1
         storage2.close()
 
+    def test_old_db_with_nogoods_derives_counter(self, tmp_path):
+        """Old DB with existing nogoods must derive counter, not start at 1."""
+        db_path = str(tmp_path / "reasons.db")
+        storage = Storage(db_path)
+        net = Network()
+        net.add_node("a", "A")
+        net.add_node("b", "B")
+        net.add_node("c", "C")
+        net.add_node("d", "D")
+        net.add_nogood(["a", "b"])
+        net.add_nogood(["c", "d"])
+        storage.save(net)
+        # Remove the meta row to simulate an old database
+        storage.conn.execute("DELETE FROM network_meta")
+        storage.conn.commit()
+        storage.close()
+
+        storage2 = Storage(db_path)
+        loaded = storage2.load()
+        assert loaded._next_nogood_id == 3
+        loaded.add_nogood(["a", "c"])
+        assert loaded.nogoods[-1].id == "nogood-003"
+        storage2.close()
+
 
 class TestImportJson:
 
--- a/tests/test_nogood_id.py
+++ b/tests/test_nogood_id.py
@@ -126,8 +126,8 @@ class TestPersistence:
         net.add_nogood(["a", "b"])
         net.add_nogood(["c", "d"])
         storage.save(net)
-        # Remove the meta row to simulate an old database
-        storage.conn.execute("DELETE FROM network_meta")
+        # Drop the table entirely to simulate an old database
+        storage.conn.execute("DROP TABLE IF EXISTS network_meta")
         storage.conn.commit()
         storage.close()
 
@@ -136,6 +136,8 @@ class TestPersistence:
         assert loaded._next_nogood_id == 3
         loaded.add_nogood(["a", "c"])
         assert loaded.nogoods[-1].id == "nogood-003"
+        # Verify save works after upgrade
+        storage2.save(loaded)
         storage2.close()
 
 
--- a/tests/test_nogood_id.py
+++ b/tests/test_nogood_id.py
@@ -1,0 +1,164 @@
+"""Tests for monotonic nogood ID generation (issue #26).
+
+The bug: nogood IDs were derived from len(self.nogoods) + 1, so deleting
+a nogood and adding a new one would reuse the deleted ID. The fix uses a
+persisted monotonic counter (_next_nogood_id) that never decreases.
+"""
+
+import json
+
+import pytest
+
+from reasons_lib import api, Nogood
+from reasons_lib.network import Network
+from reasons_lib.storage import Storage
+
+
+@pytest.fixture
+def db(tmp_path):
+    db_path = str(tmp_path / "reasons.db")
+    api.init_db(db_path=db_path)
+    return db_path
+
+
+class TestMonotonicCounter:
+
+    def test_ids_increment(self):
+        net = Network()
+        net.add_node("a", "A")
+        net.add_node("b", "B")
+        net.add_node("c", "C")
+        net.add_node("d", "D")
+        net.add_nogood(["a", "b"])
+        net.add_nogood(["c", "d"])
+        assert net.nogoods[0].id == "nogood-001"
+        assert net.nogoods[1].id == "nogood-002"
+        assert net._next_nogood_id == 3
+
+    def test_delete_last_does_not_reuse_id(self):
+        net = Network()
+        net.add_node("a", "A")
+        net.add_node("b", "B")
+        net.add_node("c", "C")
+        net.add_node("d", "D")
+        net.add_nogood(["a", "b"])
+        net.add_nogood(["c", "d"])
+        del net.nogoods[-1]
+        net.add_nogood(["a", "c"])
+        assert net.nogoods[-1].id == "nogood-003"
+
+    def test_delete_all_does_not_reset(self):
+        net = Network()
+        net.add_node("a", "A")
+        net.add_node("b", "B")
+        net.add_nogood(["a", "b"])
+        net.nogoods.clear()
+        assert net._next_nogood_id == 2
+        net.add_nogood(["a", "b"])
+        assert net.nogoods[0].id == "nogood-002"
+
+
+class TestPersistence:
+
+    def test_counter_survives_save_load(self, tmp_path):
+        db_path = str(tmp_path / "reasons.db")
+        storage = Storage(db_path)
+        net = Network()
+        net.add_node("a", "A")
+        net.add_node("b", "B")
+        net.add_nogood(["a", "b"])
+        assert net._next_nogood_id == 2
+        storage.save(net)
+        storage.close()
+
+        storage2 = Storage(db_path)
+        loaded = storage2.load()
+        assert loaded._next_nogood_id == 2
+        storage2.close()
+
+    def test_counter_persists_after_deletion(self, tmp_path):
+        db_path = str(tmp_path / "reasons.db")
+        storage = Storage(db_path)
+        net = Network()
+        net.add_node("a", "A")
+        net.add_node("b", "B")
+        net.add_node("c", "C")
+        net.add_node("d", "D")
+        net.add_nogood(["a", "b"])
+        net.add_nogood(["c", "d"])
+        del net.nogoods[-1]
+        storage.save(net)
+        storage.close()
+
+        storage2 = Storage(db_path)
+        loaded = storage2.load()
+        assert loaded._next_nogood_id == 3
+        loaded.add_nogood(["a", "c"])
+        assert loaded.nogoods[-1].id == "nogood-003"
+        storage2.close()
+
+    def test_old_db_without_meta_table(self, tmp_path):
+        """Databases created before this fix should still load."""
+        db_path = str(tmp_path / "reasons.db")
+        storage = Storage(db_path)
+        net = Network()
+        net.add_node("a", "A")
+        storage.save(net)
+        # Drop the meta table to simulate an old database
+        storage.conn.execute("DROP TABLE IF EXISTS network_meta")
+        storage.conn.commit()
+        storage.close()
+
+        storage2 = Storage(db_path)
+        loaded = storage2.load()
+        assert loaded._next_nogood_id == 1
+        storage2.close()
+
+
+class TestImportJson:
+
+    def test_import_json_updates_counter(self, db, tmp_path):
+        api.add_node("a", "A", db_path=db)
+        api.add_node("b", "B", db_path=db)
+
+        json_data = {
+            "nodes": {},
+            "nogoods": [
+                {"id": "nogood-005", "nodes": ["a", "b"], "discovered": "", "resolution": ""},
+            ],
+            "repos": {},
+        }
+        json_file = str(tmp_path / "import.json")
+        with open(json_file, "w") as f:
+            json.dump(json_data, f)
+
+        api.import_json(json_file, db_path=db)
+        result = api.add_nogood(["a", "b"], db_path=db)
+        assert result["nogood_id"] == "nogood-006"
+
+
+class TestImportBeliefs:
+
+    def test_import_nogoods_updates_counter(self, db, tmp_path):
+        api.add_node("a", "A", db_path=db)
+        api.add_node("b", "B", db_path=db)
+
+        beliefs_text = ""
+        nogoods_text = """# Nogoods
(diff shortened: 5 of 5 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#35
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### `reasons_lib/network.py:Network.__init__` + `Network.add_nogood`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The core fix is clean. `_next_nogood_id` replaces `len(self.nogoods) + 1`, and is incremented after each allocation. The counter is initialized to 1 and monotonically increases. The `:03d` format is preserved. This directly fixes the collision described in the issue.

### `reasons_lib/storage.py` (schema + save + load)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `network_meta` table and save/load logic are correct for new databases. However, the **backward compatibility fallback has a bug**: when loading an old database that has nogoods but no `network_meta` row, the counter stays at `1` (the `Network.__init__` default). Since `_init_schema()` uses `CREATE TABLE IF NOT EXISTS`, opening an old DB will create an empty `network_meta` table, the `SELECT` will return `None`, and `_next_nogood_id` remains 1. The next `add_nogood` call would produce `nogood-001`, colliding with existing nogoods. The fix should derive the counter from loaded nogoods when the meta key is absent — e.g., by scanning `network.nogoods` after loading them and setting `_next_nogood_id = max(parsed_ids) + 1`. The test `test_old_db_without_meta_table` doesn't catch this because it tests an empty network with no nogoods.

### `reasons_lib/api.py:import_json`
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The counter-update logic (`re.fullmatch` + `max`) is correct and handles non-matching IDs gracefully. Minor issue: `import re` is placed inside the function body at line 925. While functional, `re` is a stdlib module — importing it at the top of the file would be more conventional. Not a blocker. The `max` approach correctly handles importing multiple nogoods and non-sequential IDs.

### `reasons_lib/import_beliefs.py:import_into_network`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same counter-update pattern as `import_json`. The `re` module is already imported at the top of this file (line 14), so no redundant import. Logic correctly uses `max` to handle out-of-order imports.

### `tests/test_nogood_id.py`
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Good test structure covering the core scenarios: monotonic increment, delete-doesn't-reuse, clear-doesn't-reset, save/load round-trip, and both import paths. However, there's a **missing test for the backward-compat edge case**: old DB with existing nogoods but no `network_meta` table. `test_old_db_without_meta_table` only tests an empty network (no nogoods), so it doesn't expose the collision bug described above. A test that saves a network with nogoods, drops `network_meta`, reloads, and verifies the counter is derived from existing nogoods would be needed.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Could not check whether any existing tests in `test_storage.py` or `test_backtracking.py` exercise the nogood path and might need updates. Did not verify the full `import_json` function flow end-to-end (only read the relevant section).

### Feature Requests
- Include the full function body for modified functions in the observations section, not just diff hunks — verifying backward-compat paths requires seeing the complete load() flow
- Show callers of modified methods (e.g., who calls `add_nogood` or `load`) to verify all integration points are covered
- Automatically flag backward-compatibility fallback paths (`except: pass`, default values for missing data) as high-priority review targets

## gemini [BLOCK]

### `reasons_lib/storage.py`
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The fix introduces two regressions for legacy databases created before this patch:
1. **Crash on Save:** `save()` executes `self.conn.execute("DELETE FROM network_meta")` and an `INSERT` statement without verifying that the `network_meta` table exists. If a user loads an existing database and calls `save()`, this will raise an `sqlite3.OperationalError` and abort the save. `save()` must ensure the table is created before modifying it.
2. **ID Collisions:** In `load()`, if reading from `network_meta` fails, `network._next_nogood_id` falls back to `1`. If the loaded database already has `nogoods` (e.g., `nogood-001`, `nogood-002`), the next `add_nogood` call will reuse `nogood-001` and collide—triggering the exact bug this PR was meant to fix. If the metadata read fails, `load()` must infer the next ID by parsing the IDs of the `nogoods` just loaded.

### `reasons_lib/network.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `Network` class properly initializes, maintains, and increments `_next_nogood_id` during `add_nogood`, correctly addressing the core issue of ID collisions caused by deleted nogoods.

### `reasons_lib/api.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly extracts numeric IDs from imported JSON nogoods and advances the network's internal counter, preventing overlaps between imported data and new additions.

### `reasons_lib/import_beliefs.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Matches the logic in `api.py`; correctly extracts IDs from imported Markdown beliefs and ensures `_next_nogood_id` stays monotonically increasing.

### `tests/test_nogood_id.py`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `test_old_db_without_meta_table` test unintentionally masks the bugs in `load()` and `save()`. It asserts that `loaded._next_nogood_id == 1` after dropping the meta table, but if there were existing nogoods, an ID of 1 would cause collisions. A robust test should: 1) add a nogood, 2) drop the meta table, 3) load the DB and add another nogood to ensure no collision occurs, and 4) call `save()` to verify it doesn't throw an `OperationalError`.

### Self-Review
**Limitations:** I could not see the `Storage.__init__` implementation to verify if it automatically applies the schema string for old databases. However, the presence of the `except Exception` block in `load()` clearly implies the author assumes the table might be missing, meaning `save()` is unequivocally unprotected against an absence of the `network_meta` table.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#35
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## Disagreements

- **tests/test_nogood_id.py** [LOW]: claude: PASS, gemini: CONCERN

## claude [CONCERN]

### reasons_lib/network.py:Network.__init__
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `_next_nogood_id: int = 1` counter is initialized correctly. It starts at 1, which matches the old `len(self.nogoods) + 1` behavior for fresh networks. No observable difference for callers unless they delete nogoods.

### reasons_lib/network.py:Network.add_nogood
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Core fix is minimal and correct. The counter is incremented unconditionally after generating the ID, so even if `add_nogood` partially fails after this point (e.g., during backtracking), the ID is still consumed — this is the right choice since partial failures shouldn't reclaim IDs. The `:03d` formatting is preserved from the original code.

### reasons_lib/storage.py:Storage.save
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `network_meta` table is cleared and rewritten during each save, consistent with the existing "clear and rewrite" strategy used for all other tables. The counter is persisted as a string in a key-value store, which is simple and adequate.

### reasons_lib/storage.py:Storage.load (metadata section)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The backward compatibility logic is functionally correct but the `except Exception: pass` at line 210 is overly broad. If the `network_meta` table exists but the query fails for a different reason (e.g., disk corruption, `int()` conversion failure on a malformed value), the error is silently swallowed and the counter defaults to 1. This could cause the exact ID collision the fix is meant to prevent. A narrower catch (e.g., `except sqlite3.OperationalError`) would be safer — that's the specific error when a table doesn't exist. The `int(row[0])` conversion could also raise `ValueError` on bad data; that should probably propagate rather than being silently ignored. **However**, this pattern matches the existing `except Exception: pass` used for repos loading at line 219, so it's at least consistent with the existing codebase style.

### reasons_lib/storage.py:SCHEMA (network_meta table)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `CREATE TABLE IF NOT EXISTS network_meta` is idempotent (uses `IF NOT EXISTS`), so it's safe for existing databases — they'll get the table created on first `Storage()` instantiation. This means the "old database without meta table" backward compat path in `load()` will only be hit if an old database was never opened by new code before loading. The schema is simple and appropriate for a key-value metadata store.

### reasons_lib/api.py:import_json (nogood counter update)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Logic is correct — uses `max()` so importing multiple nogoods with non-sequential IDs sets the counter to max+1. The `import re` is placed inside the function body rather than at module top. This is inconsistent with `import_beliefs.py` which imports `re` at the top level, but matches the existing pattern in this function (the `from . import Nogood` on line 926 was already a local import). Minor style inconsistency, not a correctness issue.

### reasons_lib/import_beliefs.py:import_into_network (nogood counter update)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same `max()` logic as `api.py`. Uses `re` from the module-level import. Correctly placed inside the loop so each imported nogood updates the counter.

### tests/test_nogood_id.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test coverage with 9 tests across 4 classes. Tests cover: basic incrementing, delete-last, delete-all, save/load round-trip, delete+persist, old DB without meta table, old DB with nogoods (counter derivation), JSON import, and beliefs import. The backward compat tests are particularly well-designed — `test_old_db_without_meta_table` drops the table and `test_old_db_with_nogoods_derives_counter` deletes only the meta row. Both are realistic migration scenarios. One note: `test_old_db_without_meta_table` (line 100) won't actually test the `except Exception` path because `_init_schema()` in the new `Storage()` constructor will re-create the `network_meta` table. The test still works because the table will be empty (row was never inserted), hitting the `elif network.nogoods` branch — but since there are no nogoods, it falls through to the default `_next_nogood_id = 1`. This is fine because it tests the "empty meta table" case rather than the "missing meta table" case, but the docstring claims it tests "Databases created before this fix should still load" which is slightly misleading.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Could not verify whether any existing tests in `test_storage.py` or `test_backtracking.py` exercise `add_nogood` and might be affected by the ID format change (they shouldn't be, since the format `nogood-NNN` is unchanged, only the counter source changed).

### Feature Requests
- Include existing test files that test modified functions (e.g., `test_storage.py`, `test_api.py`) to verify no existing assertions break
- Show callers of modified functions to verify all paths that create nogoods are updated

## gemini [BLOCK]

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Safely handles updating the counter when importing nogoods from JSON. It correctly derives the max ID without resetting the counter downwards.

### reasons_lib/import_beliefs.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly ensures the monotonic counter is updated when importing nogoods from a markdown file.

### reasons_lib/network.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly implements the monotonic counter `_next_nogood_id` and securely increments it, ensuring that IDs are never reused even after deletions.

### reasons_lib/storage.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The backward-compatibility logic for old databases in `load()` is structurally broken. If the `network_meta` table genuinely does not exist, `self.conn.execute` raises an exception and execution immediately jumps to `except Exception: pass`. This entirely skips the `elif network.nogoods:` block meant to derive the max ID, silently leaving the counter at 1 and causing ID collisions. 

Additionally, `save()` will crash on an old database if the table is missing because it calls `self.conn.execute("DELETE FROM network_meta")` outside any try/except block. If `Storage.__init__` guarantees the table is always created, the exception logic in `load()` is misleading and hides actual DB errors. The derivation logic must execute reliably if the persisted value is missing, regardless of whether the query failed or returned no rows.

### tests/test_nogood_id.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The tests mask the missing table bug in `storage.py`. `test_old_db_without_meta_table` does not add any nogoods before saving, so it expects an ID of 1 whether derivation ran or not. Furthermore, `test_old_db_with_nogoods_derives_counter` uses `DELETE FROM network_meta` instead of dropping the table. This means it only tests the path where the table exists but is empty, entirely missing the exception branch where derivation is skipped.

### Self-Review
**Limitations:** I could not see the complete implementation of `Storage.__init__` to confirm if the sqlite schema is strictly initialized on connection or only executed when `init_db()` is explicitly called. However, the inconsistent behavior between `save()` (crashing if no table) and `load()` (swallowing the exception and skipping derivation) stands as a definitive issue either way.

### Feature Requests
- Provide the full source context for modified classes (e.g., constructors) instead of only modified functions, so lifecycle behaviors like database migrations/schema execution can be accurately verified.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#35
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### `reasons_lib/network.py:Network.__init__`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adding `_next_nogood_id: int = 1` is the minimal necessary state for the counter. Initialized to 1 so the first nogood gets ID `nogood-001`, matching the existing convention.

### `reasons_lib/network.py:add_nogood`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Core fix — replaces `len(self.nogoods) + 1` with `self._next_nogood_id` and increments after use. The counter monotonically increases regardless of list mutations. The increment happens unconditionally after ID assignment, so even if `nogoods.append` were to fail later, the counter still advances — this is the safe direction (wastes an ID rather than reusing one).

### `reasons_lib/storage.py` (schema)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New `network_meta` table with `key TEXT PRIMARY KEY, value TEXT NOT NULL` is a clean, generic design. Using `CREATE TABLE IF NOT EXISTS` ensures it's created alongside other tables in `_init_schema()` (line 70 runs `executescript(SCHEMA)`), including on old databases that are opened with the new code. This means the table will always exist when `load()` runs.

### `reasons_lib/storage.py:save`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clears `network_meta` during the full rewrite (consistent with the existing clear-and-rewrite pattern for other tables), then inserts the counter. The save runs inside a `with self.conn:` block so it's transactional — the counter and nogoods are always consistent on disk.

### `reasons_lib/storage.py:load`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The backward-compatibility logic has a subtle dead-code issue. Since `_init_schema()` always runs `CREATE TABLE IF NOT EXISTS network_meta` before `load()`, the table always exists — the `except Exception: pass` branch can never fire. If it *did* fire (e.g., some future schema change), it would skip the `elif network.nogoods:` derivation path, silently defaulting to `_next_nogood_id = 1` even when nogoods exist. This is masked because the except is unreachable in practice. The pattern is consistent with the existing `repos` loading (line 188) so it's not a blocker, but a restructure to `try/except` only around the SQL query (not the derivation fallback) would be more robust.

### `reasons_lib/api.py:import_json`
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly parses imported nogood IDs and advances the counter. Minor style issue: `import re` is placed inside the function body (line 925 in the diff) rather than at module top. This is functional but inconsistent — `import_beliefs.py` already has `import re` at module level. Not a blocker.

### `reasons_lib/import_beliefs.py:import_into_network`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same pattern as the JSON import path. Uses `re.fullmatch` (not `re.match`) to avoid partial matches. The `max()` call ensures the counter only ratchets upward, which is correct when importing nogoods with non-sequential IDs (e.g., `nogood-010`).

### `tests/test_nogood_id.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive 188-line test file covering: monotonic counter basics (3 tests), persistence round-trips including deletion scenarios and backward compat (4 tests), JSON import (1 test), and beliefs import (1 test). The tests directly exercise the deletion scenarios that motivate the fix (`del net.nogoods[-1]`, `net.nogoods.clear()`). The backward-compat tests simulate old databases by dropping the table or deleting the meta row — both are valid. One note: `test_old_db_without_meta_table` works because `Storage.__init__` recreates the table, so the test actually validates "empty meta table" rather than "missing meta table," but the end result is the same and the test is still valuable.

### Self-Review
**Limitations:** I could not see the full `_init_schema` method to confirm it uses `executescript(SCHEMA)` with the full SCHEMA string (verified only that line 70 calls `executescript`). Could not verify whether any other code path besides `add_nogood`, `import_json`, and `import_into_network` creates nogoods — there may be paths in tests or scripts not included in the diff. Could not run the test suite to confirm all 211+ tests still pass with this change.

### Feature Requests
- Include the output of `uv run pytest tests/ -v` so reviewers can confirm no regressions
- Show callers of `add_nogood` across the codebase to verify no other entry points were missed

## gemini [BLOCK]

### reasons_lib/storage.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The backward compatibility logic for old databases is fundamentally broken in two places. First, in `load()`, if the `network_meta` table is missing, `self.conn.execute("SELECT ...")` raises an `sqlite3.OperationalError`. The `except Exception:` block catches this and does a `pass`, completely skipping the `elif network.nogoods:` block that derives `_next_nogood_id` from existing nogoods. This leaves the ID counter at 1, causing the exact collisions this PR is meant to fix. The derivation fallback needs to happen outside the `try` block. Second, in `save()`, `self.conn.execute("DELETE FROM network_meta")` is called unconditionally. If a user loads an old database that lacks this table, the application will crash during save because the table does not exist.

### tests/test_nogood_id.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The tests for old databases unintentionally mask the bugs in `storage.py`. `test_old_db_without_meta_table` doesn't insert any nogoods before dropping the table, so it never checks if the fallback derivation logic actually executes correctly when the table is missing. It also does not call `save()` on the loaded old database, hiding the save crash. `test_old_db_with_nogoods_derives_counter` incorrectly simulates an old database by running `DELETE FROM network_meta` instead of `DROP TABLE`, meaning the table still exists, which avoids triggering the flawed exception branch in `load()`.

### reasons_lib/network.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The core implementation introduces a monotonically increasing counter (`_next_nogood_id`) that successfully resolves the ID collision issue caused by deletion. It correctly assigns and increments the state.

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Updates the monotonic counter correctly when importing from JSON data. Parsing the ID string to maintain the max counter value ensures that any new additions following an import will not collide.

### reasons_lib/import_beliefs.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Identical to the JSON import, correctly extracts the counter integer via regex and keeps the highest value to ensure ID uniqueness when importing from markdown representation.

### Self-Review
**Limitations:** I cannot strictly confirm whether `import re` is present globally in `reasons_lib/import_beliefs.py` (since only the diff is visible), though tests verify it runs without a `NameError`. I also do not have `api.init_db()` visible to see if the table schema might be proactively applied to old databases elsewhere, but `storage.py` clearly attempts (and fails) to handle missing schema cases itself.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#35
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/network.py:Network.__init__ + add_nogood
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Core fix is clean. `_next_nogood_id` starts at 1, increments after each use, and the old `len(self.nogoods) + 1` derivation is fully replaced. The `:03d` format is preserved. The counter is a plain instance attribute — no risk of shared state across Network instances.

### reasons_lib/storage.py:save
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `network_meta` table is created in the schema, cleared on save, and populated with the counter. The `DELETE FROM network_meta` is added alongside the other table clears, consistent with the existing save-by-replacement pattern.

### reasons_lib/storage.py:load
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The backwards-compat fallback uses a bare `except Exception: pass` to handle missing `network_meta` table. This is correct for the stated purpose but overly broad — it will silently swallow genuine errors like a corrupted `value` column or `int()` conversion failure. A narrower `except sqlite3.OperationalError` would be more precise. The fallback derivation logic itself is correct: it scans existing nogoods and picks `max_id + 1`. Additionally, `import re` is done inline inside a conditional block — minor style issue but functional. Both of these are concerns, not blockers.

### reasons_lib/storage.py:schema (network_meta table)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `network_meta` table uses `CREATE TABLE IF NOT EXISTS`, which is safe for both new and existing databases. The schema is minimal and appropriate — a key-value store for network-level metadata. Good design choice for extensibility.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

After appending each imported nogood, the counter is updated via `max(current, parsed_id + 1)`. The `re.fullmatch` correctly matches only the `nogood-NNN` format, and the `max()` call ensures the counter is never lowered. The `import re` was added at file top-level. The spurious blank line added between the comment and the `from . import Nogood` is trivial.

### reasons_lib/import_beliefs.py:import_into_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same counter-update pattern as `import_json`. Logic is correct — `re` was already imported at the top of this file. Consistent with the api.py approach.

### reasons_lib/import_agent.py:_import_nogoods (NOT in diff)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

This function bypasses `add_nogood()` and appends directly to `network.nogoods` without updating `_next_nogood_id`. It creates agent-prefixed IDs like `"agent:nogood-001"` which won't match the `nogood-(\d+)` regex, so the gap is unlikely to cause collisions in practice. However, it represents an unprotected append path — if someone later changes the ID format to omit the prefix, the counter would be stale. Consider adding a comment noting why the counter update is skipped, or extracting the counter-update logic into a helper.

### tests/test_nogood_id.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Excellent test coverage across 5 classes. Tests verify: monotonic increment, deletion doesn't reuse IDs, clear doesn't reset counter, persistence across save/load, backwards compatibility with old DBs (both with and without existing nogoods), JSON import counter update, and beliefs import counter update. The `test_old_db_with_nogoods_derives_counter` test is particularly valuable — it simulates upgrading a pre-fix database by dropping `network_meta` and verifying derivation works. One minor note: no test for the edge case where a non-numeric nogood ID exists (e.g., from agent import), but this is covered implicitly by the `re.fullmatch` guard.

### Self-Review
**Limitations:** I relied on the explore agent's report for `import_agent.py` context — I did not independently verify the exact line numbers or full function body. The `export_markdown.py` path was not checked for any nogood creation. I could not run the test suite to confirm all 190 tests pass.

### Feature Requests
- Include full file contents for all files that touch the modified subsystem, not just files in the diff — `import_agent.py` was a near-miss that could have been overlooked.
- Ability to run the test suite and report results as part of the review.

## gemini [PASS]

### `reasons_lib/api.py:import_json`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `import_json` function is correctly updated to maintain the monotonically increasing property of `_next_nogood_id` when new nogoods are imported. The use of regex safely ignores any non-standard IDs.

### `reasons_lib/import_beliefs.py:import_into_network`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `import_into_network` function correctly updates the `_next_nogood_id` state based on parsed nogoods, ensuring the network state remains correct after markdown ingestion.

### `reasons_lib/network.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Successfully replaces the append-only length-based assumption with a monotonic counter `_next_nogood_id` in `Network.__init__` and `Network.add_nogood`. This resolves the core issue of ID collisions on deletion.

### `reasons_lib/storage.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The database schema properly introduces the `network_meta` table to persist the counter. Loading handles backward compatibility gracefully by attempting to read the counter and gracefully deriving it from existing nogoods' max ID if not found (for old databases).

### `tests/test_nogood_id.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The tests are comprehensive, covering monotonic counter behavior, deletion edge cases, save/load persistence, migration logic for missing schema components, and updates via external imports (JSON/Markdown).

### Self-Review
**Limitations:** None. The diffs were self-contained and clear, with adequate context available.

Issue #2, opened 2026-03-23
Add Repos section to export-markdown output


benthomasson:
## Context

`beliefs.md` has a `## Repos` header section mapping repo names to filesystem paths (e.g., `- agents-python: ~/git/agents-python`). This is used by `beliefs check-refs` to resolve source paths.

## Proposed Change

Include a static `## Repos` header in `export-markdown` output, configured via `reasons init` or a config file.

## Priority

Low — the Repos section is currently empty in our primary use case. `reasons` already resolves source paths for `check-stale` without it.

See: `entries/2026/03/23/gap-analysis-reasons-export-markdown-vs-beliefsmd-for-primary-store-migration.md` in agents-python-expert.

PR #42, opened 2026-04-24
Add access_tags for data source provenance filtering

### LINKED ISSUE #38: Access control: filter beliefs by data source provenance (RBAC)
## Problem

When the belief network is used as a cache layer in front of data pipelines (e.g., Dataverse), beliefs derived from restricted data sources become accessible to users who lack access to those sources directly. A belief like `sales:q1-bookings-94-percent` derived from RevenueMaster should not be visible to users who can't query RevenueMaster.

This was identified while prototyping a belief-as-cache architecture in mini-analyze, where Dataverse query results are proposed as new beliefs. Different Dataverse data products have different access controls (e.g., RoverPeople for HR data, RevenueMaster for financials), and those restrictions need to flow through to cached beliefs.

### 1. `access_tags` metadata convention

Each belief carries a list of data source tags in its metadata:

```python
metadata = {
    "access_tags": ["roverpeople"],  # or ["revenuemaster"], ["public"], etc.
}
```

Premise beliefs get tagged when created. No schema change needed -- uses the existing `metadata_json` column.

### 2. Automatic inheritance for derived beliefs

When a belief is derived from premises, it inherits the union of all parent access tags. If `planning-divorced-from-reality` depends on a premise tagged `["revenuemaster"]`, the derived belief automatically inherits `["revenuemaster"]`.

This should happen in the derivation/justification logic so manual tagging isn't needed for derived beliefs.

### 3. `visible_to` filter parameter

Add a `visible_to: set[str] | None` parameter to query APIs:

- `list_nodes(visible_to={"roverpeople", "public"})` -- returns only beliefs whose `access_tags` are a subset of the user's access set
- `search(query, visible_to=...)` -- same filtering on search results
- `show_node(node_id, visible_to=...)` -- returns None or raises if user lacks access

When `visible_to` is None (default), no filtering is applied -- preserving backward compatibility.

### 4. `trace_access_tags(node_id)` utility

Traces a belief back to its premises via `trace_assumptions()` and returns the union of all access tags in the dependency chain. Useful for auditing what restrictions a derived belief carries.

## Example

```python
# Premise from HR data
api.assert_belief("ansible:cc724-headcount-242", "R&D Ansible has 242 people",
                  metadata={"access_tags": ["roverpeople"]})

# Derived belief (inherits roverpeople tag automatically)
api.derive("ansible:large-team-bus-factor-worse",
           "242-person team with bus factor 1 on safety",
           premises=["ansible:cc724-headcount-242", "triple-bus-factor-one-crisis"])
# access_tags = ["roverpeople"] (inherited from premise)

# Query with access control
api.list_nodes(visible_to={"public"})  # won't include either belief
api.list_nodes(visible_to={"roverpeople", "public"})  # includes both
```

## Context

- mini-analyze uses FTS5 over 14,831 beliefs as a semantic cache before Dataverse queries
- New beliefs are proposed from Dataverse results after each pipeline run
- Different Dataverse data products (RoverPeople, RevenueMaster, GTM, HGInsights) have different RBAC policies
- Without access filtering, the cache becomes a data exfiltration path

benthomasson:
## Summary

- Beliefs can carry `access_tags` in metadata for RBAC filtering (no schema change — uses existing `metadata_json` column)
- Derived beliefs automatically inherit the union of all parent access tags via `_inherit_access_tags()`
- Query APIs (`list_nodes`, `search`, `lookup`, `show_node`) accept `visible_to` parameter to filter by tag subset
- New `trace_access_tags()` utility traces the full dependency chain and returns all tags
- CLI: `--access-tags` on `add`, `--visible-to` on `list`/`search`/`lookup`/`show`, new `trace-access-tags` subcommand

Closes #38

## Test plan

- [x] 25 new tests covering inheritance, filtering, and tracing
- [x] 457 total tests pass (432 existing + 25 new, zero regressions)
- [x] End-to-end CLI test with tagged premises, derived beliefs, and filtering

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -559,12 +559,19 @@ def trace_assumptions(node_id: str, visible_to: list[str] | None = None, db_path
         return {"node_id": node_id, "premises": premises}
 
 
-def trace_access_tags(node_id: str, db_path: str = DEFAULT_DB) -> dict:
+def trace_access_tags(node_id: str, visible_to: list[str] | None = None, db_path: str = DEFAULT_DB) -> dict:
     """Trace backward through dependency chains and return union of all access_tags.
 
     Returns: {"node_id": str, "access_tags": list[str]}
+    Raises PermissionError if node's access_tags are not a subset of visible_to.
     """
     with _with_network(db_path) as net:
+        if node_id not in net.nodes:
+            raise KeyError(f"Node '{node_id}' not found")
+        if visible_to is not None and not _is_visible(net.nodes[node_id], visible_to):
+            raise PermissionError(
+                f"Node '{node_id}' requires access tags not in {visible_to}"
+            )
         tags = net.trace_access_tags(node_id)
         return {"node_id": node_id, "access_tags": tags}
 
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -716,6 +716,7 @@ def export_network(visible_to: list[str] | None = None, db_path: str = DEFAULT_D
             "nogoods": [
                 {"id": ng.id, "nodes": ng.nodes, "discovered": ng.discovered, "resolution": ng.resolution}
                 for ng in net.nogoods
+                if visible_to is None or all(n in net.nodes and _is_visible(net.nodes[n], visible_to) for n in ng.nodes)
             ],
             "repos": dict(net.repos),
         }
@@ -1076,7 +1077,7 @@ def export_markdown(visible_to: list[str] | None = None, db_path: str = DEFAULT_
             for nid, node in net.nodes.items():
                 if _is_visible(node, visible_to):
                     filtered.nodes[nid] = node
-            filtered.nogoods = net.nogoods
+            filtered.nogoods = [ng for ng in net.nogoods if all(n in filtered.nodes for n in ng.nodes)]
             filtered.repos = net.repos
             return _export(filtered, repos=filtered.repos)
         return _export(net, repos=net.repos)
@@ -1145,7 +1146,7 @@ def compact(budget: int = 500, truncate: bool = True, visible_to: list[str] | No
             for nid, node in net.nodes.items():
                 if _is_visible(node, visible_to):
                     filtered.nodes[nid] = node
-            filtered.nogoods = net.nogoods
+            filtered.nogoods = [ng for ng in net.nogoods if all(n in filtered.nodes for n in ng.nodes)]
             return _compact(filtered, budget=budget, truncate=truncate)
         return _compact(net, budget=budget, truncate=truncate)
 
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -473,7 +473,7 @@ def assert_node(node_id: str, db_path: str = DEFAULT_DB) -> dict:
         return {"changed": changed, "went_out": went_out, "went_in": went_in}
 
 
-def get_status(db_path: str = DEFAULT_DB) -> dict:
+def get_status(visible_to: list[str] | None = None, db_path: str = DEFAULT_DB) -> dict:
     """Get all nodes with truth values.
 
     Returns: {"nodes": list[dict], "in_count": int, "total": int}
@@ -481,6 +481,8 @@ def get_status(db_path: str = DEFAULT_DB) -> dict:
     with _with_network(db_path) as net:
         nodes = []
         for nid, node in sorted(net.nodes.items()):
+            if visible_to is not None and not _is_visible(node, visible_to):
+                continue
             nodes.append({
                 "id": nid,
                 "text": node.text,
@@ -520,23 +522,40 @@ def show_node(node_id: str, visible_to: list[str] | None = None, db_path: str = 
         }
 
 
-def explain_node(node_id: str, db_path: str = DEFAULT_DB) -> dict:
+def explain_node(node_id: str, visible_to: list[str] | None = None, db_path: str = DEFAULT_DB) -> dict:
     """Explain why a node is IN or OUT.
 
     Returns: {"steps": list[dict]}
+    Raises PermissionError if node's access_tags are not a subset of visible_to.
     """
     with _with_network(db_path) as net:
+        if node_id not in net.nodes:
+            raise KeyError(f"Node '{node_id}' not found")
+        if visible_to is not None and not _is_visible(net.nodes[node_id], visible_to):
+            raise PermissionError(
+                f"Node '{node_id}' requires access tags not in {visible_to}"
+            )
         steps = net.explain(node_id)
         return {"steps": steps}
 
 
-def trace_assumptions(node_id: str, db_path: str = DEFAULT_DB) -> dict:
+def trace_assumptions(node_id: str, visible_to: list[str] | None = None, db_path: str = DEFAULT_DB) -> dict:
     """Trace backward to find all premises a node rests on.
 
     Returns: {"node_id": str, "premises": list[str]}
+    Raises PermissionError if node's access_tags are not a subset of visible_to.
+    Filters returned premises by visible_to.
     """
     with _with_network(db_path) as net:
+        if node_id not in net.nodes:
+            raise KeyError(f"Node '{node_id}' not found")
+        if visible_to is not None and not _is_visible(net.nodes[node_id], visible_to):
+            raise PermissionError(
+                f"Node '{node_id}' requires access tags not in {visible_to}"
+            )
         premises = net.trace_assumptions(node_id)
+        if visible_to is not None:
+            premises = [p for p in premises if p in net.nodes and _is_visible(net.nodes[p], visible_to)]
         return {"node_id": node_id, "premises": premises}
 
 
@@ -664,7 +683,7 @@ def get_log(last: int | None = None, db_path: str = DEFAULT_DB) -> dict:
         return {"entries": entries}
 
 
-def export_network(db_path: str = DEFAULT_DB) -> dict:
+def export_network(visible_to: list[str] | None = None, db_path: str = DEFAULT_DB) -> dict:
     """Export the entire network as a dict.
 
     Returns: {"nodes": dict, "nogoods": list}
@@ -685,6 +704,7 @@ def export_network(db_path: str = DEFAULT_DB) -> dict:
                     "metadata": {k: v for k, v in n.metadata.items() if not k.startswith("_")},
                 }
                 for nid, n in sorted(net.nodes.items())
+                if visible_to is None or _is_visible(n, visible_to)
             },
             "nogoods": [
                 {"id": ng.id, "nodes": ng.nodes, "discovered": ng.discovered, "resolution": ng.resolution}
@@ -1035,7 +1055,7 @@ def list_repos(db_path: str = DEFAULT_DB) -> dict:
         return {"repos": dict(net.repos)}
 
 
-def export_markdown(db_path: str = DEFAULT_DB) -> str:
+def export_markdown(visible_to: list[str] | None = None, db_path: str = DEFAULT_DB) -> str:
     """Export the network as beliefs.md-compatible markdown.
 
     Returns: the markdown string
@@ -1043,6 +1063,15 @@ def export_markdown(db_path: str = DEFAULT_DB) -> str:
     from .export_markdown import export_markdown as _export
 
     with _with_network(db_path) as net:
+        if visible_to is not None:
+            from .network import Network
+            filtered = Network()
+            for nid, node in net.nodes.items():
+                if _is_visible(node, visible_to):
+                    filtered.nodes[nid] = node
+            filtered.nogoods = net.nogoods
+            filtered.repos = net.repos
+            return _export(filtered, repos=filtered.repos)
         return _export(net, repos=net.repos)
 
 
@@ -1095,7 +1124,7 @@ def hash_sources(
         return {"hashed": results, "count": len(results)}
 
 
-def compact(budget: int = 500, truncate: bool = True, db_path: str = DEFAULT_DB) -> str:
+def compact(budget: int = 500, truncate: bool = True, visible_to: list[str] | None = None, db_path: str = DEFAULT_DB) -> str:
     """Generate a token-budgeted belief state summary.
 
     Returns: the compact summary string
@@ -1103,6 +1132,14 @@ def compact(budget: int = 500, truncate: bool = True, db_path: str = DEFAULT_DB)
     from .compact import compact as _compact
 
     with _with_network(db_path) as net:
+        if visible_to is not None:
+            from .network import Network
+            filtered = Network()
+            for nid, node in net.nodes.items():
+                if _is_visible(node, visible_to):
+                    filtered.nodes[nid] = node
+            filtered.nogoods = net.nogoods
+            return _compact(filtered, budget=budget, truncate=truncate)
         return _compact(net, budget=budget, truncate=truncate)
 
 
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -18,6 +18,19 @@ from .storage import Storage
 DEFAULT_DB = "reasons.db"
 
 
+def _is_visible(node, visible_to: list[str]) -> bool:
+    """Check if a node is visible given the caller's access tags.
+
+    A node is visible if its access_tags are all contained in visible_to.
+    Nodes with no access_tags are always visible.
+    """
+    tags = node.metadata.get("access_tags", [])
+    if not tags:
+        return True
+    visible_set = set(visible_to)
+    return all(t in visible_set for t in tags)
+
+
 def _resolve_namespace(node_id: str, namespace: str | None) -> str:
     """Prefix node_id with namespace if provided and not already namespaced.
 
@@ -124,6 +137,7 @@ def add_node(
     source: str = "",
     namespace: str | None = None,
     any_mode: bool = False,
+    access_tags: list[str] | None = None,
     db_path: str = DEFAULT_DB,
 ) -> dict:
     """Add a node to the network.
@@ -138,6 +152,7 @@ def add_node(
         source: Provenance (repo:path)
         namespace: Optional namespace prefix (auto-creates ns:active premise)
         any_mode: If True, expand SL into one justification per antecedent (OR)
+        access_tags: Data source provenance tags for access control
         db_path: Path to RMS database
 
     Returns: {"node_id": str, "truth_value": str, "type": str, "premise_count": int}
@@ -192,11 +207,16 @@ def add_node(
                 j.antecedents = [_resolve_namespace(a, namespace) for a in j.antecedents]
                 j.outlist = [_resolve_namespace(o, namespace) for o in j.outlist]
 
+        metadata = {}
+        if access_tags:
+            metadata["access_tags"] = sorted(set(access_tags))
+
         node = net.add_node(
             id=node_id,
             text=text,
             justifications=justifications or None,
             source=source,
+            metadata=metadata or None,
         )
         jtype = justifications[0].type if justifications else "premise"
         max_premises = max((len(j.antecedents) for j in justifications), default=0)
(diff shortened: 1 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#42
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [BLOCK]

### reasons_lib/api.py:_is_visible
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correct subset check — a node is visible if all its `access_tags` are in `visible_to`. Nodes with no tags are always visible, preserving backward compatibility. Minor inefficiency: converts `visible_to` to a set on every call, but this is negligible in practice since it's a local utility.

### reasons_lib/api.py:add_node
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

At line 210-212, a fresh `metadata` dict is constructed containing only `access_tags`. This means if a caller passes `access_tags` alongside a namespace (which creates agent premise metadata at line 187), the namespace node's own metadata won't include access_tags and vice versa. More importantly, there is no way to pass arbitrary metadata (e.g., `confidence`) through `api.add_node` alongside `access_tags` — the metadata is exclusively access_tags or nothing. This is a pre-existing limitation but the new code entrenches it. Not a regression, but worth noting if other metadata callers exist.

### reasons_lib/api.py:show_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Raises `PermissionError` when `visible_to` is set and the node fails the visibility check. The exception message includes the `visible_to` list, which is appropriate for debugging. KeyError for missing nodes is checked first (correct priority — don't leak "this node exists but you can't see it" vs "not found").

### reasons_lib/api.py:search
**Verdict:** BLOCK
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

**Security concern — neighbor expansion leaks restricted nodes.** At line 1202-1215, after filtering `matched_ids` by `visible_to`, the neighbor expansion adds 1-hop antecedents and dependents WITHOUT any access check. These neighbors are then rendered in `_format_markdown` (line 1285-1287), `_format_json` (line 1306-1315), and `_format_minimal` — exposing their node IDs, text, and truth values. A user with `visible_to=["public"]` who searches for a public node that depends on a finance-restricted node will see the restricted node's full text in the "Related nodes" section. This is a data exfiltration path — the exact scenario the issue aims to prevent. Fix: filter `neighbor_ids` through `_is_visible` before passing to format functions.

### reasons_lib/api.py:lookup
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Filtering applied at the iteration level before building the search block. No neighbor expansion, so no leak path. Correct implementation.

### reasons_lib/api.py:list_nodes
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Filter applied inline during the node iteration loop. Backward compatible — `visible_to=None` skips the check entirely. Properly handles multi-tag nodes requiring all tags be present in the caller's set.

### reasons_lib/api.py:trace_access_tags
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean delegation to `net.trace_access_tags()`. Returns dict with node_id and sorted tag list. Useful auditing utility as specified in the issue.

### reasons_lib/network.py:_inherit_access_tags
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correct union semantics — merges the node's own tags with all antecedent tags across all justifications. Sorted output ensures deterministic metadata. Called from both `add_node` and `add_justification`, covering both creation and late-binding scenarios. Correctly skips premises (no justifications = early return). Silently skips forward-referenced antecedents (`if ant_id in self.nodes`), consistent with the rest of the network's handling of forward refs.

### reasons_lib/network.py:trace_access_tags
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

DFS walk through dependency subgraph collecting all access_tags. Uses `visited` set to handle diamond and cyclic dependencies correctly. Returns sorted list for deterministic output. Raises KeyError for missing nodes (consistent with other trace methods).

### reasons_lib/network.py — propagation gap
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

`_propagate()` does not call `_inherit_access_tags` when truth values change. This means tags are static after node creation — they won't update if a new antecedent is introduced indirectly. This is likely an intentional design choice (tags represent provenance at derivation time, not a dynamic property), but it's undocumented. If a node has multiple justifications and one becomes active that has different antecedents than the originally active one, the tags won't reflect the newly active justification's ancestry. This is acceptable for an RBAC system (tags should be the union of *all possible* sources, not just the currently active path), but should be documented.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Clean CLI integration. `--access-tags` on `add`, `--visible-to` on `show`/`search`/`lookup`/`list`, and `trace-access-tags` subcommand. `_parse_visible_to` helper avoids repetition. The `cmd_show` error handler correctly catches `PermissionError`. No CLI-level tests, but the API tests cover the underlying logic.

### tests/test_access_tags.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Good coverage of inheritance (single, union, chain, diamond, explicit+inherited merge), visibility filtering (list, show, lookup, search), and trace. Two concerns: (1) The test class `TestTraceAccessTags` defines its own `db_path` fixture at line 225 that shadows the module-level fixture at line 10 — this works but is confusing and should use the module fixture. (2) **Missing test for the neighbor leak**: no test verifies that `search(visible_to=...)` doesn't leak restricted nodes through neighbor expansion — this is the actual security bug I flagged above. (3) No tests for the CLI layer (e.g., `--access-tags`, `--visible-to` arg parsing).

### Unfiltered API surfaces
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

Several APIs that expose node content do not support `visible_to` filtering: `explain_node()`, `trace_assumptions()`, `get_status()`, `export()`, `export_markdown()`, `compact()`. While the issue spec only lists `list_nodes`, `search`, and `show_node`, the stated goal is preventing data exfiltration through the cache layer. A user who can't see a finance-restricted node via `list_nodes` can still call `explain_node("finance-node")` or `trace_assumptions("finance-node")` to see its text and structure. These are lower-priority since they require knowing the node ID, but `get_status()` lists all nodes without filtering and is a direct bypass. Consider at minimum adding `visible_to` to `get_status()`.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Did not see the full `list_nodes` function body to verify filter placement relative to depth calculation and other filters. Did not check if `import_beliefs` or `import_json` code paths handle access_tags.

### Feature Requests
- Include full function bodies for all modified functions in the observation section, not just the diff context
- Show callers of modified API functions (e.g., who calls `search()`, `list_nodes()`) to verify all entry points are covered
- Auto-run the test suite and include results in the review context
- Flag when a filtering/security feature is added to some API surfaces but not others in the same module

## gemini [BLOCK]

### reasons_lib/api.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** VIOLATES
**Test Coverage:** PARTIAL
**Integration:** WIRED

In the `search` function, access filtering is applied to `matched_ids` but omitted for `neighbor_ids`. If a visible node matches the search, any restricted antecedents or dependents will be included in the expanded graph output, directly bypassing access controls and leaking sensitive data. Furthermore, `show_node` returns raw dependent and antecedent IDs without filtering, which may leak restricted data if the identifiers themselves contain descriptive payload data (e.g., `ansible:cc724-headcount-242`).

### reasons_lib/network.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** VIOLATES
**Test Coverage:** PARTIAL
**Integration:** WIRED

When `add_justification` is called, `_inherit_access_tags` correctly updates the target node's access tags but fails to propagate these changes forward. Any existing dependents of the target node will not inherit the newly added tags, leaving them incorrectly exposed to unauthorized users and violating the requirement that derived beliefs automatically inherit access controls.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Parses the comma-separated `--access-tags` and `--visible-to` arguments correctly, handles the new `PermissionError` cleanly when showing a node, and successfully wires the new `trace-access-tags` command into the CLI interface.

### tests/test_access_tags.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** PARTIAL
**Integration:** WIRED

The test suite verifies basic inheritance and node-level filtering, but fails to cover the critical architectural edge cases that break the RBAC logic: 1) `api.search` leaking restricted neighbor nodes during 1-hop expansion, and 2) `network.add_justification` failing to propagate newly attached access tags to existing dependents.

### Self-Review
**Limitations:** Could not view the full function body of `api.search` to see the exact mechanism that serializes the `neighbor_ids` into the string output, although the diff logic clearly shows the `_is_visible` check is entirely bypassed during the neighborhood expansion step.

### Feature Requests
- Include the complete function bodies for any method that is modified, not just the diff hunks, to easily trace how un-filtered variables (like `neighbor_ids`) are ultimately returned or formatted.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#42
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### `reasons_lib/network.py:_inherit_access_tags`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly merges access_tags from all justification antecedents into the node's metadata. Handles the case where the node already has explicit tags by union-merging. Correctly skips premises (no justifications). Sorting ensures deterministic output. Called from both `add_node` (line 110) and `add_justification` (line 461).

### `reasons_lib/network.py:_propagate_access_tags`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

The BFS propagation pushes tags forward to dependents when a justification is added. However, `_propagate_access_tags` is called from `add_justification` but **not** from `add_node`. This is acceptable because a newly-added node has no dependents yet. But there's a more subtle gap: **tags are never retracted.** If a premise's access_tags are manually removed (e.g., via direct metadata manipulation or data reclassification), dependents retain the stale tags. The propagation is additive-only. For a security feature, this is the conservative direction (over-restrict rather than under-restrict), but it's worth noting as a design limitation. Also, the `visited` set prevents infinite loops in diamond graphs, which is correct.

### `reasons_lib/network.py:trace_access_tags`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Walks backward through all justification antecedents collecting the union of access_tags. Uses visited set to handle diamonds. Raises KeyError for missing nodes. Returns sorted list for deterministic output. Correctly collects tags from intermediate nodes too, not just premises — this means it captures tags added at any level of the dependency chain.

### `reasons_lib/api.py:_is_visible`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Subset check — a node is visible iff all its access_tags are in the caller's visible_to set. Nodes with no tags are always visible (backward compatible). The `all(t in visible_set for t in tags)` is equivalent to `set(tags) <= visible_set` and correct. Minor inefficiency: `visible_set` is recomputed per call from the list, but this is negligible for the expected cardinalities.

### `reasons_lib/api.py:add_node` (access_tags parameter)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The metadata construction at lines 210-212 has a subtle issue: `if access_tags:` is falsy for an empty list `[]`, which is fine. But more importantly, if the caller passes **other metadata** (e.g., via a future parameter), it would be overwritten because `metadata` is initialized as a fresh dict. Currently there's no `metadata` parameter on `add_node()` at the API level so this isn't a bug today, but it sets up a future footgun. Also, `metadata or None` on line 219 means an empty dict becomes `None`, which is then handled by `metadata or {}` in `network.py:add_node`. This is correct but fragile.

### `reasons_lib/api.py:show_node` (visible_to)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Raises `PermissionError` when the node's tags aren't a subset of visible_to. When `visible_to` is None, no check is performed (backward compatible). The error message includes the visible_to set, which could leak information about what tags the caller has — minor information disclosure but acceptable for a local CLI tool.

### `reasons_lib/api.py:search` (visible_to filtering)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Access filtering is applied in two places: once on matched_ids (line 1194-1200) and once on neighbor_ids (line 1220-1225). This correctly prevents the "search neighbor leak" mentioned in the commit message — without neighbor filtering, a restricted node could appear as a 1-hop neighbor of a public match. The ordering is correct: filter matches first, expand neighbors from filtered matches, then filter neighbors. One edge case: a restricted node that is a *neighbor* of a public match won't appear, but the public match's justification list might still *name* the restricted node's ID in formatted output. This is a minor metadata leak (ID only, not text).

### `reasons_lib/api.py:lookup` (visible_to filtering)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple filter in the iteration loop at line 1125-1126. Correct placement — before the search term matching, so restricted nodes are skipped entirely.

### `reasons_lib/api.py:list_nodes` (visible_to filtering)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Filter is applied in the main loop alongside other filters (premises, challenged, depth). Correct placement and semantics.

### `reasons_lib/api.py:trace_access_tags`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Thin wrapper around `net.trace_access_tags()`. Returns dict with node_id and sorted access_tags list. No visible_to check on the node itself — this is correct because `trace_access_tags` is an auditing/introspection function, not a data-access function.

### `reasons_lib/cli.py` (all changes)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Clean CLI integration. `_parse_visible_to` helper avoids code duplication. `cmd_show` handles `PermissionError`. `cmd_trace_access_tags` handles `KeyError`. `--access-tags` and `--visible-to` use comma-separated format consistently. The `cmd_show` function duplicates the `_parse_visible_to` logic inline (lines 213-215) instead of calling the helper — minor inconsistency but not a bug. **No CLI-level tests** for the argument parsing or command output formatting, but the API-level tests cover the underlying logic.

### `tests/test_access_tags.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

253 lines covering inheritance (single parent, union, chain, diamond, explicit+inherited merge, no-tag passthrough), add_justification propagation, storage round-trip, visibility filtering (list_nodes, show_node, lookup, search, search neighbors), and trace_access_tags (premise, chain, diamond, missing node, API wrapper). Good coverage of edge cases. Tests correctly verify that multi-tag nodes require ALL tags (not any). The `test_search_neighbors_filtered` test at line 220 specifically validates the neighbor leak fix. One gap: no test for `get_status()` with access filtering (but `get_status` doesn't support `visible_to`, which is noted below).

### Overall Assessment: Missing `visible_to` on `get_status`, `explain_node`, `trace_assumptions`, `export_network`, `export_markdown`
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The issue specifies `visible_to` for "query APIs" and lists `list_nodes`, `search`, and `show_node`. The implementation covers these plus `lookup`. However, `get_status()` returns all nodes without any access filtering, and `explain_node()` / `trace_assumptions()` could reveal restricted data through their outputs (premise names, justification chains). The `export_network()` and `export_markdown()` functions dump the entire network unfiltered. For a local CLI tool this is acceptable since the filtering is advisory, but for a networked service these would be data exfiltration paths. The issue's "Without access filtering, the cache becomes a data exfiltration path" warning applies to these functions too.

### Self-Review
**Limitations:** Could not run the test suite to verify all 253 lines of test_access_tags.py pass. Did not examine storage.py's `save()` method in full detail to verify that metadata changes from `_propagate_access_tags` are persisted (the context manager `_with_network` saves on exit, so propagated tags during the same session should persist, but cross-session propagation depends on save timing). Did not verify whether `_format_markdown` in search output leaks restricted node IDs through justification references in visible nodes.

### Feature Requests
- Show the full body of modified functions (not just diff hunks) in the review input to reduce the need for file reads
- Include callers of modified public API functions to verify integration completeness
- Flag public API functions in the same module that were NOT modified, to help detect missing integration (e.g., `get_status` lacks `visible_to`)

## gemini [CONCERN]

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The API correctly integrates `visible_to` filtering across `list_nodes`, `search`, `lookup`, and `show_node`. Access tags are stored in the existing `metadata` dictionary as specified. The filtering logic (`_is_visible`) correctly allows untagged nodes to remain public while strictly requiring the caller to possess all tags associated with restricted nodes.

### reasons_lib/cli.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** PARTIAL

The `_parse_visible_to` function uses a truthiness check `if getattr(args, "visible_to", None):`. If a caller explicitly provides an empty string (`--visible-to=""`) to represent a user with zero access tags (expecting to see only public/untagged data), this evaluates to `False`, returning `None`, which completely disables filtering and inadvertently exposes all restricted data. It should explicitly check for `None` to safely distinguish between an omitted argument (disable filtering) and an empty string (filter to empty tags).

### reasons_lib/network.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The tag propagation correctly pushes tags forward via breadth-first traversal, but has two logical edge cases typical of Truth Maintenance Systems that should be evaluated:
1. It unions tags across *all* justifications (OR-logic). If a fact is proven independently by public data (J1) and restricted data (J2), it inherits the restricted tags. This prevents a user with public access from viewing a fact that was already proven by a public source. 
2. In `_inherit_access_tags`, `inherited = set(node.metadata.get("access_tags", []))` means tags can only monotonically grow. If a justification is ever removed (or a node redefined), the inherited restricted tags will incorrectly persist on the node because the system does not track the difference between explicitly applied tags and inherited tags.

### tests/test_access_tags.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Excellent test coverage (253 lines). It thoroughly covers inheritance propagation (linear chains, diamond patterns), access filtering across various read queries (`show`, `list_nodes`, `search`, `lookup`), and validates the new `trace_access_tags` utility.

### Self-Review
**Limitations:** I did not have visibility into how `cmd_nogood` or hypothetical justification removals affect the graph structure in `Network`. The monotonic growth of tags (Concern #2 in `network.py`) might be completely acceptable if FTL Reasons is an append-only network, but it was flagged as questionable to ensure data isn't permanently tainted if graph edges are mutable.

PR #50, opened 2026-04-28
Add PostgreSQL-native storage backend

### LINKED ISSUE #46: PostgreSQL-native storage backend for enterprise/multi-user deployment
## Summary

Add a PostgreSQL-native storage backend (`reasons_lib.pg`) with row-level operations, replacing the current load-all/save-all pattern. This enables concurrent writers, multi-tenant deployment, and integration with enterprise systems like agents-python and expert-service.

## Problem

ftl-reasons has two storage modes today, neither suitable for enterprise:

1. **SQLite + in-memory Network** — CLI loads `reasons.db` into a `Network` object, operates in memory, writes back. Single user, single writer.

2. **PostgreSQL + in-memory Network** — expert-service's `PgStorage` loads the full network from PG into memory, operates, then clears all rows and writes everything back. Multi-tenant by `project_id`, but the load/save is destructive and doesn't support concurrent writers.

The core problem is the in-memory `Network` object — it assumes one process owns the full state. A `retract()` call currently loads thousands of nodes into Python, flips bits in memory, deletes all rows, and reinserts everything. This doesn't scale.

## Proposed solution: `reasons_lib.pg`

A new PostgreSQL-native backend where each operation is a database transaction, not a full network reload.

### Row-level operations

When someone calls `retract(node_id)`, it should:
1. UPDATE the single row's `truth_value`
2. Cascade via SQL queries that walk justification dependencies
3. Log to `propagation_log`
4. Return changed node IDs

Instead of: load N nodes into Python, flip bits in memory, delete N rows, insert N rows.

### Same API surface

```python
from reasons_lib.pg import PgApi

api = PgApi(connection_string, project_id=project_id)

# Same operations, but each is a SQL transaction
api.assert_belief(node_id, text)
api.retract(node_id)  # cascades via SQL
api.search(query)     # uses tsvector + GIN index
api.what_if_retract(node_id)  # read-only simulation
api.list_nodes(status="IN", visible_to={"public", "roverpeople"})
```

### Propagation in SQL

The TMS propagation algorithm (BFS justification evaluation, cascading retraction) would be implemented as either:
- Application-level SQL queries in Python (simpler, portable)
- PL/pgSQL stored procedures (faster for large networks, keeps logic in DB)

The algorithm is recursive but not complex — walk dependents, evaluate each justification's antecedents and outlist, update truth_value if changed, repeat.

### Enterprise capabilities enabled

| Capability | How |
|-----------|-----|
| Concurrent writers | Row-level transactions instead of full replace |
| Multi-tenant | `project_id` scoping (already in expert-service schema) |
| RBAC (#38) | `visible_to` filter in SQL WHERE clauses on `access_tags` |
| Full-text search | `tsvector` + GIN index (already exists in expert-service) |
| Semantic search | pgvector embeddings |
| Audit trail | `propagation_log` table + optional PG triggers |
| Backup/HA | Standard PostgreSQL operations |

## Architecture

```
reasons_lib/
  api.py          # Existing SQLite API (unchanged, CLI + single-user)
  network.py      # In-memory Network (reference implementation)
  pg.py           # NEW: PostgreSQL-native API (row-level ops)
  pg_propagate.py # NEW: TMS propagation as SQL queries
```

Keep `reasons_lib.Network` as the reference implementation for CLI and single-user use. The PG backend provides the same API surface but backed by transactions.

## Migration path

1. Keep SQLite backend for CLI (`reasons` command, single-user)
2. Build `reasons_lib.pg` with row-level operations
3. expert-service migrates from `PgStorage` (load/save) to `reasons_lib.pg` (incremental)
4. agents-python integrates via `reasons_lib.pg` for shared belief networks
5. RBAC filtering (#38) implemented at query layer

## Context

- expert-service already has the PG schema (`rms_nodes`, `rms_justifications`, `rms_nogoods`, `rms_propagation_log`) with GIN indexes
- `reasons ask` (#43) works against SQLite today, should work against PG with the same API
- RBAC (#38) requires query-time filtering that's natural in SQL but awkward in the in-memory Network
- agents-python integration will need concurrent access to shared belief networks
- mini-analyze's belief-as-cache pattern proposes beliefs from external data sources — concurrent writes from multiple pipeline runs need row-level ops

benthomasson:
## Summary

- Adds `reasons_lib.pg.PgApi` — PostgreSQL-native API where each operation is a SQL transaction (no full-network load/save)
- Enables concurrent writers and multi-tenant deployment via `project_id` composite PKs
- Schema compatible with expert-service existing `rms_*` tables
- Application-level BFS propagation with batched SQL queries per level
- 57 new tests covering add/retract/assert/cascade/outlist/search/nogoods/explain
- psycopg v3 as optional dependency (`pip install 'ftl-reasons[pg]'`)

## What is included

Core mutations: `add_node`, `retract_node`, `assert_node`, `add_justification`
Reads: `get_status`, `show_node`, `search` (tsvector + GIN), `list_nodes`, `get_log`
Nogoods: `add_nogood` (with DDB), `find_culprits`, `explain_node`, `trace_assumptions`
Multi-tenancy: each operation scoped by `project_id`, projects fully isolated

## What is deferred

Namespace support, `what_if_retract/assert`, `challenge/defend/supersede`, `export/import`, `deduplicate`, async wrapper

## Test plan

- [x] 57 PG tests pass with `DATABASE_URL` set (PostgreSQL 15)
- [x] 57 PG tests skip cleanly without `DATABASE_URL`
- [x] 611 existing SQLite tests unaffected
- [x] Propagation correctness: retraction cascade, restoration, diamond, outlist, multiple justifications, retracted pin

Closes #46

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,8 @@ dependencies = []
 
 [project.optional-dependencies]
 test = ["pytest", "pytest-cov"]
+pg = ["psycopg[binary]>=3.1"]
+test-pg = ["psycopg[binary]>=3.1", "pytest", "pytest-cov"]
 
 [project.scripts]
 reasons = "reasons_lib.cli:main"
--- a/reasons_lib/pg.py
+++ b/reasons_lib/pg.py
@@ -111,7 +111,9 @@ class PgApi:
     def __enter__(self):
         return self
 
-    def __exit__(self, *exc):
+    def __exit__(self, exc_type, *exc):
+        if exc_type is not None:
+            self.conn.rollback()
         self.close()
 
     # ── Schema ──────────────────────────────────────────────────
@@ -397,21 +399,18 @@ class PgApi:
         pid = self.project_id
 
         with self.conn.cursor() as cur:
-            # tsvector search + id ILIKE fallback
-            terms = query.strip().split()
-            ts_query = " & ".join(f"'{t}'" for t in terms if t)
-
-            if ts_query:
+            # plainto_tsquery handles arbitrary user input safely
+            if query.strip():
                 cur.execute(
                     "SELECT id, text, truth_value, source, metadata "
                     "FROM rms_nodes "
                     "WHERE project_id = %s "
-                    "AND (to_tsvector('english', text) @@ to_tsquery('english', %s) "
+                    "AND (to_tsvector('english', text) @@ plainto_tsquery('english', %s) "
                     "     OR id ILIKE %s) "
                     "ORDER BY ts_rank(to_tsvector('english', text), "
-                    "         to_tsquery('english', %s)) DESC "
+                    "         plainto_tsquery('english', %s)) DESC "
                     "LIMIT 20",
-                    (pid, ts_query, f"%{query}%", ts_query),
+                    (pid, query, f"%{query}%", query),
                 )
             else:
                 cur.execute(
--- a/reasons_lib/pg.py
+++ b/reasons_lib/pg.py
@@ -1,0 +1,1233 @@
+"""PostgreSQL-native storage backend for the dependency network.
+
+Each operation is a SQL transaction — no full-network load/save.
+Enables concurrent writers and multi-tenant deployment.
+
+Requires psycopg v3: pip install 'psycopg[binary]>=3.1'
+"""
+
+import json
+from collections import deque
+from datetime import datetime
+
+try:
+    import psycopg
+    from psycopg.rows import dict_row
+except ImportError:
+    psycopg = None  # type: ignore[assignment]
+
+
+SCHEMA = """
+CREATE TABLE IF NOT EXISTS rms_nodes (
+    id TEXT NOT NULL,
+    project_id UUID NOT NULL,
+    text TEXT NOT NULL,
+    truth_value TEXT NOT NULL DEFAULT 'IN' CHECK (truth_value IN ('IN', 'OUT')),
+    source TEXT DEFAULT '',
+    source_hash TEXT DEFAULT '',
+    date TEXT DEFAULT '',
+    metadata JSONB DEFAULT '{}',
+    created_at TIMESTAMPTZ DEFAULT now(),
+    PRIMARY KEY (id, project_id)
+);
+
+CREATE TABLE IF NOT EXISTS rms_justifications (
+    id SERIAL PRIMARY KEY,
+    node_id TEXT NOT NULL,
+    project_id UUID NOT NULL,
+    type TEXT NOT NULL CHECK (type IN ('SL', 'CP')),
+    antecedents JSONB NOT NULL DEFAULT '[]',
+    outlist JSONB NOT NULL DEFAULT '[]',
+    label TEXT DEFAULT '',
+    FOREIGN KEY (node_id, project_id) REFERENCES rms_nodes(id, project_id) ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS rms_nogoods (
+    id TEXT NOT NULL,
+    project_id UUID NOT NULL,
+    nodes JSONB NOT NULL DEFAULT '[]',
+    discovered TEXT DEFAULT '',
+    resolution TEXT DEFAULT '',
+    PRIMARY KEY (id, project_id)
+);
+
+CREATE TABLE IF NOT EXISTS rms_propagation_log (
+    id SERIAL PRIMARY KEY,
+    project_id UUID NOT NULL,
+    timestamp TEXT NOT NULL,
+    action TEXT NOT NULL,
+    target TEXT NOT NULL,
+    value TEXT NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS rms_network_meta (
+    key TEXT NOT NULL,
+    project_id UUID NOT NULL,
+    value TEXT NOT NULL,
+    PRIMARY KEY (key, project_id)
+);
+"""
+
+INDEXES = """
+CREATE INDEX IF NOT EXISTS idx_rms_nodes_project ON rms_nodes(project_id);
+CREATE INDEX IF NOT EXISTS idx_rms_nodes_status ON rms_nodes(project_id, truth_value);
+CREATE INDEX IF NOT EXISTS idx_rms_justifications_node ON rms_justifications(node_id, project_id);
+CREATE INDEX IF NOT EXISTS idx_rms_nogoods_project ON rms_nogoods(project_id);
+CREATE INDEX IF NOT EXISTS idx_rms_log_project ON rms_propagation_log(project_id);
+CREATE INDEX IF NOT EXISTS idx_rms_nodes_fts ON rms_nodes USING gin(to_tsvector('english', text));
+CREATE INDEX IF NOT EXISTS idx_rms_justifications_antecedents ON rms_justifications USING gin(antecedents);
+CREATE INDEX IF NOT EXISTS idx_rms_justifications_outlist ON rms_justifications USING gin(outlist);
+"""
+
+
+def _require_psycopg():
+    if psycopg is None:
+        raise ImportError(
+            "psycopg is required for PostgreSQL support. "
+            "Install it with: pip install 'psycopg[binary]>=3.1'"
+        )
+
+
+class PgApi:
+    """PostgreSQL-native API for the dependency network.
+
+    Each method executes as a SQL transaction. No in-memory Network object.
+    """
+
+    def __init__(self, conninfo, project_id):
+        _require_psycopg()
+        if isinstance(conninfo, str):
+            self.conn = psycopg.connect(conninfo, autocommit=False)
+            self._owns_conn = True
+        else:
+            self.conn = conninfo
+            self._owns_conn = False
+        self.project_id = str(project_id)
+
+    def close(self):
+        if self._owns_conn:
+            self.conn.close()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *exc):
+        self.close()
+
+    # ── Schema ──────────────────────────────────────────────────
+
+    def init_db(self):
+        with self.conn.cursor() as cur:
+            cur.execute(SCHEMA)
+            cur.execute(INDEXES)
+        self.conn.commit()
+        return {"project_id": self.project_id, "created": True}
+
+    # ── Core mutations ──────────────────────────────────────────
+
+    def add_node(self, node_id, text, sl="", cp="", unless="", label="",
+                 source="", access_tags=None):
+        pid = self.project_id
+        now = datetime.now().isoformat(timespec="seconds")
+        metadata = {}
+        if access_tags:
+            metadata["access_tags"] = sorted(access_tags)
+
+        with self.conn.cursor() as cur:
+            cur.execute(
+                "SELECT 1 FROM rms_nodes WHERE id = %s AND project_id = %s",
+                (node_id, pid),
+            )
+            if cur.fetchone():
+                raise ValueError(f"Node '{node_id}' already exists")
+
+            cur.execute(
+                "INSERT INTO rms_nodes (id, project_id, text, source, date, metadata) "
+                "VALUES (%s, %s, %s, %s, %s, %s)",
+                (node_id, pid, text, source, now, json.dumps(metadata)),
+            )
+
+            justifications = self._parse_justifications(sl, cp, unless, label)
+
+            for j in justifications:
+                cur.execute(
+                    "INSERT INTO rms_justifications (node_id, project_id, type, antecedents, outlist, label) "
+                    "VALUES (%s, %s, %s, %s, %s, %s)",
+                    (node_id, pid, j["type"],
+                     json.dumps(j["antecedents"]), json.dumps(j["outlist"]), j["label"]),
+                )
+
+            # Inherit access tags from antecedents
+            if justifications:
+                self._inherit_access_tags(cur, node_id, justifications)
+
+            # Compute initial truth value
+            if justifications:
+                truth = self._compute_truth(cur, node_id)
+            else:
+                truth = "IN"
+
+            cur.execute(
+                "UPDATE rms_nodes SET truth_value = %s WHERE id = %s AND project_id = %s",
+                (truth, node_id, pid),
+            )
+
+            self._log(cur, "add", node_id, truth)
+
+            # Count premises for return value
+            cur.execute(
+                "SELECT COUNT(*) FROM rms_nodes WHERE project_id = %s "
+                "AND NOT EXISTS (SELECT 1 FROM rms_justifications j "
+                "WHERE j.node_id = rms_nodes.id AND j.project_id = rms_nodes.project_id)",
+                (pid,),
+            )
+            premise_count = cur.fetchone()[0]
+
+        self.conn.commit()
+        return {
+            "node_id": node_id,
+            "truth_value": truth,
+            "type": "premise" if not justifications else "derived",
+            "premise_count": premise_count,
+        }
+
+    def add_justification(self, node_id, sl="", cp="", unless="", label=""):
+        pid = self.project_id
+
+        with self.conn.cursor() as cur:
+            cur.execute(
+                "SELECT truth_value FROM rms_nodes WHERE id = %s AND project_id = %s",
+                (node_id, pid),
+            )
+            row = cur.fetchone()
+            if not row:
+                raise KeyError(f"Node '{node_id}' not found")
+            old_value = row[0]
+
+            justifications = self._parse_justifications(sl, cp, unless, label)
+            if not justifications:
+                raise ValueError("No justification specified (use --sl or --cp)")
+
+            for j in justifications:
+                cur.execute(
+                    "INSERT INTO rms_justifications (node_id, project_id, type, antecedents, outlist, label) "
+                    "VALUES (%s, %s, %s, %s, %s, %s)",
+                    (node_id, pid, j["type"],
+                     json.dumps(j["antecedents"]), json.dumps(j["outlist"]), j["label"]),
+                )
+
+            self._inherit_access_tags(cur, node_id, justifications)
+
+            new_value = self._compute_truth(cur, node_id)
+            changed = []
+
+            if old_value != new_value:
+                cur.execute(
+                    "UPDATE rms_nodes SET truth_value = %s WHERE id = %s AND project_id = %s",
+                    (new_value, node_id, pid),
+                )
+                changed.append(node_id)
+                went_out, went_in = self._propagate(cur, node_id)
+                changed.extend(went_out)
+                changed.extend(went_in)
+
+            self._log(cur, "add-justification", node_id, new_value)
+
+        self.conn.commit()
+        return {
+            "node_id": node_id,
+            "old_truth_value": old_value,
+            "new_truth_value": new_value,
+            "changed": changed,
+        }
+
+    def retract_node(self, node_id, reason=""):
+        pid = self.project_id
+
+        with self.conn.cursor() as cur:
+            cur.execute(
+                "SELECT truth_value, metadata FROM rms_nodes WHERE id = %s AND project_id = %s",
+                (node_id, pid),
+            )
+            row = cur.fetchone()
+            if not row:
+                raise KeyError(f"Node '{node_id}' not found")
+
+            old_value, metadata = row[0], row[1]
+            if isinstance(metadata, str):
+                metadata = json.loads(metadata)
+
+            metadata["_retracted"] = True
+            if reason:
+                metadata["retract_reason"] = reason
+
+            if old_value == "OUT":
+                cur.execute(
+                    "UPDATE rms_nodes SET metadata = %s WHERE id = %s AND project_id = %s",
+                    (json.dumps(metadata), node_id, pid),
+                )
+                self.conn.commit()
+                return {"changed": [], "went_out": [], "went_in": []}
+
+            cur.execute(
+                "UPDATE rms_nodes SET truth_value = 'OUT', metadata = %s "
+                "WHERE id = %s AND project_id = %s",
+                (json.dumps(metadata), node_id, pid),
+            )
+            self._log(cur, "retract", node_id, reason or "OUT")
+
+            went_out, went_in = self._propagate(cur, node_id)
+
+        self.conn.commit()
+        all_changed = [node_id] + went_out + went_in
+        return {"changed": all_changed, "went_out": [node_id] + went_out, "went_in": went_in}
+
+    def assert_node(self, node_id):
+        pid = self.project_id
+
+        with self.conn.cursor() as cur:
+            cur.execute(
+                "SELECT truth_value, metadata FROM rms_nodes WHERE id = %s AND project_id = %s",
+                (node_id, pid),
+            )
+            row = cur.fetchone()
+            if not row:
+                raise KeyError(f"Node '{node_id}' not found")
+
+            old_value, metadata = row[0], row[1]
+            if isinstance(metadata, str):
+                metadata = json.loads(metadata)
+
+            if old_value == "IN":
+                return {"changed": [], "went_out": [], "went_in": []}
+
+            metadata.pop("_retracted", None)
+            cur.execute(
+                "UPDATE rms_nodes SET truth_value = 'IN', metadata = %s "
+                "WHERE id = %s AND project_id = %s",
+                (json.dumps(metadata), node_id, pid),
+            )
+            self._log(cur, "assert", node_id, "IN")
+
+            went_out, went_in = self._propagate(cur, node_id)
+
+        self.conn.commit()
+        all_changed = [node_id] + went_out + went_in
+        return {"changed": all_changed, "went_out": went_out, "went_in": [node_id] + went_in}
+
+    # ── Read operations ─────────────────────────────────────────
+
+    def get_status(self, visible_to=None):
+        pid = self.project_id
+        with self.conn.cursor() as cur:
+            cur.execute(
+                "SELECT n.id, n.text, n.truth_value, n.metadata, "
+                "(SELECT COUNT(*) FROM rms_justifications j "
+                " WHERE j.node_id = n.id AND j.project_id = n.project_id) AS jcount "
+                "FROM rms_nodes n WHERE n.project_id = %s ORDER BY n.id",
+                (pid,),
+            )
+            rows = cur.fetchall()
+
+        nodes = []
+        for row in rows:
+            nid, text, tv, meta, jcount = row
+            if isinstance(meta, str):
+                meta = json.loads(meta)
+            if visible_to is not None and not self._is_visible(meta, visible_to):
+                continue
+            nodes.append({
+                "id": nid,
+                "text": text,
+                "truth_value": tv,
+                "justification_count": jcount,
+            })
+
+        in_count = sum(1 for n in nodes if n["truth_value"] == "IN")
+        return {"nodes": nodes, "in_count": in_count, "total": len(nodes)}
+
+    def show_node(self, node_id, visible_to=None):
+        pid = self.project_id
+        with self.conn.cursor() as cur:
+            cur.execute(
+                "SELECT id, text, truth_value, source, source_hash, metadata "
+                "FROM rms_nodes WHERE id = %s AND project_id = %s",
+                (node_id, pid),
+            )
+            row = cur.fetchone()
+            if not row:
+                raise KeyError(f"Node '{node_id}' not found")
+
+            nid, text, tv, source, source_hash, meta = row
+            if isinstance(meta, str):
+                meta = json.loads(meta)
+
+            if visible_to is not None and not self._is_visible(meta, visible_to):
+                raise PermissionError(f"Access denied for node '{node_id}'")
+
+            cur.execute(
+                "SELECT type, antecedents, outlist, label FROM rms_justifications "
+                "WHERE node_id = %s AND project_id = %s ORDER BY id",
+                (node_id, pid),
+            )
+            justifications = []
+            for jrow in cur.fetchall():
+                jtype, ants, outs, jlabel = jrow
+                if isinstance(ants, str):
+                    ants = json.loads(ants)
+                if isinstance(outs, str):
+                    outs = json.loads(outs)
+                j = {"type": jtype, "antecedents": ants, "outlist": outs, "label": jlabel}
+                justifications.append(j)
+
+            dependents = sorted(self._find_dependents(cur, [node_id]))
+
+        return {
+            "id": nid,
+            "text": text,
+            "truth_value": tv,
+            "source": source,
+            "source_hash": source_hash,
+            "justifications": justifications,
+            "dependents": dependents,
+            "metadata": meta,
+        }
+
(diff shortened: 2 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#50
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: ADDRESSES
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Optional dependencies for `pg` and `test-pg` are cleanly defined. Using `psycopg[binary]>=3.1` is correct for psycopg v3. The extras don't pollute the default install.

---

### reasons_lib/pg.py:SCHEMA
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Schema matches the existing expert-service tables mentioned in the issue (`rms_nodes`, `rms_justifications`, `rms_nogoods`, `rms_propagation_log`). Composite primary key `(id, project_id)` enables multi-tenancy. `CREATE TABLE IF NOT EXISTS` is idempotent. The FK with `ON DELETE CASCADE` on justifications is correct. GIN indexes on `antecedents` and `outlist` JSONB arrays support the `@>` containment queries used in `_find_dependents`.

### reasons_lib/pg.py:PgApi.__init__
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Accepts either a connection string or an existing connection object — good. However, `__exit__` only calls `self.close()` and does not handle rollback on exception. If a mutation method raises mid-transaction (after cursor work but before `self.conn.commit()`), the connection is left with an open failed transaction. psycopg v3's `with conn:` context manager auto-commits/rollbacks, but `PgApi` doesn't leverage that. A user who wraps `PgApi` in a `with` block and gets an exception will have the connection closed but not rolled back, leaving the next user of a pooled connection in a broken state.

### reasons_lib/pg.py:add_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly inserts node, creates justifications, inherits access tags, computes initial truth value, and commits atomically. The duplicate check via SELECT-then-INSERT has a theoretical TOCTOU race under concurrent writers (another session could insert between the SELECT and INSERT), but the PK constraint will catch it — the error message would be less friendly (IntegrityError vs. ValueError). Acceptable for a first implementation.

### reasons_lib/pg.py:add_justification
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** PARTIAL
**Test Coverage:** PARTIAL
**Integration:** WIRED

The method doesn't check the `_retracted` flag. In the reference Network implementation, `add_justification` calls `_compute_truth` then `_propagate`, and `_propagate` skips retracted nodes — but the node *itself* having `_retracted` could still get its truth value updated by `_compute_truth`. The test `test_add_justification_changes_truth` covers the happy path (retracted via `retract_node` then adding justification changes to IN) but this implicitly clears the retraction by changing truth value without clearing the `_retracted` metadata flag. This means the node has `truth_value='IN'` but `metadata._retracted=True`, which would cause `_propagate` to skip it on future cascades — a subtle inconsistency.

### reasons_lib/pg.py:retract_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly sets `_retracted` flag, handles already-OUT case, updates truth value, logs, and propagates. Return shape `{changed, went_out, went_in}` is richer than the reference API's flat list, which is fine — it's strictly more informative.

### reasons_lib/pg.py:assert_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly clears `_retracted` flag via `metadata.pop("_retracted", None)`, matching the reference implementation. Handles already-IN case. Propagation is correct.

### reasons_lib/pg.py:_propagate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Faithfully ports the BFS propagation from the reference `Network._propagate`. Key behaviors preserved: (1) skips `_retracted` nodes, (2) skips premises (no justifications), (3) evaluates all justifications (ANY valid → IN), (4) tracks visited set to prevent cycles, (5) batch-fetches truth values for efficiency. The batched approach is a nice optimization over the reference's per-node queries. `_find_dependents` correctly searches both `antecedents` and `outlist` JSONB arrays, ensuring non-monotonic propagation (outlist unblocking) works.

### reasons_lib/pg.py:search
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The tsquery construction on line 402 (`" & ".join(f"'{t}'" for t in terms)`) wraps each term in single quotes and passes it as a parameterized value to `to_tsquery`. This is not SQL injection — the value is parameterized. However, the single-quote wrapping is part of tsquery syntax, and user input containing single quotes (e.g., searching for "it's") would produce malformed tsquery like `'it's'` which would cause a PostgreSQL syntax error from `to_tsquery`. The terms should be sanitized or use `plainto_tsquery` / `websearch_to_tsquery` instead, which handle arbitrary user input safely.

### reasons_lib/pg.py:_find_dependents
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Loops over `node_ids` issuing one query per node. For large cascades with many changed nodes, this could be N+1 queries. Could be batched into a single query using `jsonb_array_elements_text` or multiple `@>` conditions combined with `OR`. This is a performance concern rather than a correctness bug — the results are correct.

### reasons_lib/pg.py:add_nogood
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly implements dependency-directed backtracking: records nogood, checks if contradiction is active, finds culprits via entrenchment scoring, retracts the least-entrenched premise, and propagates. The fallback to "fewest dependents" when no culprits found is reasonable. The early return with `self.conn.commit()` on line 591 for inactive nogoods is correct.

### reasons_lib/pg.py:_is_visible
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Matches the reference `api.py:_is_visible` exactly — nodes with no access_tags are visible to everyone, otherwise all tags must be in the `visible_to` set.

### reasons_lib/pg.py (missing methods)
VERDICT: CONCERN
CORRECTNESS: N/A
SPEC_COMPLIANCE: PARTIAL
ISSUE_COMPLIANCE: PARTIAL
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: MISSING
REASONING: The issue claims "same API surface" but the PG backend is missing several methods from the reference API: `what_if_retract`, `what_if_assert` (mentioned explicitly in the issue), `challenge`, `defend`, `convert_to_premise`, `supersede`, `summarize`, `export_network`, `export_markdown`, `import_beliefs`, `import_json`, `compact`, `check_stale`, `hash_sources`, `deduplicate`, `derive_prompt`, `derive_apply`, `propagate` (full recompute), `get_belief_set`. The issue does say "same API surface" with `what_if_retract` shown in the example code block. This is a significant gap relative to the stated scope, though acceptable as a first PR if the missing methods are planned follow-ups.

---

### reasons_lib/pg.py:__exit__ (transaction safety)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

`__exit__` just calls `close()` without rolling back uncommitted work. If an exception occurs mid-mutation (after cursor writes but before `commit()`), the connection is closed with pending changes. For owned connections this is fine (close implies rollback in psycopg v3). But for passed-in connections (`_owns_conn=False`), `close()` is a no-op, leaving the connection in a dirty transaction state. The test fixture works around this with an explicit `api.conn.rollback()` in cleanup, but production callers might not.

### tests/conftest.py
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: ADDRESSES
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Clean fixture setup. Uses unique `uuid4()` project_id per test for isolation. Cleanup deletes by `project_id` in correct FK-respecting order. The `pytest.mark.pg` marker and `skip_no_pg` skip decorator ensure tests don't fail without a PG instance. The `api.conn.rollback()` before cleanup handles failed transaction state properly.

---

### tests/test_pg.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

466 lines, 40+ test cases covering all implemented methods. Test classes are well-organized by feature area. Covers core TMS behaviors: premise/derived truth values, retraction cascades, restoration cascades, diamond dependencies, outlist blocking/unblocking/reblocking, multiple justifications, `_retracted` pin semantics, access tag inheritance, visibility filtering, multi-tenancy isolation, nogoods with backtracking, entrenchment-based culprit selection, explain/trace. The `TestMultiTenancy.test_projects_isolated` test is particularly good — verifies same node IDs in different projects are independent.

### tests/test_pg.py:TestAddJustification
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Only two test cases for `add_justification`. Missing: test that adding a justification to a retracted node (with `_retracted=True`) handles the metadata flag correctly. As noted above, `add_justification` can change truth to IN without clearing `_retracted`, which would cause `_propagate` to skip this node in future cascades.

### Self-Review
**Limitations:** Could not run the PG tests to verify they pass against an actual database. Could not verify whether the existing `conftest.py` conflicts with any existing test configuration (the conftest was new). Did not verify the JSONB `@>` containment operator behavior with edge cases like empty arrays or arrays with special characters. Could not see expert-service's existing `PgStorage` to verify migration compatibility.

### Feature Requests
- Include the full existing API method list alongside the new implementation to automatically flag missing method coverage
- Show callers of the module (e.g., imports of `reasons_lib.pg` elsewhere in the repo) to verify integration
- For multi-file PRs, include dependency graph showing which new files reference which existing files

## gemini [BLOCK]

### pyproject.toml
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly adds `psycopg` v3 optional dependencies for the new PostgreSQL backend and associated testing.

### reasons_lib/pg.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

The implementation introduces critical bugs in the core Truth Maintenance System (TMS) logic:
1. **Broken BFS Propagation:** `_propagate` uses a `visited` set that spans across batches, preventing nodes from being re-evaluated in the same propagation run. If node Z is evaluated in batch 1 and added to `visited`, but another of its antecedents changes in batch 2, Z is skipped. This breaks non-monotonic cascades (e.g., `unless` rules) and leaves the graph in an inconsistent state.
2. **Invalid Nogood Tracing:** `_trace_assumptions_recursive` incorrectly traces through ALL justifications (including invalid ones) and collects premises regardless of their current truth value. This causes `add_nogood` to potentially blame and attempt to "retract" a premise that is already `OUT`, failing to resolve the contradiction. It must only trace the single valid justification supporting the node.
3. **Missing Propagation on Add:** `add_node` does not call `_propagate` when adding a new node that evaluates to `IN`. If an existing node already has a justification referencing this new node (which is allowed since JSONB references aren't foreign-keyed), the dependent node will not be updated.
4. **Unjustified Assertions:** `assert_node` unconditionally forces `truth_value = 'IN'` even for derived nodes without verifying their justifications. This can create unjustified `IN` nodes, breaking architectural invariants.
Integration is PARTIAL as the module is built but not wired into the CLI or other consumers (though this is expected per the outlined migration path).

### tests/conftest.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Provides robust test fixtures for isolating tests using unique `project_id` scoping, executing against PostgreSQL, and correctly tearing down the database state.

### tests/test_pg.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

While basic coverage is good, the test suite is missing critical edge cases that would have caught the fatal flaws in `pg.py`. There are no tests for:
- Complex non-monotonic cascades requiring a node to be re-evaluated multiple times during a single propagation.
- Nogood resolution on paths where intermediate nodes possess multiple justifications (to ensure invalid paths aren't traced).
- Adding a new node that subsequently fulfills a pre-existing justification on another node.

### Self-Review
**Limitations:** No major limitations. The full source of the new `pg.py` was provided, making it possible to conduct a deep algorithmic review of the graph traversal logic.

benthomasson:
## Review Response

### Gemini BLOCK — all 4 claims are false positives

Each claim describes behavior that exactly matches the reference `network.py` implementation:

1. **"Broken BFS Propagation"** — The `visited` set preventing re-evaluation is how `network.py:785` works. It prevents infinite loops in cyclic graphs, not re-evaluation bugs.

2. **"Invalid Nogood Tracing"** — `network.py:228-256` (`trace_assumptions`) also traces through ALL justifications, not just valid ones. This is intentional — you want all premises that *could* support a node.

3. **"Missing Propagation on Add"** — `network.py:109-120` also does not propagate to existing dependents when adding a new node. Same design.

4. **"Unjustified Assertions"** — `network.py:150-169` (`assert_node`) unconditionally sets IN. That is the semantics of `assert` in a TMS — it is an explicit user override.

### Claude CONCERN — two valid issues, both fixed

1. **tsquery single-quote injection** — Switched from `to_tsquery` with manual quoting to `plainto_tsquery` which handles arbitrary user input safely. Fixed in 0a10341.

2. **`__exit__` missing rollback** — Added `conn.rollback()` on exception in `__exit__`. Fixed in 0a10341.

### Acknowledged (not bugs, deferred by design)

- **Missing API methods** — Intentionally deferred to follow-up PRs as stated in the PR description.
- **`_find_dependents` N+1 queries** — Performance optimization, not correctness. Can batch in follow-up.
- **`add_justification` + `_retracted` flag** — Matches reference behavior in `network.py:445-478`.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#50
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: ADDRESSES
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Clean optional dependency groups. `pg` for runtime, `test-pg` for test runs. psycopg v3 is the right choice. The `[binary]` sub-package avoids needing libpq headers at install time.
---

### reasons_lib/pg.py:SCHEMA
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Schema matches the issue's description of `rms_nodes`, `rms_justifications`, `rms_nogoods`, `rms_propagation_log` tables. Primary keys are composite `(id, project_id)` for multi-tenancy. Foreign key on justifications cascades on delete. `truth_value` CHECK constraint enforces IN/OUT only, consistent with the project's "no STALE status" design decision in CLAUDE.md. `CREATE TABLE IF NOT EXISTS` is safe for repeated `init_db()` calls.

### reasons_lib/pg.py:INDEXES
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Good index coverage: GIN on `to_tsvector` for full-text search, GIN on `antecedents`/`outlist` JSONB for the `@>` containment queries used in `_find_dependents`. Status index for filtered queries. All match the issue's stated capabilities (FTS via tsvector + GIN).

### reasons_lib/pg.py:PgApi.__init__
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Accepts either a connection string or an existing connection object — good for testing and integration with connection pools. `autocommit=False` is correct for explicit transaction control. `project_id` is stringified, matching UUID text storage.

### reasons_lib/pg.py:PgApi.__exit__
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `__exit__` rolls back on exception and closes, but does NOT commit on clean exit. This means if a caller uses `with PgApi(...) as api:` without the api methods calling `self.conn.commit()` internally, uncommitted work could be lost. Currently each mutation method calls `self.conn.commit()` individually, so this works — but the asymmetry is a footgun. If someone adds a method that forgets `self.conn.commit()`, the `__exit__` won't save them. Also, `close()` is always called even on the success path, which could close a connection with an open uncommitted transaction (e.g., if someone reads without a commit between). Minor: the success path doesn't explicitly commit before close, relying on each method having committed already.

### reasons_lib/pg.py:PgApi.add_node
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** PARTIAL
**Test Coverage:** COVERED
**Integration:** PARTIAL

Core logic is sound — checks for duplicates, inserts node, parses justifications, computes initial truth value, logs. However: (1) Missing `namespace` parameter that the existing `api.add_node()` supports. (2) Missing `any_mode` parameter. (3) The `premise_count` subquery is computed inside the cursor context but AFTER the commit boundary has already been reached by `self.conn.commit()` — wait, actually the commit is AFTER the cursor block, so this is fine. (4) No validation that antecedent node IDs in justifications actually exist before adding them — the in-memory Network would raise on unknown nodes. If you add a derived node `sl="nonexistent"`, it will silently create a justification referencing a missing node and compute the truth as OUT (because `truth_cache.get("nonexistent", "OUT")` defaults to "OUT"). This diverges from the in-memory Network behavior.

### reasons_lib/pg.py:PgApi.add_justification
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly checks node existence, adds justification, recomputes truth, propagates if changed, commits. The test `test_add_justification_changes_truth` covers the case of a retracted node gaining a valid justification and becoming IN. Access tag inheritance is handled.

### reasons_lib/pg.py:PgApi.retract_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Sets `_retracted` metadata flag, updates truth to OUT, propagates cascade. Correctly handles already-OUT nodes by updating metadata only and returning empty changed list. The `_retracted` flag is checked during propagation to skip pinned-retracted nodes, matching the in-memory Network's behavior. Tests cover basic retraction, cascade, already-out, reason metadata, and not-found error.

### reasons_lib/pg.py:PgApi.assert_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly clears `_retracted` flag, sets IN, propagates. Already-IN case returns empty. Tests verify restoration cascade through diamond dependencies.

### reasons_lib/pg.py:PgApi._propagate
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** PARTIAL
**Test Coverage:** COVERED
**Integration:** WIRED

The BFS propagation has a subtle issue with the `truth_cache`. The cache is rebuilt from scratch each BFS iteration (each time the queue is drained into a `batch`). When multiple levels of dependents are processed, the `truth_cache` for iteration N+1 is fetched fresh from the database — which is correct since we UPDATE the DB inline. However, the `truth_cache` does NOT include the `changed_id` itself or nodes from earlier batches unless they happen to be referenced by the current batch's justifications. The line `truth_cache[dep_id] = new_tv` at line 750 updates the cache for subsequent evaluations within the SAME batch, which is good. But if a dependent references a node that changed in an earlier batch and isn't in `all_referenced`, the DB read will get the updated value anyway. So correctness is maintained, but the per-batch cache rebuild is wasteful. Additionally, `_find_dependents` loops per node_id with individual queries rather than batching — O(N) queries per BFS level where N is the batch size. This is a performance concern for large networks but not a correctness bug.

### reasons_lib/pg.py:PgApi.search
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The `plainto_tsquery` fix is good — prevents tsquery injection. However, the ILIKE pattern `f"%{query}%"` does not escape LIKE wildcard characters (`%`, `_`, `\`). A query containing `%` or `_` would match more broadly than intended. This is a minor issue since it's a search function (broader matching is arguably acceptable), but it's a deviation from exact substring matching. Not a security vulnerability since it can't cause SQL injection (parameterized), but it's a correctness concern.

### reasons_lib/pg.py:PgApi.list_nodes
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** PARTIAL
**Test Coverage:** COVERED
**Integration:** PARTIAL

Missing `challenged`, `min_depth`, `max_depth` filter parameters that exist in the SQLite `api.list_nodes()`. The `has_dependents` filter uses f-string SQL construction (`f"FROM rms_nodes n WHERE {where}"`) — while the `where` variable is built from hardcoded condition strings and parameterized values, this pattern is safe here but fragile. The `has_dependents` implementation queries for nodes that appear in OTHER nodes' justifications, which is correct. Tests cover all implemented filters.

### reasons_lib/pg.py:PgApi.add_nogood
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly implements nogood registration, checks active contradiction, finds culprits via entrenchment scoring, retracts the least-entrenched premise. The fallback to fewest-dependents when no culprits are found matches the in-memory Network's behavior. Uses `ON CONFLICT DO UPDATE` for the next_nogood_id counter, which is correct for concurrent access. The early-return `self.conn.commit()` inside the cursor context on line 590 is fine since the cursor is still open for the outer `with` block.

### reasons_lib/pg.py:PgApi._find_dependents
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation iterates per node_id, issuing one SQL query per ID. For a batch of N node_ids during BFS propagation, this means N separate queries. The JSONB containment operator `@>` doesn't support `ANY(array)` directly for the needle, but this could be restructured with a single query using `jsonb_array_elements_text` or a UNION-based approach. For large networks, this is a performance bottleneck since propagation calls `_find_dependents` at each BFS level. Not a correctness issue, but the issue specifically asks for "row-level operations" that scale — this N+1 query pattern may not.

### reasons_lib/pg.py (missing methods)
VERDICT: CONCERN
CORRECTNESS: N/A
SPEC_COMPLIANCE: PARTIAL
ISSUE_COMPLIANCE: PARTIAL
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: UNTESTED
INTEGRATION: MISSING
REASONING: The PgApi is missing several methods from the existing API surface: `what_if_retract`, `what_if_assert` (read-only simulation — explicitly mentioned in the issue!), `challenge`, `defend` (dialectical argumentation — a core TMS feature per CLAUDE.md), `export_network`, `compact`, `convert_to_premise`, `summarize`, `supersede`, `deduplicate`, `import_beliefs`, `import_json`, `export_markdown`, `derive_prompt`, `get_belief_set`, `trace_access_tags`, `check_stale`, `hash_sources`. The issue specification explicitly mentions `what_if_retract` in the API example. The "Same API surface" requirement from the issue is only partially met.
---

### reasons_lib/pg.py (no CLI integration)
VERDICT: CONCERN
CORRECTNESS: N/A
SPEC_COMPLIANCE: PARTIAL
ISSUE_COMPLIANCE: PARTIAL
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: MISSING
REASONING: The CLI has no way to use the PG backend. There's no `--backend pg` flag, no `DATABASE_URL` env var handling in the CLI, and `api.py` has no abstraction to swap between SQLite and PG backends. The issue's migration path says "expert-service migrates from PgStorage to reasons_lib.pg" — this is a library-only integration, which is reasonable for an initial commit. But the issue also says "`reasons ask` should work against PG with the same API" — that integration path doesn't exist yet.
---

### tests/conftest.py
VERDICT: CONCERN
CORRECTNESS: QUESTIONABLE
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: The cleanup uses f-string table names in SQL (`f"DELETE FROM {table} WHERE project_id = %s"`). While the table names are hardcoded literals in the loop, this pattern sets a precedent. The cleanup order is correct (dependents before parent due to foreign keys). However, the `api.conn.rollback()` call before cleanup assumes the connection might be in an error state — this is defensive and good. One issue: if `init_db()` fails (e.g., schema already exists with a different structure), the fixture will error without cleanup. The `CREATE TABLE IF NOT EXISTS` mitigates this.
---

### tests/test_pg.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

466 lines of well-structured tests covering: add (premise, derived IN/OUT, outlist, duplicate, access tags, inheritance), retract (basic, already-out, cascade, reason, not-found), assert (restore, already-in, restore cascade), propagation (diamond, outlist block/unblock/reblock, multiple justifications, retracted-pin skip), status (empty, with nodes, visible_to filter), show (premise, derived, dependents, not-found, access denied), search (text, ID, no results, compact/json format, neighbors), list (all, status, premises, dependents, namespace), log (after add, last), add-justification (changes truth, not-found), nogood (active, inactive, not-found), culprits (entrenchment ordering), explain (premise, derived IN/OUT), trace (premise, chain, diamond), multi-tenancy (project isolation). Good coverage of the implemented surface. The `test_retracted_pin_skipped` test is particularly good — it verifies that a node retracted directly stays OUT even when its dependency is restored.

### tests/test_pg.py:TestMultiTenancy
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Creates a second PgApi with a different project_id, verifies that nodes with the same ID are isolated between projects. Cleanup of the second project's data is handled in the `finally` block. This directly validates the multi-tenancy claim from the issue.

### Self-Review
**Limitations:** - Could not run the tests against a real PostgreSQL instance to verify runtime behavior
- Did not see the full Network.retract() and Network.assert_node() to compare propagation semantics line-by-line
- Cannot verify the JSONB GIN index actually works with the `@>` containment queries without a running database
- Did not check whether psycopg v3 returns JSONB columns as dicts or strings (the code handles both with `isinstance(meta, str)` checks, which is defensive but suggests uncertainty about driver behavior)

### Feature Requests
- Include the full source of modified or referenced functions (like `Network._propagate`) alongside the diff for behavioral comparison
- Show callers of new modules to verify integration claims
- Automated API surface comparison: list methods present in api.py but missing from pg.py

## gemini [CONCERN]

### `reasons_lib/pg.py`
**Verdict:** CONCERN
**Correctness:** BROKEN
**Spec Compliance:** PARTIAL
**Test Coverage:** COVERED
**Integration:** MISSING

The PostgreSQL storage backend implements the requested API surface but contains several critical flaws in its Truth Maintenance System (TMS) logic, concurrency, and performance:
1. **Missing Propagation in `add_node`**: When a node is added and evaluates to `IN`, it does not call `_propagate()`. If other nodes previously referenced this node in their antecedents or outlists (which are evaluated as `OUT` by default when missing), their truth values will not be updated when the node is created.
2. **Incorrect Assumption Tracing**: `_trace_assumptions_recursive` unconditionally follows all justifications. It must only follow antecedents of the *currently valid* justification (where antecedents are `IN` and outlist is `OUT`). Tracing invalid justifications causes `find_culprits` to falsely identify and retract premises that aren't actually supporting the node.
3. **Flawed `visited` Logic in `_propagate`**: The BFS uses a permanent `visited` set per propagation wave. In a non-monotonic TMS (due to outlists/`unless`), a node's truth value can oscillate during a single wave as its peers are updated. Blacklisting evaluated nodes prevents them from settling to their correct final truth values.
4. **Race Condition in `nogood_id`**: Reading `next_nogood_id`, incrementing in Python, and performing an `INSERT ... ON CONFLICT DO UPDATE` creates a read-modify-write race condition. Concurrent writers (a stated PR goal) will collide and cause unique constraint violations on `rms_nogoods`. This should use an atomic `UPDATE ... RETURNING` clause.
5. **N+1 Query Issue**: `_find_dependents` runs a `SELECT` query in a loop for every given `node_id`. During batched propagation, this causes severe N+1 query bottlenecks. It should use `jsonb_array_elements_text` with an `IN (...)` or `ANY(...)` clause to fetch dependents in one query.
6. **Memory Inefficiency**: `list_nodes(has_dependents=True)` fetches all dependency targets for the *entire project* into memory to populate `ids_with_deps` before filtering, rather than using an `EXISTS` clause on the query itself.
7. **Integration**: The `PgApi` class is implemented but never instantiated or imported anywhere in the core repository logic (like `cli.py` or `api.py`), so it is completely unreachable outside of tests.

### `tests/conftest.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The fixture accurately provides isolated database states per test using unique `project_id` UUIDs and performs adequate cleanup of inserted records.

### `tests/test_pg.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Excellent test coverage over the new `PgApi` methods. Note that while tests pass, they miss complex scenarios (like adding forward-referenced nodes or tracing inactive justifications) that would expose the underlying logic bugs in `pg.py`.

### `pyproject.toml`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Safely adds `psycopg` to optional dependencies to avoid breaking default installations while enabling the PG feature.

### Self-Review
**Limitations:** Could not inspect the existing `reasons_lib/network.py` reference implementation to verify how the original in-memory version handles cycles and `visited` node protections during its propagation loop, but general TMS principles highlight the flaws present in the PG implementation.

### Feature Requests
- Include full file contexts for reference implementations (like `network.py` in this case) when reviewing new architectural implementations that are meant to mirror existing complex logic.

PR #52, opened 2026-04-28
Add challenge, defend, and compact to PgApi

### LINKED ISSUE #51: PgApi: add challenge, defend, and compact methods
## Context

PR #50 landed the PostgreSQL-native storage backend (`PgApi`). It covers the core TMS operations (add/retract/assert, nogoods, search, explain, trace) but deferred several methods.

Three of these are needed for expert-service integration — they're used by the chat agent tools (`expert_service/chat/tools.py`):

### `challenge(target_id, reason, challenge_id=None)`

Creates a challenge node that knocks the target OUT. This is a composition of existing operations:
1. Generate challenge ID (e.g., `challenge-{target_id}-001`) if not provided
2. `add_node(challenge_id, reason, source="challenge")`
3. `retract_node(target_id)` with metadata linking to the challenge
4. Store the challenge relationship in metadata

The SQLite `api.challenge()` implementation is at `reasons_lib/api.py`.

### `defend(target_id, challenge_id, reason, defense_id=None)`

Neutralizes a challenge and restores the target:
1. Generate defense ID if not provided
2. `retract_node(challenge_id)` — knock out the challenge
3. `assert_node(target_id)` — restore the target
4. Store defense metadata

### `compact(budget=500)`

Token-budgeted network summary. Priority order: nogoods first, then OUT nodes, then IN nodes by importance. Returns a string within the token budget.

This is a read-only formatting function that can be built on top of `list_nodes()` + `get_status()`. The SQLite implementation is at `reasons_lib/api.py:compact()`.

## Why these matter

All three are exposed as LangChain tools in expert-service's chat agent. Without them, migrating expert-service from the load-all/save-all `PgStorage` to row-level `PgApi` would require keeping the old storage layer just for these three operations.

## Implementation notes

- `challenge` and `defend` are thin wrappers around existing `add_node`/`retract_node`/`assert_node` — they add metadata conventions, not new TMS logic
- `compact` is purely read-side — no mutations, just formatting with a token budget
- All three should follow the existing PgApi pattern: SQL transactions, project_id scoping, access_tags/visible_to support
- Tests should cover: challenge knocks target OUT, defend restores it, compact respects budget, compact prioritizes nogoods > OUT > IN

Ref: #46, #50

benthomasson:
## Summary
- Adds challenge(), defend(), and compact() methods to the PostgreSQL-native PgApi class, porting the three highest-priority missing methods from issue #46
- challenge() creates a premise node and adds it to the target outlist, knocking the target OUT with cascade propagation
- defend() uses symmetric challenge-the-challenge mechanism — creates a defense node that challenges the challenge, restoring the original target
- compact() generates token-budgeted belief state summaries (nogoods > OUT > IN priority) directly from SQL, matching the output format of compact.py
- Extracts _challenge_internal() as a shared helper so both challenge() and defend() execute in a single atomic transaction

## Test plan
- [x] 8 challenge tests: premise/derived/already-out targets, custom ID, auto-ID collision, not-found, metadata, cascade
- [x] 5 defend tests: restores target, cascade restoration, metadata, not-found, custom ID
- [x] 5 compact tests: empty network, IN nodes, OUT nodes with reasons, budget limiting, visible_to filtering
- [x] All 75 PG tests pass (57 existing + 18 new)
- [x] All 611 existing tests pass (no regressions)

Closes #51

### DIFF
--- a/reasons_lib/pg.py
+++ b/reasons_lib/pg.py
@@ -317,6 +317,72 @@ class PgApi:
         all_changed = [node_id] + went_out + went_in
         return {"changed": all_changed, "went_out": went_out, "went_in": [node_id] + went_in}
 
+    # ── Dialectical operations ─────────────────────────────────
+
+    def challenge(self, target_id, reason, challenge_id=None):
+        with self.conn.cursor() as cur:
+            result = self._challenge_internal(cur, target_id, reason, challenge_id)
+        self.conn.commit()
+        return result
+
+    def defend(self, target_id, challenge_id, reason, defense_id=None):
+        pid = self.project_id
+
+        with self.conn.cursor() as cur:
+            # Verify target and challenge exist
+            cur.execute(
+                "SELECT 1 FROM rms_nodes WHERE id = %s AND project_id = %s",
+                (target_id, pid),
+            )
+            if not cur.fetchone():
+                raise KeyError(f"Node '{target_id}' not found")
+            cur.execute(
+                "SELECT 1 FROM rms_nodes WHERE id = %s AND project_id = %s",
+                (challenge_id, pid),
+            )
+            if not cur.fetchone():
+                raise KeyError(f"Challenge '{challenge_id}' not found")
+
+            # Generate defense ID
+            if defense_id is None:
+                defense_id = f"defense-{challenge_id}"
+                suffix = 1
+                while True:
+                    cur.execute(
+                        "SELECT 1 FROM rms_nodes WHERE id = %s AND project_id = %s",
+                        (defense_id, pid),
+                    )
+                    if not cur.fetchone():
+                        break
+                    suffix += 1
+                    defense_id = f"defense-{challenge_id}-{suffix}"
+
+            # Challenge the challenge (defense = challenge against the challenge)
+            result = self._challenge_internal(cur, challenge_id, reason, defense_id)
+
+            # Update defense node metadata
+            cur.execute(
+                "SELECT metadata FROM rms_nodes WHERE id = %s AND project_id = %s",
+                (defense_id, pid),
+            )
+            meta = cur.fetchone()[0]
+            if isinstance(meta, str):
+                meta = json.loads(meta)
+            meta["defense_target"] = challenge_id
+            meta["defends"] = target_id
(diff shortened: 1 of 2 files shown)


benthomasson:
## Code Review Summary (Round 3)

**Models:** Claude, Gemini | **Gate:** BLOCK (Gemini false positive)

### Claude: CONCERN (no blockers)
- All 6 observations are PASS/VALID except two minor CONCERN notes:
  - `defend()` relies on propagation to restore the target — if the target was `retract_node()`'d before being challenged, propagation skips `_retracted` nodes. This matches `Network.defend()` behavior exactly.
  - `compact()` duplicates formatting logic from `compact.py` — maintenance observation, not a bug. Necessary since PgApi works with SQL, not in-memory Network objects.

### Gemini: BLOCK — false positive (same pattern as PR #50 reviews)

Gemini claims challenging a premise creates a "broken" justification because `INSERT INTO rms_justifications` with empty antecedents and the challenge in the outlist "creates conditional positive support" that makes the target "magically become IN" when the challenge is defeated.

This is exactly how `Network.challenge()` works at `network.py:524-529`:

```python
# Target is a premise — convert to justified node
target.justifications.append(
    Justification(type="SL", antecedents=[], outlist=[challenge_id])
)
```

A justification with empty antecedents is vacuously valid (all zero antecedents are trivially IN). The synthetic SL justification gives the outlist somewhere to live — without it, there's no place to attach the challenge. When the challenge goes OUT (via defend), the outlist condition is satisfied and the premise correctly returns to IN. This is standard TMS semantics, not a bug.

This is the same pattern as PR #50 where Gemini flagged 4 "BROKEN" issues across two review rounds, all of which matched the reference `network.py` implementation exactly.

### Changes across 3 review rounds
- Round 1: Added edge-case tests (defend duplicate ID, multiple challenges, compact nogoods/sorting/summaries)
- Round 2: Fixed `source='challenge'` on challenge nodes, added nogood visibility filtering in `compact()`
- Round 3: No new issues found

### Test results
- 82/82 PG tests pass (57 existing + 25 new)
- 611/611 existing tests pass (no regressions)

PR #54, opened 2026-04-29
Add list-negative command with LLM classification

benthomasson:
## Summary

- Adds `list-negative` command that finds IN beliefs describing problems, defects, or risks
- Uses keyword pre-filtering to reduce candidates, then `claude -p` for LLM classification
- Returns only genuinely negative beliefs, avoiding false positives from beliefs about error handling mechanisms
- Adds `list_negative()` to `api.py`, `cmd_list_negative` to `cli.py`, and 5 tests with mocked LLM

## Test plan

- [x] 5 new tests pass (empty db, no keyword matches, classifies negatives, filters all, visible_to)
- [x] Full test suite passes (622 tests, 0 failures)
- [ ] Manual test on redhat-expert database with real LLM

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1575,14 +1575,14 @@ def list_negative(
         response = ask._invoke_claude(prompt)
 
         negative_ids = set()
-        match = re.search(r"\[.*?\]", response, re.DOTALL)
-        if match:
+        for match in re.finditer(r"\[.*?\]", response, re.DOTALL):
             try:
                 ids = json.loads(match.group())
                 if isinstance(ids, list):
                     negative_ids = set(ids)
+                    break
             except json.JSONDecodeError:
-                pass
+                continue
 
         candidate_map = {nid: text for nid, text in candidates}
         negative = [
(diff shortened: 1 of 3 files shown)


benthomasson:
## Code Review Summary (3 rounds, Claude + Gemini)

**Final gate: PASS (Gemini) / CONCERN-minor (Claude)**

### Issues found and fixed:
1. **Fragile JSON parsing** (both models) — line-by-line parsing broke on multi-line JSON arrays. Fixed with `re.finditer` to try each `[...]` match until one parses.
2. **Prose brackets before JSON** (Claude) — regex grabbed `[the analysis]` instead of `["a"]`. Fixed by the same `finditer` loop.
3. **Backtick fencing** (Claude) — belief text now fenced in backticks to reduce prompt injection surface.
4. **Missing tests** (Claude) — added tests for: unknown IDs from LLM, exception propagation, multi-line JSON, malformed responses, prose-with-brackets.

### Remaining minor notes (not blocking):
- Database connection held open during LLM call (~300s max). Read-only SQLite, no practical issue.
- Substring matching (`missing` matches `commissioning`, `lack` matches `black`). By design — LLM filters false positives.
- Backticks in belief text could break markdown fencing. Low severity for trusted data.

### Test coverage: 10 tests, all mocked LLM

PR #56, opened 2026-04-29
Add what_if_retract and what_if_assert to PgApi

### LINKED ISSUE #55: Add what_if_retract and what_if_assert to PgApi
## Summary

`what_if_retract` and `what_if_assert` exist in the SQLite backend (`api.py`) but not in `PgApi` (`pg.py`). expert-service uses PgApi and needs these for the issue finder workflow -- ranking root cause premises by cascade impact.

## Use case

The issue finder lists gated beliefs (OUT nodes blocked by IN outlist nodes) and traces them back to shared premises. `what_if_retract` on a root premise shows the full cascade: how many blockers go OUT, how many gated beliefs get restored, total impact. This turns diagnosis into prioritization.

Example:

```
what-if retract it:iad2-geographic-single-point-of-failure
  Would retract 3 nodes (depth 1-2)
  Would restore 7 gated beliefs (depth 2-3)
  Total affected: 10
```

## Implementation approach

Run the retract/assert in a PostgreSQL transaction, collect affected nodes with depth and dependent counts, then rollback. The cascade propagation logic already exists in `PgApi.retract_node` and `PgApi.assert_node` -- wrap them in a savepoint or transaction that rolls back.

## Expected API

Match the SQLite signatures:

- `what_if_retract(node_id)` returns `{node_id, already_out, retracted, restored, total_affected}`
- `what_if_assert(node_id)` returns `{node_id, already_in, retracted, restored, total_affected}`

Each item in retracted/restored: `{id, text, depth, dependents}`

## Context

- SQLite implementation: `api.py` lines 327-443
- Tests: `test_cli.py` TestWhatIf and TestWhatIfEdgeCases
- Consumer: expert-service issue finder (`rms_find_issues` tool)
- Related: PRs #53 (list-gated), #54 (list-negative)

benthomasson:
## Summary

- Adds `what_if_retract()` and `what_if_assert()` to PgApi for simulating cascade effects without mutating the database
- Uses transaction + ROLLBACK: runs the real retract/assert + propagation, collects before/after diffs, then rolls back
- Includes `_cascade_depth_pg()` for BFS depth computation and `_collect_what_if_results()` shared helper
- Returns `{node_id, already_out/in, retracted, restored, total_affected}` matching the SQLite API signatures

Closes #55

## Test plan

- [x] 7 new PG tests: cascade, already_out, not_found, no_mutation, assert_restores, already_in, outlist_restoration
- [x] Full PG suite passes (96 tests)
- [x] Full non-PG suite passes (627 tests)

### DIFF
--- a/reasons_lib/pg.py
+++ b/reasons_lib/pg.py
@@ -321,41 +321,41 @@ class PgApi:
 
     def what_if_retract(self, node_id):
         pid = self.project_id
-        with self.conn.cursor() as cur:
-            cur.execute(
-                "SELECT truth_value FROM rms_nodes WHERE id = %s AND project_id = %s",
-                (node_id, pid),
-            )
-            row = cur.fetchone()
-            if not row:
-                self.conn.rollback()
-                raise KeyError(f"Node '{node_id}' not found")
-            if row[0] == "OUT":
-                self.conn.rollback()
-                return {
-                    "node_id": node_id, "already_out": True,
-                    "retracted": [], "restored": [], "total_affected": 0,
-                }
+        try:
+            with self.conn.cursor() as cur:
+                cur.execute(
+                    "SELECT truth_value FROM rms_nodes WHERE id = %s AND project_id = %s",
+                    (node_id, pid),
+                )
+                row = cur.fetchone()
+                if not row:
+                    raise KeyError(f"Node '{node_id}' not found")
+                if row[0] == "OUT":
+                    return {
+                        "node_id": node_id, "already_out": True,
+                        "retracted": [], "restored": [], "total_affected": 0,
+                    }
 
-            cur.execute(
-                "SELECT id, truth_value FROM rms_nodes WHERE project_id = %s",
-                (pid,),
-            )
-            before = {r[0]: r[1] for r in cur.fetchall()}
+                cur.execute(
+                    "SELECT id, truth_value FROM rms_nodes WHERE project_id = %s",
+                    (pid,),
+                )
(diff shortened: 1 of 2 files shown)


benthomasson (commented):
Clean implementation — reuses existing _propagate() and _find_dependents() via transaction+rollback.

One edge case to note: _cascade_depth_pg follows antecedent relationships but outlist-restored nodes are reached via the outlist path, so they get depth=0 instead of the real cascade distance. Informational only, not a correctness issue.

Tests are solid, especially the no-mutation verification. LGTM.

benthomasson:
## Code Review Summary (2 rounds, Claude + Gemini)

**Final gate: PASS (Gemini) / CONCERN-informational (Claude)**

### Issues found and fixed:
1. **Missing try/finally for rollback** (both models, round 1) — wrapped both methods in try/finally to guarantee rollback even if _propagate raises
2. **Missing test coverage** (Claude, round 1) — added assert_not_found, assert_no_mutation, and dependents field tests (10 tests total)
3. **Outlist depth concern** (from issue submitter) — verified PG implementation correctly computes depth=1 for outlist-restored nodes because _find_dependents queries both antecedents and outlist columns. Added depth assertion. SQLite version may have this bug but PG does not.

### Remaining informational notes (not blocking):
- return 0 fallback in _cascade_depth_pg is semantically ambiguous but matches SQLite and should not occur in practice
- Full transaction rollback (vs savepoint) is safe because PgApi always commits at method boundaries
- O(N^2) depth calculation is fine for an occasionally-called diagnostic query

### Test coverage: 10 PG tests, all passing

PR #73, opened 2026-04-30
Add porter stemming and progressive query relaxation to FTS search

### LINKED ISSUE #72: reasons ask: FTS search misses beliefs due to strict AND + no stemming
## Summary

`reasons ask` fails to find beliefs containing dotted proper nouns like "Models.corp" and "GitLab.cee" even though the beliefs exist in the network. Caused 2/25 failures in a structured evaluation of the redhat-expert knowledge base.

## Root Cause

The interaction of three factors in `_fts_search()` (`reasons_lib/api.py:1289`):

1. **No stemming** — FTS5 default tokenizer (`unicode61`) does not stem. "deactivation" != "deactivated", "maximum" != "max", "limit" != "limits".
2. **Strict all-terms AND** — Every query term must appear in the belief text. One miss = zero results.
3. **LLM query formulation** — The LLM adds plausible synonyms ("deactivation" for "auto-deactivated") that are semantically correct but lexically absent.

## Reproduction

```bash
cd ~/git/redhat-expert

# These beliefs exist:
reasons search "Models.corp sandbox"    # finds it:modelscorp-sandbox-expires-21-days
reasons search "pipeline job"           # finds it:gitlabcee-pipeline-job-limits

# But reasons ask can't find them:
reasons ask "How long does Models.corp sandbox access last before deactivation?"
# -> "not found" (false negative)

reasons ask "What is the maximum number of jobs in a single GitLab.cee pipeline?"
# -> hallucinated wrong answer (said 200, actual is 250)
```

## Failure Traces

**Q3:** LLM searched `Models.corp sandbox access deactivation`. FTS query: `"Models.corp" "sandbox" "access" "deactivation"`. Failed because belief text says "auto-deactivated" not "deactivation". Dropping "deactivation" returns the belief immediately.

**Q4:** LLM searched `GitLab pipeline jobs maximum limit` then `GitLab CI CD pipeline`. Both failed because "maximum"/"limit"/"CI"/"CD" don't appear in the belief text (which says "max 250 jobs").

### 1. Add porter stemmer to FTS5 (highest value, lowest effort)

```sql
CREATE VIRTUAL TABLE nodes_fts USING fts5(id, text, tokenize="porter unicode61");
```

This makes "deactivation" match "deactivated" and "limit" match "limits". Requires rebuilding the FTS index.

### 2. Progressive query relaxation

If strict AND returns zero, drop terms one at a time and retry:

```python
for n in range(len(terms), max(1, len(terms) - 2), -1):
    for combo in combinations(terms, n):
        results = fts_search(combo)
        if results:
            return results
```

### 3. Suppress hallucination on miss

Separately, when no beliefs are found, the synthesis prompt should more strongly instruct the LLM to say "not found" rather than answering from training data. Q4 fabricated "200 jobs" (generic GitLab default) when the actual belief says 250.

## Context

Full evaluation: 25 questions across 5 categories scored 23/25 A-range. The two F grades both trace to this FTS bug. Detailed write-up in `redhat-expert/entries/2026/04/30/reasons-ask-fts-search-query-bug.md`.

benthomasson:
## Summary

- Adds porter stemmer to FTS5 index so word forms match their stems (e.g. "deactivation" matches "auto-deactivated")
- Adds progressive query relaxation: when strict AND returns nothing and query has >2 terms, drops terms one at a time until results are found
- Drop/recreate FTS table on save() to ensure existing databases get the new tokenizer

Fixes #72

## Test plan

- [x] `test_porter_stemming` — "deactivation" matches "deactivated"
- [x] `test_porter_stemming_plural` — "job" matches "jobs"
- [x] `test_progressive_relaxation` — 4-term query with 1 absent term still finds belief
- [x] `test_no_false_positives` — unrelated terms return nothing
- [x] Full test suite: 633 passed, 0 regressions

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1286,27 +1286,39 @@ def search(query: str, visible_to: list[str] | None = None, db_path: str = DEFAU
             return _format_markdown(net, matched_ids, neighbor_ids)
 
 
+def _fts_query(conn, terms: list[str]) -> list[str]:
+    fts_query = " ".join(f'"{t}"' for t in terms)
+    cursor = conn.execute(
+        "SELECT id FROM nodes_fts WHERE nodes_fts MATCH ? ORDER BY rank LIMIT 20",
+        (fts_query,),
+    )
+    return [row[0] for row in cursor.fetchall()]
+
+
 def _fts_search(query: str, db_path: str) -> list[str]:
-    """Search using FTS5 full-text index."""
+    """Search using FTS5 full-text index with porter stemming and progressive relaxation."""
     import sqlite3
+    from itertools import combinations
     try:
         conn = sqlite3.connect(db_path)
-        # FTS5 match: all terms must appear (implicit AND)
-        # Quote each term to avoid FTS syntax issues
-        terms = query.strip().split()
-        fts_query = " ".join(f'"{t}"' for t in terms if t)
-        if not fts_query:
+        terms = [t for t in query.strip().split() if t]
+        if not terms:
             conn.close()
             return []
-        cursor = conn.execute(
-            "SELECT id FROM nodes_fts WHERE nodes_fts MATCH ? ORDER BY rank LIMIT 20",
-            (fts_query,),
-        )
-        results = [row[0] for row in cursor.fetchall()]
+
+        results = _fts_query(conn, terms)
+
+        if not results and len(terms) > 2:
+            for n in range(len(terms) - 1, max(0, len(terms) - 3), -1):
+                for combo in combinations(terms, n):
+                    results = _fts_query(conn, list(combo))
+                    if results:
+                        conn.close()
+                        return results
+
         conn.close()
         return results
     except (sqlite3.OperationalError, sqlite3.DatabaseError):
-        # FTS table doesn't exist or query failed
         return []
 
 
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1301,23 +1301,23 @@ def _fts_search(query: str, db_path: str) -> list[str]:
     from itertools import combinations
     try:
         conn = sqlite3.connect(db_path)
-        terms = [t for t in query.strip().split() if t]
-        if not terms:
+        try:
+            terms = [t for t in query.strip().split() if t]
+            if not terms:
+                return []
+
+            results = _fts_query(conn, terms)
+
+            if not results and len(terms) > 2:
+                for n in range(len(terms) - 1, max(0, len(terms) - 3), -1):
+                    for combo in combinations(terms, n):
+                        results = _fts_query(conn, list(combo))
+                        if results:
+                            return results
+
+            return results
+        finally:
             conn.close()
-            return []
-
-        results = _fts_query(conn, terms)
-
-        if not results and len(terms) > 2:
-            for n in range(len(terms) - 1, max(0, len(terms) - 3), -1):
-                for combo in combinations(terms, n):
-                    results = _fts_query(conn, list(combo))
-                    if results:
-                        conn.close()
-                        return results
-
-        conn.close()
-        return results
     except (sqlite3.OperationalError, sqlite3.DatabaseError):
         return []
 
--- a/reasons_lib/storage.py
+++ b/reasons_lib/storage.py
@@ -57,7 +57,7 @@ CREATE TABLE IF NOT EXISTS network_meta (
     value TEXT NOT NULL
 );
 
-CREATE VIRTUAL TABLE IF NOT EXISTS nodes_fts USING fts5(id, text);
+CREATE VIRTUAL TABLE IF NOT EXISTS nodes_fts USING fts5(id, text, tokenize="porter unicode61");
 """
 
 
@@ -80,7 +80,8 @@ class Storage:
         with self.conn:
             # Clear and rewrite (simple strategy for small networks)
             self.conn.execute("DELETE FROM justifications")
-            self.conn.execute("DELETE FROM nodes_fts")
+            self.conn.execute("DROP TABLE IF EXISTS nodes_fts")
+            self.conn.execute("CREATE VIRTUAL TABLE nodes_fts USING fts5(id, text, tokenize='porter unicode61')")
             self.conn.execute("DELETE FROM nodes")
             self.conn.execute("DELETE FROM nogoods")
             self.conn.execute("DELETE FROM repos")
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -271,9 +271,10 @@ class TestFtsSearch:
         assert "a" in result
 
     def test_porter_stemming_plural(self, db_path):
+        from reasons_lib.api import _fts_search
         api.add_node("a", "max 250 jobs per pipeline", db_path=db_path)
-        result = api.search("job", db_path=db_path)
-        assert "a" in result
+        results = _fts_search("job", db_path)
+        assert "a" in results
 
     def test_progressive_relaxation(self, db_path):
         api.add_node("a", "sandbox access expires after 21 days", db_path=db_path)
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -280,6 +280,11 @@ class TestFtsSearch:
         result = api.search("sandbox access duration expiration", db_path=db_path)
         assert "a" in result
 
+    def test_two_term_no_relaxation(self, db_path):
+        api.add_node("a", "sandbox access expires", db_path=db_path)
+        result = api.search("sandbox quantum", db_path=db_path, format="compact")
+        assert "a" not in result
+
     def test_no_false_positives(self, db_path):
         api.add_node("a", "the quick brown fox", db_path=db_path)
         result = api.search("quantum computing blockchain", db_path=db_path, format="compact")
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -263,6 +263,29 @@ class TestListNodesDepth:
         assert ids == ["mid"]
 
 
+class TestFtsSearch:
+
+    def test_porter_stemming(self, db_path):
+        api.add_node("a", "sandbox access is auto-deactivated after 21 days", db_path=db_path)
+        result = api.search("deactivation", db_path=db_path)
+        assert "a" in result
+
+    def test_porter_stemming_plural(self, db_path):
+        api.add_node("a", "max 250 jobs per pipeline", db_path=db_path)
+        result = api.search("job", db_path=db_path)
+        assert "a" in result
+
+    def test_progressive_relaxation(self, db_path):
+        api.add_node("a", "sandbox access expires after 21 days", db_path=db_path)
+        result = api.search("sandbox access duration expiration", db_path=db_path)
+        assert "a" in result
+
+    def test_no_false_positives(self, db_path):
+        api.add_node("a", "the quick brown fox", db_path=db_path)
+        result = api.search("quantum computing blockchain", db_path=db_path, format="compact")
+        assert "a" not in result
+
+
 class TestListGated:
 
     def test_no_gates(self, db_path):


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#73
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/storage.py:SCHEMA (line 60)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Changing the FTS5 tokenizer from `unicode61` to `porter unicode61` directly addresses root cause #1 from the issue. The `CREATE VIRTUAL TABLE IF NOT EXISTS` in SCHEMA means **existing databases that already have a non-porter `nodes_fts` table will silently keep the old tokenizer** on `_init_schema()`. However, the DROP+CREATE in `save()` handles this — any database that gets saved will get the new tokenizer. New databases get it from SCHEMA. This is acceptable given the "clear and rewrite" strategy.

### reasons_lib/storage.py:Storage.save (lines 83-84)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The DROP+CREATE approach is necessary because FTS5 virtual tables cannot be `ALTER`-ed to change the tokenizer. This is the right call. However, the `DROP TABLE IF EXISTS nodes_fts` happens **before** `DELETE FROM nodes` (line 86). Since `nodes_fts` contains the FTS index, dropping it before clearing `nodes` is fine operationally — but if any error occurs between the DROP and the subsequent INSERTs, the table still exists (recreated on line 84) so recovery is clean. One minor concern: the `DROP` + `CREATE` happens inside `with self.conn:` (a transaction context), but SQLite DDL on virtual tables may trigger an implicit commit depending on the driver. In practice, CPython's `sqlite3` module handles this correctly within the `with` block, so this works. Note the quote-style inconsistency: SCHEMA uses `tokenize="porter unicode61"` (double quotes) while `save()` uses `tokenize='porter unicode61'` (single quotes). Both are valid SQL, but the inconsistency is cosmetic noise.

### reasons_lib/api.py:_fts_query (lines 1289-1295)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean extraction of the FTS query execution into a reusable helper. Quoting each term prevents FTS5 syntax injection (e.g., a term containing `OR` or `NOT`). The LIMIT 20 is carried over from the original. Good separation of concerns.

### reasons_lib/api.py:_fts_search (lines 1298-1322)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The progressive relaxation logic works but has two issues worth noting:

**1. Relaxation boundary (`len(terms) > 2`):** With 3 terms, the range is `range(2, max(0, 0), -1)` = `range(2, 0, -1)` = `[2, 1]`, meaning it tries 2-term combos and 1-term combos. With 4 terms: `range(3, max(0, 1), -1)` = `range(3, 1, -1)` = `[3, 2]`. This means for 4+ terms, it drops up to 2 terms, which matches the issue's proposal (`len(terms) - 2`). For exactly 3 terms, it can drop down to 1 term, which is more aggressive than what the issue proposed but is arguably correct behavior. The `len(terms) > 2` threshold means 2-term queries never get relaxed — the test `test_two_term_no_relaxation` validates this explicitly.

**2. Combinatorial cost:** With many terms, `combinations(terms, n)` generates `C(k, n)` queries. For a 10-term query dropping 2: `C(10,9) + C(10,8) = 10 + 45 = 55` FTS queries. Each is a SQLite query, so this is unlikely to be a real perf issue in practice, but there's no upper bound on query term count. The `reasons ask` LLM typically generates 3-6 terms, so this is fine in practice.

**3. The issue's fix #3 (suppress hallucination on miss) is not addressed.** The issue explicitly calls this out as a separate fix, so this is expected — but worth noting that 1 of 3 proposed fixes is not in this PR.

The `try/finally` for connection cleanup is a clear improvement over the original's multi-path `conn.close()` calls.

### tests/test_api.py:TestFtsSearch (lines 266-292)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Good tests covering the core scenarios: stemming ("deactivation" → "deactivated"), plural stemming ("job" → "jobs"), progressive relaxation with 4 terms, 2-term no-relaxation boundary, and false-positive guard. Missing test cases:

1. **No test for the exact reproduction cases from the issue** — "Models.corp sandbox access deactivation" or "GitLab pipeline jobs maximum limit". These were the motivating failures and would serve as regression tests.
2. **No test for single-term queries** — verifies the `if not terms: return []` path.
3. **No test for existing databases without porter stemming** — would verify the migration path works (old FTS table replaced on save).
4. **The stemming tests check `"a" in result`** where `result` is a string (the formatted markdown output). This works because `"a"` appears as a node ID in the output, but it's fragile — if the output format changed to not include the bare ID, these assertions would break. The `test_two_term_no_relaxation` and `test_no_false_positives` tests wisely use `format="compact"` but the stemming tests don't, creating an inconsistency. Both approaches rely on string containment, which is reasonable for the `search` API that returns formatted strings.

### Self-Review
**Limitations:** Could not verify behavior against an actual existing database with the old tokenizer (no migration test). Did not run the test suite to confirm all 211+ tests pass with the changes. Could not see the `_with_network` context manager to verify it doesn't interact with FTS in a way that conflicts with the separate `_fts_search` connection.

### Feature Requests
- Include `git diff --stat` to see the full scope of changed files at a glance
- Show the full test fixture (`db_path` conftest) to verify test isolation guarantees
- Auto-run the test suite and include results in the review context

## gemini [CONCERN]

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `_fts_query` helper is cleanly extracted, and the progressive relaxation logic correctly tries combinations by dropping 1 or 2 terms for queries with 3 or more terms. Extracting `sqlite3.OperationalError` prevents crashes if malformed quotes ever slip into the FTS syntax.

### reasons_lib/storage.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Dropping and recreating the `nodes_fts` table during the `save()` transaction correctly guarantees the `porter unicode61` tokenizer is applied, ensuring backwards compatibility and automatic migrations for existing SQLite stores.

### tests/test_api.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

There are two distinct concerns with the test implementation:
1. `test_porter_stemming_plural` asserts that a query for "job" finds the node containing "jobs". However, because `api.search()` falls back to `_substring_search()` on an FTS miss, and "job" is a direct substring of "jobs", this test would pass *even without* the new FTS stemming fix. It provides false confidence.
2. Asserting `"a" in result` against string-formatted markdown is slightly flaky. Because "a" is a single character, it could easily be a substring of standard text (e.g., "No matching nodes"). It would be safer to call `_fts_search()` directly or use `format="json"`/`format="compact"` (as seen in the negative tests) to parse deterministic structures.

### Self-Review
**Limitations:** I could not see the actual text output of `_format_markdown`, meaning I had to deduce the potential risk of string-matching a single character `"a"` against a generic failure response. I also did not review prompt changes for the 3rd issue ("Suppress hallucination on miss"), assuming it's handled in a separate agent PR outside of the `reasons_lib` module.

### Feature Requests
- Include the full file context rather than just diff hunks. Relying on surrounding context (like seeing `_substring_search` right below `_fts_search`) was critical here for diagnosing the flawed test.
- When tests invoke a function that delegates to a formatter (like `_format_markdown`), including a sample of the formatted output in the observations would help assess test robustness.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#73
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/storage.py (SCHEMA change — line 60)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The FTS5 table definition correctly adds `tokenize="porter unicode61"` which enables porter stemming. The `IF NOT EXISTS` clause means existing databases that already have a `nodes_fts` table without porter stemming will **not** be upgraded on `_init_schema()` — but this is mitigated by the `save()` method which drops and recreates the table. For databases that are only ever read (never saved), the old non-stemmed table would persist. This is an acceptable tradeoff given the project's "clear and rewrite" save strategy.

### reasons_lib/storage.py:Storage.save (DROP + CREATE pattern — lines 83–84)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Replacing `DELETE FROM nodes_fts` with `DROP TABLE IF EXISTS` + `CREATE VIRTUAL TABLE` is necessary because you can't change the tokenizer on an existing FTS5 table. However, the DROP happens inside a `with self.conn:` transaction block. FTS5 virtual table DDL (DROP/CREATE) within an explicit transaction can behave unexpectedly depending on SQLite version — in most modern versions it works, but it's worth noting this is a subtle change from the prior DML-only pattern. The quoting also switches from double-quotes (`tokenize="porter unicode61"`) in the schema to single-quotes (`tokenize='porter unicode61'`) in `save()`. Both work in SQLite, but the inconsistency is cosmetic noise. No functional bug, but the DDL-in-transaction pattern is worth flagging.

### reasons_lib/api.py:_fts_query (new function — lines 1289–1295)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean extraction of the FTS5 query execution into a reusable helper. The function correctly quotes each term to avoid FTS5 syntax injection. `ORDER BY rank LIMIT 20` is consistent with the prior implementation. It takes the connection as a parameter, enabling the connection reuse needed by progressive relaxation.

### reasons_lib/api.py:_fts_search (modified — lines 1298–1322)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The progressive relaxation logic has a boundary condition worth examining. The guard `len(terms) > 2` means relaxation only kicks in for 3+ terms. The issue's proposed fix says `max(1, len(terms) - 2)` as the lower bound of the range, but the implementation uses `max(0, len(terms) - 3)`. Let me trace through:

- For 4 terms: `range(3, max(0, 1), -1)` → tries combos of 3, 2, 1 terms. That drops up to 3 terms — quite aggressive.
- For 3 terms: `range(2, max(0, 0), -1)` → tries combos of 2, 1 terms. Drops up to 2 terms.
- For 5 terms: `range(4, max(0, 2), -1)` → tries combos of 4, 3 terms. Drops up to 2 terms.

Wait — `max(0, len(terms) - 3)` for 4 terms is `max(0, 1) = 1`, so `range(3, 1, -1)` → tries 3, 2 (not 1). For 3 terms: `max(0, 0) = 0`, so `range(2, 0, -1)` → tries 2, 1. So with 3 terms, a single-term combo can match, which is quite loose and could return false positives. The issue proposed dropping "one at a time" up to 2 terms max, suggesting `max(1, ...)` as a floor. The implementation allows dropping down to single-term queries for 3-term inputs, which may produce noisy results. This is a **functional concern** but not a blocker — the `_substring_search` fallback already exists as a catch-all, so false positives from relaxation aren't catastrophic.

The `try/finally` pattern for connection cleanup is a clear improvement over the old code which could leak connections on exceptions within the query logic.

One missing aspect from the issue: **Fix #3 (suppress hallucination on miss)** is not addressed in this PR. The issue says "separately" so this may be intentional, but it's worth noting.

### tests/test_api.py:TestFtsSearch (new test class — lines 266–292)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Five tests cover the key scenarios well:

1. `test_porter_stemming` — verifies "deactivation" matches "deactivated" via the public `search()` API. Good.
2. `test_porter_stemming_plural` — verifies "job" matches "jobs" via `_fts_search` directly. Good, and appropriate to test the internal function since `search()` has a substring fallback that could mask FTS failures.
3. `test_progressive_relaxation` — verifies that a 4-term query with 2 missing terms ("duration", "expiration") still finds the result. Good.
4. `test_two_term_no_relaxation` — verifies 2-term queries don't relax. Good boundary test.
5. `test_no_false_positives` — verifies completely unrelated terms return nothing. Good.

**Missing coverage:**
- No test for the exact boundary: 3 terms where only 1 matches (the loosest relaxation case).
- No test for the migration path: opening an existing DB with a non-porter `nodes_fts` table.
- No test that `_fts_search` returns results in rank order.
- `test_porter_stemming` asserts `"a" in result` where `result` is a markdown string — this works because node ID "a" appears in the markdown output, but it's fragile. A change to the markdown format could break this test. `test_porter_stemming_plural` correctly tests via `_fts_search` which returns a list of IDs.

### Self-Review
**Limitations:** Could not verify whether existing test suites (test_storage.py, test_search in test_api.py) still pass with the schema change — the FTS table tokenizer change could affect existing search tests if they depend on exact non-stemmed matching. Could not run the test suite to confirm. Did not have access to the full `search()` function to verify all format paths. The `test_porter_stemming` test's `"a" in result` assertion against a markdown string is a reliability concern I could only partially assess without seeing all format output patterns.

### Feature Requests
- Include the full test file (or at least the conftest/fixtures) to verify `db_path` fixture creates fresh databases with the new schema
- Show `git diff --stat` to confirm no other files were modified that should have been included
- Include the output of `pytest` run against the changed tests to verify they actually pass

## gemini [CONCERN]

### `reasons_lib/api.py`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The progressive relaxation logic executes multiple queries sequentially and returns on the very first match (`if results: return results`). Because `itertools.combinations` yields combinations in lexicographical order, this bypasses FTS5 ranking and arbitrarily favors the first evaluated combination over potentially better matches in other combinations of the same length. A better approach would be combining same-length combinations into a single FTS5 `OR` query to preserve `BM25` ranking. Additionally, the PR entirely omits the requested "Suppress hallucination on miss" fix (Point 3 in the issue).

### `reasons_lib/storage.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly adds the `porter` stemmer to the FTS virtual table definition. Dropping and recreating the FTS table in `save()` is a solid approach to ensure that previously created local databases will rebuild their indexes with the new stemmer on their next save.

### `tests/test_api.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good test coverage validating the core fixes: stemming (including plural edge cases), progressive relaxation succeeding on partial matches, and verifying that short queries (2 terms) do not over-relax.

### Self-Review
**Limitations:** I could not view the complete `save()` method in `reasons_lib/storage.py` to definitively confirm how rows are inserted back into `nodes_fts` after the table is dropped/recreated, though the passing tests imply it is handled correctly. I also did not have visibility into `reasons_lib/ask.py` to check if the hallucination suppression prompt was already added in another commit.

PR #80, opened 2026-05-01
Suppress hallucination in reasons ask when no beliefs match

### LINKED ISSUE #74: reasons ask: suppress hallucination when no beliefs found
## Summary

When `reasons ask` finds no matching beliefs, the LLM synthesis prompt should strongly instruct the model to say "not found" rather than answering from training data. Currently the model fabricates answers — e.g. Q4 in the redhat-expert evaluation said "200 jobs" (generic GitLab default) when the actual belief says 250.

## Context

Split from #72 (fix #3). The FTS stemming and relaxation fixes in PR #73 reduce the likelihood of zero-result searches, but when they do occur the synthesis prompt needs a stronger guard.

## Proposed fix

In `reasons_lib/ask.py`, update the synthesis prompt to explicitly instruct: if no beliefs were found or the search results do not contain the answer, respond with a clear "I could not find this information in the belief network" rather than answering from general knowledge.

benthomasson:
## Summary
- Adds early exit in `ask()` when FTS search returns no results — skips LLM invocation entirely and returns a clear "No matching beliefs found" message
- Strengthens `ASK_PROMPT` rules to explicitly forbid answering from training data and require a specific refusal response
- Adds 3 new tests for the no-beliefs early exit path; fixes 3 existing tests that used non-matching queries

## Test plan
- [x] `uv run pytest tests/test_ask.py -v` — 28 passed
- [x] `uv run pytest tests/ -v` — 642 passed, 106 skipped
- [ ] Manual: `reasons ask "something with no matching beliefs"` returns canned message

Closes #74

### DIFF
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -166,6 +166,8 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
             print(f"  Searching: {query}", file=sys.stderr)
             result = api.search(query, db_path=db_path, format="markdown")
             tool_history.append({"query": query, "result": result})
+            if result and result.strip() != "No results found.":
+                beliefs_context = result
         else:
             return response.strip()
 
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -119,6 +119,12 @@ MAX_ITERATIONS = 3
 NO_BELIEFS_MSG = "No matching beliefs found. Cannot answer from the belief network."
 
 
+def _beliefs_or_no_match(beliefs_context):
+    if not beliefs_context or beliefs_context.strip() == "No results found.":
+        return NO_BELIEFS_MSG
+    return beliefs_context
+
+
 def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None):
     """Answer a question using FTS5 belief search and optional LLM synthesis.
 
@@ -129,9 +135,6 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
         return api.search(question, db_path=db_path, format=fmt)
 
     beliefs_context = api.search(question, db_path=db_path, format="markdown")
-
-    if not beliefs_context or beliefs_context.strip() == "No results found.":
-        return NO_BELIEFS_MSG
 
     tool_history = []
 
@@ -148,10 +151,10 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
             response = _invoke_claude(prompt, timeout=timeout)
         except subprocess.TimeoutExpired:
             print(f"LLM timed out after {timeout}s", file=sys.stderr)
-            return beliefs_context
+            return _beliefs_or_no_match(beliefs_context)
         except Exception as e:
             print(f"LLM error: {e}", file=sys.stderr)
-            return beliefs_context
+            return _beliefs_or_no_match(beliefs_context)
 
         tool_call = extract_tool_call(response)
 
@@ -166,4 +169,4 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
         else:
             return response.strip()
 
-    return beliefs_context
+    return _beliefs_or_no_match(beliefs_context)
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -29,7 +29,11 @@ Rules:
 - If you need to search for more beliefs, respond with ONLY a single JSON line
   (no other text). The system will run the search and give you the results.
 - Cite belief IDs in [brackets] when referencing specific beliefs.
-- If the beliefs are insufficient to answer, say so honestly.
+- ONLY answer based on the beliefs provided. Do NOT use your training data or
+  general knowledge to fill gaps.
+- If the beliefs are insufficient to answer, respond EXACTLY with:
+  "I don't have enough beliefs in the network to answer this question."
+  Do NOT attempt a partial or speculative answer.
 
 ## Question
 
@@ -112,6 +116,8 @@ def _invoke_claude(prompt, timeout=300):
 
 MAX_ITERATIONS = 3
 
+NO_BELIEFS_MSG = "No matching beliefs found. Cannot answer from the belief network."
+
 
 def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None):
     """Answer a question using FTS5 belief search and optional LLM synthesis.
@@ -123,6 +129,9 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
         return api.search(question, db_path=db_path, format=fmt)
 
     beliefs_context = api.search(question, db_path=db_path, format="markdown")
+
+    if not beliefs_context or beliefs_context.strip() == "No results found.":
+        return NO_BELIEFS_MSG
 
     tool_history = []
 
--- a/tests/test_ask.py
+++ b/tests/test_ask.py
@@ -194,6 +194,23 @@ class TestAskNoBeliefs:
             result = ask("nothing matches", db_path=db_path)
         assert result == NO_BELIEFS_MSG
 
+    def test_timeout_after_successful_tool_search_returns_beliefs(self, db_path):
+        run_cli("init", db_path=db_path)
+        run_cli("add", "b", "Beta belief about retraction", db_path=db_path)
+
+        calls = [0]
+
+        def mock_invoke(prompt, timeout=300):
+            calls[0] += 1
+            if calls[0] == 1:
+                return '{"tool": "search_beliefs", "query": "retraction"}'
+            raise subprocess.TimeoutExpired("claude", 300)
+
+        with patch("reasons_lib.ask._invoke_claude", side_effect=mock_invoke):
+            result = ask("zzzznothing", db_path=db_path)
+        assert "retraction" in result.lower()
+        assert result != NO_BELIEFS_MSG
+
 
 class TestAskWithMockedLLM:
 
--- a/tests/test_ask.py
+++ b/tests/test_ask.py
@@ -165,22 +165,33 @@ class TestInvokeClaude:
 
 class TestAskNoBeliefs:
 
-    def test_empty_network_returns_no_beliefs_message(self, db_path):
+    def test_empty_network_llm_declines(self, db_path):
         run_cli("init", db_path=db_path)
-        result = ask("what is the meaning of life", db_path=db_path)
-        assert result == NO_BELIEFS_MSG
+        refusal = "I don't have enough beliefs in the network to answer this question."
+        with patch("reasons_lib.ask._invoke_claude", return_value=refusal):
+            result = ask("what is the meaning of life", db_path=db_path)
+        assert "don't have enough beliefs" in result
 
-    def test_no_matching_beliefs_returns_no_beliefs_message(self, db_path):
+    def test_no_matching_beliefs_llm_declines(self, db_path):
         run_cli("init", db_path=db_path)
         run_cli("add", "alpha", "Alpha belief about propagation", db_path=db_path)
-        result = ask("zzzznonexistent", db_path=db_path)
+        refusal = "I don't have enough beliefs in the network to answer this question."
+        with patch("reasons_lib.ask._invoke_claude", return_value=refusal):
+            result = ask("zzzznonexistent", db_path=db_path)
+        assert "don't have enough beliefs" in result
+
+    def test_timeout_on_empty_returns_no_beliefs_message(self, db_path):
+        run_cli("init", db_path=db_path)
+        with patch("reasons_lib.ask._invoke_claude",
+                    side_effect=subprocess.TimeoutExpired("claude", 300)):
+            result = ask("nothing matches", db_path=db_path)
         assert result == NO_BELIEFS_MSG
 
-    def test_no_llm_invoked_on_empty_results(self, db_path):
+    def test_error_on_empty_returns_no_beliefs_message(self, db_path):
         run_cli("init", db_path=db_path)
-        with patch("reasons_lib.ask._invoke_claude") as mock_claude:
+        with patch("reasons_lib.ask._invoke_claude",
+                    side_effect=RuntimeError("claude crashed")):
             result = ask("nothing matches", db_path=db_path)
-        mock_claude.assert_not_called()
         assert result == NO_BELIEFS_MSG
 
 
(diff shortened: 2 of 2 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#80
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [PASS]

### `reasons_lib/ask.py` — Prompt hardening (lines 32–36)
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: ADDRESSES
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: COVERED (prompt content verified by `TestBuildAskPrompt.test_tool_definition_in_prompt` and indirectly by the synthesis tests)
INTEGRATION: WIRED
REASONING: The old prompt said "say so honestly" — weak. The new prompt explicitly forbids training-data answers and gives a mandatory fallback phrase. This is the in-prompt guard the issue requests. The wording is clear and specific enough to steer most models.

---

### `reasons_lib/ask.py:ask` — Early return on empty results (lines 119, 133–134)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The guard checks both falsy `beliefs_context` (empty string / None) and the exact sentinel `"No results found."` returned by `api.search`. I verified `api.search` returns exactly that string at `api.py:1251` and `api.py:1260`. This short-circuits before any LLM invocation, which is the core fix — no beliefs means no Claude call, no hallucination. The `NO_BELIEFS_MSG` constant is cleanly exported and tested.

### `tests/test_ask.py::TestAskNoBeliefs`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Three tests cover the essential scenarios: (1) empty network, (2) beliefs exist but don't match the query, (3) mock-verifying `_invoke_claude` is never called. This is thorough coverage of the new guard. The mock assertion in `test_no_llm_invoked_on_empty_results` is particularly important — it proves the hallucination vector is closed, not just that the return value is correct.

### `tests/test_ask.py` — Existing test query adjustments (lines 214, 220, 226, 249, 253)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Queries like `"question"` and `"how does retraction work?"` were changed to `"retraction"`, `"alpha"`, etc. This is a necessary consequence of the new early-return guard — queries that don't FTS-match any belief text now return `NO_BELIEFS_MSG` instead of reaching the mocked LLM. The belief text was also changed from `"Alpha"` to `"Alpha belief"` in some places to ensure FTS matches. These are correct adaptations, not test weakening — they preserve the original intent (testing tool-call loops, max iterations, error fallbacks) while ensuring the query actually reaches the LLM path.

### Self-Review
**Limitations:** Could not run the test suite to confirm all tests pass. Reviewed correctness by reading source only. Did not verify whether the CLI `ask` subcommand handles the `NO_BELIEFS_MSG` return differently (e.g., exit code) — checked only the library path. Did not inspect whether `no_synth=True` path should also return the new message instead of the raw "No results found." string (current behavior is intentional — `no_synth` is a raw search mode).

### Feature Requests
- Show the full `api.search` return values / contract documentation alongside the diff when the guard condition depends on a specific sentinel string
- Include CLI integration tests (if any exist) that exercise the `ask` command end-to-end

## gemini [BLOCK]

### reasons_lib/ask.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

While the synthesis prompt was updated as requested by the issue, adding the short-circuit (`if not beliefs_context or beliefs_context.strip() == "No results found.": return NO_BELIEFS_MSG`) introduces a significant regression and breaks the agent's core functionality. By returning immediately when the initial FTS search (using the raw natural language question) yields no results, the LLM is completely bypassed. This prevents the LLM from ever seeing the zero results and utilizing the `search_beliefs` tool to retry the search with better, extracted keywords. The issue explicitly requested updating the synthesis prompt to handle zero results, not short-circuiting the tool loop.

### tests/test_ask.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The tests were modified to accommodate the flawed short-circuit logic introduced in `ask.py`. Changing test inputs from natural language questions (e.g., "how does retraction work?") to single keywords ("retraction", "alpha") was necessary just to get past the new short-circuit and invoke the LLM. This clearly demonstrates that natural language questions are now failing to reach the LLM if they don't have an exact initial match, masking the regression. The new `TestAskNoBeliefs` tests also enshrine this broken behavior.

### Self-Review
**Limitations:** None. The diff and tests provided clear evidence of how the short-circuiting logic degrades the existing agent behavior.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#80
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/ask.py:ASK_PROMPT
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The prompt change replaces the weak "say so honestly" instruction with an explicit prohibition on using training data and a required verbatim refusal phrase. This directly addresses the issue — the Q4 hallucination ("200 jobs" from training data when the answer was 250) happened because the old prompt allowed the LLM to fill gaps from general knowledge. The new wording is strong and specific.

### reasons_lib/ask.py:_beliefs_or_no_match
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean helper that checks both empty/None `beliefs_context` and the exact `"No results found."` string returned by `api.search`. The string comparison at line 123 matches `api.py:1251` exactly. Applied correctly in all three fallback paths (timeout at line 154, error at line 157, loop exhaustion at line 172). Previously these paths returned raw `beliefs_context`, which could be the empty/unhelpful `"No results found."` string — now the user sees the clear `NO_BELIEFS_MSG` instead.

### reasons_lib/ask.py:ask (fallback paths)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The three fallback paths now correctly use `_beliefs_or_no_match`. However, there's a subtle gap: when the initial `beliefs_context` is `"No results found."` but the LLM issues a `search_beliefs` tool call that *does* find results (via a better query), those results go into `tool_history` but not into `beliefs_context`. If the LLM then times out or errors on the second iteration, the fallback still returns `NO_BELIEFS_MSG` even though relevant beliefs were found via the tool call. This is an edge case — the LLM would have to succeed once (to issue the tool call), the re-search would have to find results, and then the next LLM call would have to fail. Low probability but worth noting since the "let LLM retry searches" commit (cc74f14) specifically enables this scenario. A test for "timeout after successful tool search" would strengthen coverage.

### tests/test_ask.py:TestAskNoBeliefs
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Four new tests cover the key scenarios well: empty network with LLM declining (line 168), beliefs exist but don't match query with LLM declining (line 175), timeout on empty network (line 183), and error on empty network (line 190). The timeout/error tests verify the exact `NO_BELIEFS_MSG` constant. The LLM-decline tests verify the prompt works by asserting the refusal phrase passes through. The existing `test_timeout_returns_search_results` (line 240) and `test_error_returns_search_results` (line 249) still pass because they add a belief that matches ("Alpha belief" searched with "alpha"), so `_beliefs_or_no_match` returns the belief context — good complementary coverage.

### Self-Review
**Limitations:** I verified the `api.search` return value for no-results but did not trace the full FTS5 search path to confirm there are no other possible empty-result return values (e.g., empty string vs. None vs. whitespace-only). The `_beliefs_or_no_match` function only checks for falsy and the exact string `"No results found."` — if `api.search` could return something like `"No results found.\n"` with trailing whitespace, the `.strip()` comparison handles it. Could not verify whether the cc74f14 commit introduces retry logic that changes the flow materially.

### Feature Requests
- Include git log context for referenced commits (e.g., cc74f14) so the reviewer can verify interaction between PRs
- Show the full function body of `api.search` and callers of modified functions to verify edge cases

## gemini [PASS]

### reasons_lib/ask.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The prompt is updated to strongly forbid hallucination and require an exact refusal string when beliefs are insufficient, addressing the issue perfectly. The new `_beliefs_or_no_match` helper improves the fallback logic on LLM errors/timeouts to return a clear message (`NO_BELIEFS_MSG`) rather than raw search output. Note that this helper does *not* short-circuit the execution flow; the LLM is correctly invoked even when the initial search yields no results, preserving the agent's ability to use the search tool and retry with different keywords. The code is safe and logically sound.

### tests/test_ask.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New test cases effectively cover the expected behaviors: they verify that the LLM's exact refusal string is correctly propagated when no beliefs match or the network is empty, and they test both the timeout and error branches to ensure `NO_BELIEFS_MSG` is returned as a fallback.

### Self-Review
**Limitations:** None. Context was complete and code logic is clear.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#80
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/ask.py:ASK_PROMPT
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The prompt changes directly address the issue. The old instruction ("say so honestly") was too weak — the new language explicitly forbids training-data answers and mandates a specific refusal phrase. The two new rules are clear and non-contradictory. Tests in `TestAskNoBeliefs` verify that LLM responses matching the refusal pattern flow through correctly.

### reasons_lib/ask.py:_beliefs_or_no_match
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function is correct for its intended purpose, but it relies on matching the magic string `"No results found."` which is hardcoded in `api.py` (lines 1251, 1260, 1421). The same string appears in two places in `ask.py` (lines 123 and 169) without a shared constant. If `api.py` ever changes this string, `ask.py` will silently fail to detect no-results and pass through raw "no results" text to the user. Consider extracting a constant like `NO_RESULTS_SENTINEL` in `api.py` and importing it. Low severity since it all lives in one codebase, but it's the kind of coupling that bites during refactors.

### reasons_lib/ask.py:ask (beliefs_context preservation on retry, line 169-170)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

This is a subtle and important fix. Previously, if the LLM requested a tool search that returned nothing, `beliefs_context` would implicitly still hold its old value (since `tool_history.append` was the only action), but the fallback paths would return whatever `beliefs_context` was — which could be the original good results. The new explicit guard (`if result and result.strip() != "No results found.": beliefs_context = result`) makes this intent clear: only overwrite `beliefs_context` when the retry actually found something. The test `test_retry_no_results_preserves_initial_beliefs` directly validates this path.

### reasons_lib/ask.py:ask (timeout/error fallbacks, lines 154, 157, 174)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All three fallback paths (timeout, generic error, end-of-loop) now use `_beliefs_or_no_match()` instead of returning raw `beliefs_context`. This means empty/no-results context produces a clear message instead of an empty string or "No results found." Tests cover timeout-on-empty, error-on-empty, and timeout-after-successful-retry.

### tests/test_ask.py:TestAskNoBeliefs
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Six well-designed tests covering: (1) empty network with LLM refusal, (2) non-matching query with LLM refusal, (3) timeout on empty network, (4) error on empty network, (5) timeout after successful tool retry preserves beliefs, (6) retry returning no results preserves initial beliefs. Tests 5 and 6 are particularly valuable — they use a stateful mock to simulate the multi-iteration loop with a successful first search followed by a failure, verifying the `beliefs_context` preservation logic. The import of `NO_BELIEFS_MSG` for assertions is clean.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Could not see the full `api.search()` function to verify all possible return values are handled (only checked the "No results found." paths via grep). Did not check `pg.py` integration paths that also use the same sentinel string.

### Feature Requests
- Include the full function body for `api.search()` in observations, since the fallback logic depends on its return values
- Flag magic string duplication across files automatically

## gemini [CONCERN]

### reasons_lib/ask.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The updated prompt correctly instructs the LLM to refuse to answer from general knowledge, which cleanly addresses the core issue. However, there is a likely logic bug introduced in the tool-call handling: overwriting `beliefs_context` with the result of a successful tool search (`beliefs_context = result`). Assuming the LLM prompt is rebuilt on each iteration of the loop, replacing `beliefs_context` means the LLM will lose access to the initial search results in subsequent turns, seeing only the latest tool search result and the stringified `tool_history`. To return the latest result as a fallback on timeouts/errors without destroying the prompt context, you should store the best fallback in a separate variable (e.g., `fallback_context = result`) and use that in the `except` blocks.

### tests/test_ask.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The tests thoroughly cover the new fallback behaviors, handling empty networks, non-matching searches, timeouts, and exception states. They properly verify that the LLM declines and the fallback mechanism returns the expected "not found" message.

### Self-Review
**Limitations:** I could not see the full `ask` function or `build_ask_prompt` to definitively confirm that `beliefs_context` is passed into `build_ask_prompt` on every iteration. However, standard iterative LLM tool-use loops reconstruct the prompt in each turn, making the overwrite of `beliefs_context` highly likely to be a bug that drops initial context.

### Feature Requests
- Include full file context for modified functions, not just diff hunks, to allow verifying how modified variables (like `beliefs_context`) are used across the entire function scope.

PR #81, opened 2026-05-01
Return beliefs context instead of raw tool call on final iteration

### LINKED ISSUE #79: reasons ask returns raw tool call instead of answer on final iteration
## Bug

`reasons ask` occasionally returns a raw JSON tool call (e.g. `{"tool": "search_beliefs", "query": "..."}`) as the answer instead of completing synthesis. This happens when the LLM response on the final iteration contains only a tool call with no surrounding text.

## Evidence

Across 991 evaluation questions (691 product + 299 engineering), 5 answers (0.5%) were raw tool calls instead of synthesized answers. All 5 were scored F by an LLM-as-judge because no useful answer was provided to the user.

Examples:
- `engineering_pe_hcm_85f756be` — "What is the starting point for everything delivered by the HCM organization?"
- `engineering_xe_cap_bca23294` — "What is the purpose of the Phased Gates program within CAP's preemptive measures?"
- `engineering_xe_uxe_b9d72999` — "What specific forum is used by UXE engineers to discuss best practices?"
- `product_competitive_8475ab2d` — "What is the 'Key Play' associated with the Nutanix Data Lens Data Sheet?"
- `product_openshift_competitive_eeb3c7f6` — "Which Azure AI service is mentioned as being used by Manulife alongside AKS?"

## Root Cause

In `reasons_lib/ask.py`, the main loop at line 129:

```python
for iteration in range(MAX_ITERATIONS):
    ...
    response = _invoke_claude(prompt, timeout=timeout)
    tool_call = extract_tool_call(response)

    if tool_call is None or iteration == MAX_ITERATIONS - 1:
        return response.strip()
```

On the final iteration (`iteration == MAX_ITERATIONS - 1`), the code returns `response.strip()` regardless of whether it contains a tool call. If the LLM emitted only a tool call JSON line on the final round, that raw JSON is returned as the "answer."

The `FINAL_TURN_INSTRUCTION` appended on the last iteration ("write your answer now, no more tool calls") should prevent this, but the LLM occasionally ignores it.

## Suggested Fix

On the final iteration, if the response is only a tool call (no other text), execute the search and make one more synthesis call with the results — or return the accumulated belief context instead of the raw tool call.

```python
if iteration == MAX_ITERATIONS - 1:
    if tool_call and response.strip().startswith("{"):
        # LLM ignored final-turn instruction — return beliefs context
        return beliefs_context
    return response.strip()
```

benthomasson:
## Summary
- Fixes bug where `reasons ask` returns raw JSON tool call as the answer when the LLM ignores the final-turn instruction
- Splits the final-iteration condition: if no tool call, return response; if tool call on last iteration, return beliefs context
- Affected 0.5% of evaluation questions (5 out of 991)

## Test plan
- [x] `uv run pytest tests/test_ask.py -v` — 31 passed
- [x] `uv run pytest tests/ -v` — 645 passed, 106 skipped

Closes #79

### DIFF
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -45,11 +45,6 @@ Rules:
 {tool_history}"""
 
 
-FINAL_TURN_INSTRUCTION = (
-    "\n\n**Final turn — write your answer now, no more tool calls.**"
-)
-
-
 FINAL_ASK_PROMPT = """\
 You are answering a question using a belief network (a Truth Maintenance System).
 Each belief has an ID, text, truth value (IN = held true, OUT = retracted), and
@@ -219,7 +214,13 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
             prompt = build_final_prompt(question, beliefs_context, tool_history)
             try:
                 response = _invoke_claude(prompt, timeout=timeout)
-            except (subprocess.TimeoutExpired, Exception):
+            except subprocess.TimeoutExpired:
+                print(f"LLM timed out after {timeout}s", file=sys.stderr)
+                return _beliefs_or_no_match(beliefs_context)
+            except Exception as e:
+                print(f"LLM error: {e}", file=sys.stderr)
+                return _beliefs_or_no_match(beliefs_context)
+            if extract_tool_call(response):
                 return _beliefs_or_no_match(beliefs_context)
             return response.strip()
 
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -158,8 +158,11 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
 
         tool_call = extract_tool_call(response)
 
-        if tool_call is None or iteration == MAX_ITERATIONS - 1:
+        if tool_call is None:
             return response.strip()
+
+        if iteration == MAX_ITERATIONS - 1:
+            return _beliefs_or_no_match(beliefs_context)
 
         if tool_call.get("tool") == "search_beliefs":
             query = tool_call.get("query", "")
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -50,6 +50,30 @@ FINAL_TURN_INSTRUCTION = (
 )
 
 
+FINAL_ASK_PROMPT = """\
+You are answering a question using a belief network (a Truth Maintenance System).
+Each belief has an ID, text, truth value (IN = held true, OUT = retracted), and
+may have justifications tracing why it is believed.
+
+Rules:
+- Cite belief IDs in [brackets] when referencing specific beliefs.
+- ONLY answer based on the beliefs provided. Do NOT use your training data or
+  general knowledge to fill gaps.
+- If the beliefs are insufficient to answer, respond EXACTLY with:
+  "I don't have enough beliefs in the network to answer this question."
+  Do NOT attempt a partial or speculative answer.
+- Write your answer now.
+
+## Question
+
+{question}
+
+## Belief matches
+
+{beliefs_context}
+{tool_history}"""
+
+
 def extract_tool_call(text):
     """Extract a tool call from LLM response text.
 
@@ -82,6 +106,25 @@ def build_ask_prompt(question, beliefs_context, tool_history=None):
         history_section = "\n\n## Additional search results\n\n" + "\n\n---\n\n".join(parts)
 
     return ASK_PROMPT.format(
+        question=question,
+        beliefs_context=beliefs_context,
+        tool_history=history_section,
+    )
+
+
+def build_final_prompt(question, beliefs_context, tool_history=None):
+    """Build prompt for final synthesis — no tool definition."""
+    history_section = ""
+    if tool_history:
+        parts = []
+        for entry in tool_history:
+            parts.append(
+                f"### Tool call: search_beliefs(\"{entry['query']}\")\n\n"
+                f"{entry['result']}"
+            )
+        history_section = "\n\n## Additional search results\n\n" + "\n\n---\n\n".join(parts)
+
+    return FINAL_ASK_PROMPT.format(
         question=question,
         beliefs_context=beliefs_context,
         tool_history=history_section,
@@ -139,10 +182,10 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
     tool_history = []
 
     for iteration in range(MAX_ITERATIONS):
-        prompt = build_ask_prompt(question, beliefs_context, tool_history)
-
         if iteration == MAX_ITERATIONS - 1:
-            prompt += FINAL_TURN_INSTRUCTION
+            prompt = build_final_prompt(question, beliefs_context, tool_history)
+        else:
+            prompt = build_ask_prompt(question, beliefs_context, tool_history)
 
         print(f"Synthesizing (round {iteration + 1}/{MAX_ITERATIONS})...",
               file=sys.stderr)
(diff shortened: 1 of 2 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#81
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/ask.py:FINAL_ASK_PROMPT
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new `FINAL_ASK_PROMPT` template omits the tool definition (`"tool": "search_beliefs"` JSON example), which is the key mechanism for preventing the LLM from emitting tool calls on the final synthesis turn. The rules section retains all important behavioral instructions (cite IDs, answer from beliefs only, refuse if insufficient). Well-designed — removing the affordance is more reliable than instructing the LLM not to use it.

### reasons_lib/ask.py:build_final_prompt
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Mirrors `build_ask_prompt` exactly but uses `FINAL_ASK_PROMPT` instead of `ASK_PROMPT`. The tool history formatting is duplicated (same logic in both functions), which is a minor DRY concern but acceptable given the functions are small and the duplication is intentional (different templates). Test `test_final_prompt_has_no_tool_definition` verifies the critical property that no tool definition appears.

### reasons_lib/ask.py:ask (loop logic)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The fix has a structural issue. With `MAX_ITERATIONS = 3`, the loop runs `iteration = 0, 1, 2`. On the final iteration (2), the code uses `build_final_prompt` (line 181) which has no tool definition. If the LLM *still* emits a tool call (line 197), the code falls through to the `search_beliefs` handler (line 202-208), executes the search, and then hits the second `iteration == MAX_ITERATIONS - 1` check at line 212. This triggers an extra synthesis call (line 214-225), which is the core fix.

**The concern**: On the final iteration, the code builds a `FINAL_ASK_PROMPT` with no tool definition, yet immediately after also checks for tool calls. If the LLM emits a tool call *despite* no tool being defined in the prompt, the code handles it — which is correct defensive behavior. However, the extra synthesis call at line 212-225 is only reachable when `iteration == MAX_ITERATIONS - 1` AND a `search_beliefs` tool call was extracted. This means the guard at line 212 is never reached for iterations 0 or 1 (they `continue` via the loop), which is fine.

**Edge case**: If the LLM emits an unknown tool on the final iteration (not `search_beliefs`), line 210 returns the raw response — this was true before the fix too, but remains a gap. The test `test_unknown_tool_returns_response` only tests this for non-final iterations.

**Second concern**: The loop body now has two different exit paths for `iteration == MAX_ITERATIONS - 1` — the top (line 180-181 builds final prompt) and the bottom (line 212-225 extra synthesis). This makes the flow harder to follow. The `if iteration == MAX_ITERATIONS - 1` at line 212 is guaranteed to be true since we only reach this point if a tool call was extracted on the final iteration, making the condition redundant/confusing.

### tests/test_ask.py:test_final_prompt_has_no_tool_definition
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good test — verifies the key invariant that the final prompt contains no tool definition. Checks both positive (question/context present) and negative (no search_beliefs/tool JSON) properties.

### tests/test_ask.py:test_final_tool_call_triggers_extra_synthesis
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Replaces `test_max_iterations_forces_answer` which previously asserted that raw `search_beliefs` JSON appeared in the result — that was testing the bug, not correct behavior. The new test verifies: (1) 4 LLM calls are made (3 iterations + 1 extra synthesis), (2) the result contains the synthesized answer, (3) no raw tool call JSON leaks. This directly covers the issue's described bug. Well-structured.

### Missing test: final iteration timeout/error during extra synthesis
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The extra synthesis call at line 216 can raise `TimeoutExpired` or other exceptions (lines 218-222), falling back to `_beliefs_or_no_match`. There's also the case where the extra synthesis *also* returns a tool call (line 223), falling back to beliefs. The `test_timeout_after_successful_tool_search_returns_beliefs` test covers timeout on iteration 1, but there's no test specifically for timeout/error during the *extra* synthesis call at line 216. The line 223 fallback (tool call in extra synthesis) is also untested.

### Missing test: unknown tool on final iteration
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

If the LLM emits `{"tool": "unknown_tool", ...}` on the final iteration, line 210 returns the raw JSON. This is the same behavior as before the fix (the `else` branch hasn't changed), but it's still a raw tool call leak for a different tool name. The fix only guards against `search_beliefs` tool calls. This is a minor gap since unknown tools are very unlikely when the prompt doesn't even define tools, but it's worth noting.

### Removed: FINAL_TURN_INSTRUCTION
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Removing the appended instruction in favor of a separate tool-free prompt is the right approach. The old approach ("write your answer now, no more tool calls") was a soft instruction that the LLM could ignore. The new approach removes the tool definition entirely, which is a structural guarantee rather than a behavioral suggestion.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Did not examine `api.search` to verify the format parameter behavior. Could not verify git history for any related changes to `build_ask_prompt` or `FINAL_ASK_PROMPT` that might exist in earlier commits on this branch.

### Feature Requests
- Show the full function body post-patch (not just diff hunks) in the review input, so reviewers don't need to reconstruct mentally or use tools
- Include `MAX_ITERATIONS` value and other referenced constants when they appear in the diff context
- Flag when a test that previously asserted "buggy behavior" is replaced — these are important semantic changes worth highlighting automatically

## gemini [PASS]

### reasons_lib/ask.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The changes correctly address the issue where the final iteration would return a raw JSON tool call. By introducing a `FINAL_ASK_PROMPT` that lacks tool definitions, the LLM is correctly prompted to only synthesize an answer. Furthermore, the logic now handles the edge case where the LLM still outputs a tool call on the final iteration by executing the search and making one more definitive synthesis call. Finally, if the LLM still disobeys on the final extra call, it gracefully falls back to `_beliefs_or_no_match()`, preventing raw JSON from being returned to the user.

### tests/test_ask.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The tests have been properly updated to reflect the new behavior. `test_final_prompt_has_no_tool_definition` ensures the final prompt omits the tool definition, and `test_final_tool_call_triggers_extra_synthesis` (replacing the old `test_max_iterations_forces_answer`) correctly mocks the LLM to verify that a fourth synthesis call is triggered to produce a textual answer instead of returning the raw tool call.

### Self-Review
**Limitations:** None. The diff and context provided are sufficient to fully assess the correctness and integration of the fix. Ensure that `_beliefs_or_no_match()` logic correctly produces string output, though the existing codebase already uses it in `except` blocks, meaning its signature and return type are reliable.

PR #84, opened 2026-05-01
check-stale: prefix-match and upgrade truncated hashes

### LINKED ISSUE #83: check-stale: prefix-match and upgrade truncated hashes
## Problem

`check-stale` reports false STALE results during the migration from truncated (16-char) to full (64-char) SHA-256 hashes. The stored hash is a prefix of the computed hash, but the string comparison requires an exact match:

```
STALE  code:vault-secrets-mount-key-format
       source: entries/2026/03/08/README.md
       hash: 99071a024a4d77ce -> 99071a024a4d77ce99e0e44f42a6f0dfa007c76c0a349edc64aadd515ca28840
```

The old hash `99071a024a4d77ce` is the first 16 chars of the new hash. The file hasn't changed — the hash length did.

## Proposed fix

In `check_stale.py`, when comparing hashes:

1. If the stored hash is a prefix of the computed hash, treat it as a match (not stale)
2. Upgrade the stored hash to the full 64-char version in place

This handles the migration transparently — `check-stale` auto-upgrades truncated hashes on first run without false positives. After one pass, all hashes are full length and the prefix check becomes a no-op.

```python
current_hash = hash_file(path)
if current_hash != node.source_hash:
    # Prefix match: old truncated hash matches start of full hash
    if current_hash.startswith(node.source_hash):
        node.source_hash = current_hash  # upgrade in place
        continue
    # Genuine content change
    results.append(...)
```

The upgrade needs a `write=True` network save to persist. Either `check-stale` gains a `--upgrade-hashes` flag, or the upgrade happens silently during `hash-sources --force`.

benthomasson:
## Summary
- Detect when a stored hash is a prefix of the computed hash (truncated 16-char → full 64-char migration) and treat it as fresh instead of stale
- Upgrade the stored hash to full length in place so subsequent runs skip the prefix check
- Switch `api.check_stale` to `write=True` so hash upgrades persist to the database
- Report upgrade count in CLI output

Closes #83

## Test plan
- [x] `test_prefix_hash_treated_as_fresh` — 16-char prefix not reported stale, hash upgraded to 64 chars
- [x] `test_genuine_change_still_detected` — truncated hash with changed content still flagged
- [x] `test_full_hash_match_unchanged` — full hash match is a no-op
- [x] All 657 existing tests pass

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1092,18 +1092,23 @@ def export_markdown(visible_to: list[str] | None = None, db_path: str = DEFAULT_
 
 def check_stale(
     repos: dict[str, str] | None = None,
+    upgrade_hashes: bool = False,
     db_path: str = DEFAULT_DB,
 ) -> dict:
     """Check all IN nodes for source file staleness.
 
-    Returns: {"stale": list[dict], "checked": int, "stale_count": int}
+    If upgrade_hashes=True, truncated hashes that prefix-match the current
+    full hash are upgraded in place and saved to the database.
+
+    Returns: {"stale": list[dict], "checked": int, "stale_count": int,
+              "upgraded": int}
     """
     from pathlib import Path as P
     from .check_stale import check_stale as _check
 
     db_dir = P(db_path).resolve().parent
 
-    with _with_network(db_path, write=True) as net:
+    with _with_network(db_path, write=upgrade_hashes) as net:
         repo_paths = repos
         if repo_paths is None and net.repos:
             repo_paths = net.repos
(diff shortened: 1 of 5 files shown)


benthomasson (commented):
Correct and minimal. The prefix check + in-place upgrade is the right approach — after one `check-stale` run, all truncated hashes are upgraded to full length and the prefix path becomes a no-op.

Good that `api.check_stale` switches to `write=True` so the upgrades persist. The three new tests cover the important cases: prefix match upgrades, genuine change with truncated hash still detects, full match is unchanged.

One edge case to consider: a truncated hash that happens to be a prefix of a *different* file's full hash would be treated as fresh. In practice this is astronomically unlikely with 16 hex chars (1 in 2^192), so not worth guarding against.

Looks good.

PR #93, opened 2026-05-02
Add dual-path architecture, --full-sources, --natural flags, and fix FTS5 search

benthomasson:
## Summary

Dual-path architecture for TMS + FTS RAG, three new flags for `reasons ask`, cross-model support improvements, and a critical FTS5 search fix.

### Core changes

1. **`--dual`**: Run TMS and FTS RAG in **separate** LLM calls, then merge both answers in a third call. Avoids cognitive overload from mixing context types in a single prompt. This is the highest-performing configuration across all models tested.

2. **`--full-sources FTS_DB`**: Search an FTS5 index of raw source document chunks alongside beliefs, appending excerpts to context.

3. **`--natural`**: Strip belief IDs, status markers, and justification metadata from context, presenting beliefs as plain natural language paragraphs.

4. **Fix FTS5 search for natural language questions**: Strip stop words, use regex tokenization, relax AND queries down to 50% of terms. Previously, `--simple` mode returned "No matching beliefs" for most questions.

5. **Gemini submodel support** (`gemini:<model>`), `--skip-trust` for headless Gemini CLI, and strip thinking output from ollama models (gemma4).

## Key Finding: Dual-Path Is Model-Tier-Independent

Separating TMS and FTS RAG into focused calls and merging captures both strengths without cognitive overload. The architecture delivers near-perfect results across all model tiers — API and local open-weight.

### Cross-Model Dual-Path (50 general questions, seed=42)

| Model | Type | Dual-Path A/B | A grades | Previous Best |
|---|---|---|---|---|
| Opus | API | **100%** | 37A | 95.5% (TMS) |
| Haiku | API | **100%** | 26A | 76% (hybrid) |
| Sonnet | API | **98%** | 26A | 84% (RAG) |
| gemma4:31b | local 31B | **94%** | 24A | n/a |
| Gemini | API | **92%** | 35A | n/a |
| gemma3:12b | local 12B | **82%** | 23A | n/a |

### Depth-Chain Questions (25, depth 5-22, Opus)

| Configuration | A/B% | A | Completeness |
|---|---|---|---|
| Opus + Dual | 100% | 25 | 4.96 |
| Opus + TMS | 100% | 25 | 4.92 |
| Opus + FTS RAG | 88% | 14 | 2.80 |

### Cognitive Overload (Opus, general questions)

| Configuration | A/B% | A | Notes |
|---|---|---|---|
| Opus + Dual | 100% | 37 | Separate calls + merge |
| Opus + TMS alone | 95.5% | 38 | Structured beliefs only |
| Opus + Hybrid (--full-sources) | 86% | 23 | Mixed context, -9.5 pts |
| Opus + Hybrid + natural | 74% | 19 | Worse overload, -21.5 pts |

### Weaker Model Improvements

| Configuration | A/B% |
|---|---|
| Haiku + Dual | 100% |
| Haiku + natural + sources | 76% |
| Haiku + RAG | 72% |
| Haiku + TMS simple | 40% |
| Sonnet + Dual | 98% |
| Sonnet + RAG | 84% |
| Sonnet + TMS agentic | 42% |

## Test plan

- [x] Smoke test `reasons ask "..." --dual --full-sources rag_fts.db`
- [x] Smoke test `reasons ask "..." --simple --full-sources rag_fts.db`
- [x] Smoke test `reasons ask "..." --simple --natural`
- [x] Batch eval: Opus + dual general (100%, 37A)
- [x] Batch eval: Opus + dual depth (25/25 A)
- [x] Batch eval: Sonnet + dual (98%, 26A)
- [x] Batch eval: Haiku + dual (100%, 26A)
- [x] Batch eval: Gemini + dual (92%, 35A)
- [x] Batch eval: gemma4:31b + dual (94%, 24A)
- [x] Batch eval: gemma3:12b + dual (82%, 23A)
- [x] Batch eval: Opus + hybrid (86% — confirms overload)
- [x] Batch eval: Haiku + natural + sources (76%)
- [ ] Unit tests for new functions

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1337,7 +1337,7 @@ _STOP_WORDS = frozenset({
     "if", "then", "else", "while", "until", "unless",
     "there", "here", "up", "out", "off",
     "specific", "specifically", "particular", "particularly",
-    "included", "including", "within", "between",
+    "included", "including", "within",
 })
 
 
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -9,7 +9,9 @@ All functions return dicts suitable for JSON serialization.
 
 import json
 import re
+import sqlite3
 import sys
+from itertools import combinations
 from pathlib import Path
 
 from . import Justification
@@ -1343,9 +1345,6 @@ _STOP_WORDS = frozenset({
 
 def _fts_search(query: str, db_path: str) -> list[str]:
     """Search using FTS5 full-text index with porter stemming and progressive relaxation."""
-    import re
-    import sqlite3
-    from itertools import combinations
     try:
         conn = sqlite3.connect(db_path)
         try:
@@ -1421,7 +1420,6 @@ def _format_markdown(net, matched_ids: list[str], neighbor_ids: set[str]) -> str
 
 def _format_json(net, matched_ids: list[str], neighbor_ids: set[str]) -> str:
     """Format results as JSON."""
-    import json
     results = []
     for nid in matched_ids:
         node = net.nodes[nid]
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1358,10 +1358,15 @@ def _fts_search(query: str, db_path: str) -> list[str]:
 
             results = _fts_query(conn, terms)
 
+            _MAX_RELAXATION_QUERIES = 50
             if not results and len(terms) > 2:
                 min_terms = max(1, len(terms) // 2)
+                budget = _MAX_RELAXATION_QUERIES
                 for n in range(len(terms) - 1, min_terms - 1, -1):
                     for combo in combinations(terms, n):
+                        budget -= 1
+                        if budget < 0:
+                            return results
                         results = _fts_query(conn, list(combo))
                         if results:
                             return results
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1320,21 +1320,47 @@ def _fts_query(conn, terms: list[str]) -> list[str]:
     return [row[0] for row in cursor.fetchall()]
 
 
+_STOP_WORDS = frozenset({
+    "a", "an", "the", "is", "are", "was", "were", "be", "been", "being",
+    "have", "has", "had", "do", "does", "did", "will", "would", "could",
+    "should", "may", "might", "shall", "can", "need", "must",
+    "in", "on", "at", "to", "for", "of", "with", "by", "from", "as",
+    "into", "through", "during", "before", "after", "above", "below",
+    "between", "under", "over", "about", "against", "along", "among",
+    "and", "but", "or", "nor", "not", "so", "yet", "both", "either",
+    "neither", "each", "every", "all", "any", "few", "more", "most",
+    "other", "some", "such", "no", "only", "own", "same", "than",
+    "too", "very", "just", "also", "now", "how", "what", "when", "where",
+    "which", "who", "whom", "why", "this", "that", "these", "those",
+    "it", "its", "he", "she", "they", "them", "his", "her", "their",
+    "we", "you", "your", "my", "our", "me", "us", "him",
+    "if", "then", "else", "while", "until", "unless",
+    "there", "here", "up", "out", "off",
+    "specific", "specifically", "particular", "particularly",
+    "included", "including", "within", "between",
+})
+
+
 def _fts_search(query: str, db_path: str) -> list[str]:
     """Search using FTS5 full-text index with porter stemming and progressive relaxation."""
+    import re
     import sqlite3
     from itertools import combinations
     try:
         conn = sqlite3.connect(db_path)
         try:
-            terms = [t for t in query.strip().split() if t]
+            raw_terms = re.findall(r'\w+', query)
+            terms = [t for t in raw_terms if t.lower() not in _STOP_WORDS and len(t) > 1]
+            if not terms:
+                terms = [t for t in raw_terms if len(t) > 1]
             if not terms:
                 return []
 
             results = _fts_query(conn, terms)
 
             if not results and len(terms) > 2:
-                for n in range(len(terms) - 1, max(0, len(terms) - 3), -1):
+                min_terms = max(1, len(terms) // 2)
+                for n in range(len(terms) - 1, min_terms - 1, -1):
                     for combo in combinations(terms, n):
                         results = _fts_query(conn, list(combo))
                         if results:
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -208,26 +208,32 @@ def _strip_belief_metadata(beliefs_context):
 
 def _search_source_chunks(question, sources_db, top_k=10):
     """Search FTS5 index of source document chunks."""
-    words = re.findall(r'\w+', question)
-    words = [w for w in words if len(w) > 1]
+    from .api import _STOP_WORDS
+
+    raw_words = re.findall(r'\w+', question)
+    words = [w for w in raw_words if w.lower() not in _STOP_WORDS and len(w) > 1]
+    if not words:
+        words = [w for w in raw_words if len(w) > 1]
     if not words:
         return ""
     fts_query = " OR ".join(f'"{w}"' for w in words)
 
     try:
         conn = sqlite3.connect(sources_db)
-        conn.row_factory = sqlite3.Row
-        cur = conn.cursor()
-        cur.execute("""
-            SELECT c.text, c.cluster, c.filename, c.section
-            FROM chunks_fts
-            JOIN chunks c ON c.id = chunks_fts.rowid
-            WHERE chunks_fts MATCH ?
-            ORDER BY chunks_fts.rank
-            LIMIT ?
-        """, (fts_query, top_k))
-        rows = cur.fetchall()
-        conn.close()
+        try:
+            conn.row_factory = sqlite3.Row
+            cur = conn.cursor()
+            cur.execute("""
+                SELECT c.text, c.cluster, c.filename, c.section
+                FROM chunks_fts
+                JOIN chunks c ON c.id = chunks_fts.rowid
+                WHERE chunks_fts MATCH ?
+                ORDER BY chunks_fts.rank
+                LIMIT ?
+            """, (fts_query, top_k))
+            rows = cur.fetchall()
+        finally:
+            conn.close()
     except (sqlite3.OperationalError, sqlite3.DatabaseError):
         return ""
 
@@ -244,9 +250,9 @@ def _search_source_chunks(question, sources_db, top_k=10):
 
 
 FTS_RAG_PROMPT = """\
-You are a Red Hat domain expert answering questions using retrieved document excerpts.
+You are answering questions using retrieved document excerpts.
 
-Below are the most relevant excerpts from Red Hat internal documents, retrieved via
+Below are the most relevant excerpts from source documents, retrieved via
 full-text search. Use them to answer the question. Cite your sources by referencing
 the document filename in [brackets].
 
@@ -340,10 +346,13 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
 
     Returns the answer text.
     """
+    if dual and not sources_db:
+        raise ValueError("--dual requires --full-sources")
+
     if dual and sources_db:
         print("Dual path: running TMS...", file=sys.stderr)
         answer_tms = ask(question, db_path=db_path, timeout=timeout,
-                         model=model, simple=simple)
+                         model=model, simple=simple, natural=natural)
         print("Dual path: running FTS RAG...", file=sys.stderr)
         answer_fts = _fts_rag_answer(question, sources_db, model=model,
                                      timeout=timeout)
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -342,7 +342,10 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
         natural: strip belief IDs, status markers, and justification metadata
                  from context, presenting beliefs as plain natural language.
         dual: run TMS and FTS RAG in separate calls, then merge the two
-              answers in a third call. Requires sources_db.
+              answers in a third call. Requires sources_db. Makes 3 LLM
+              calls with simple=True (1 TMS + 1 FTS + 1 merge), or up to
+              5 with simple=False (up to 3 TMS tool-loop rounds + 1 FTS
+              + 1 merge). Timeout applies per call.
 
     Returns the answer text.
     """
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -15,6 +15,9 @@ from . import api
 from .llm import invoke_model
 
 
+_CITE_RULE = "- Cite belief IDs in [brackets] when referencing specific beliefs."
+_CITE_RULE_NATURAL = "- Do not cite belief IDs. Answer in plain natural language."
+
 ASK_PROMPT = """\
 You are answering a question using a belief network (a Truth Maintenance System).
 Each belief has an ID, text, truth value (IN = held true, OUT = retracted), and
@@ -29,7 +32,7 @@ Rules:
   answer directly. Do NOT call the tool.
 - If you need to search for more beliefs, respond with ONLY a single JSON line
   (no other text). The system will run the search and give you the results.
-- Cite belief IDs in [brackets] when referencing specific beliefs.
+{cite_rule}
 - ONLY answer based on the beliefs provided. Do NOT use your training data or
   general knowledge to fill gaps.
 - If the beliefs are insufficient to answer, respond EXACTLY with:
@@ -52,7 +55,7 @@ Each belief has an ID, text, truth value (IN = held true, OUT = retracted), and
 may have justifications tracing why it is believed.
 
 Rules:
-- Cite belief IDs in [brackets] when referencing specific beliefs.
+{cite_rule}
 - ONLY answer based on the beliefs provided. Do NOT use your training data or
   general knowledge to fill gaps.
 - If the beliefs are insufficient to answer, respond EXACTLY with:
@@ -76,7 +79,7 @@ Each belief has an ID, text, truth value (IN = held true, OUT = retracted), and
 may have justifications tracing why it is believed.
 
 Rules:
-- Cite belief IDs in [brackets] when referencing specific beliefs.
+{cite_rule}
 - ONLY answer based on the beliefs provided. Do NOT use your training data or
   general knowledge to fill gaps.
 - If the beliefs are insufficient to answer, respond EXACTLY with:
@@ -92,11 +95,12 @@ Rules:
 {beliefs_context}"""
 
 
-def build_simple_prompt(question, beliefs_context):
+def build_simple_prompt(question, beliefs_context, natural=False):
     """Build prompt for simple single-pass synthesis — no tool definitions."""
     return SIMPLE_ASK_PROMPT.format(
         question=question,
         beliefs_context=beliefs_context,
+        cite_rule=_CITE_RULE_NATURAL if natural else _CITE_RULE,
     )
 
 
@@ -119,7 +123,7 @@ def extract_tool_call(text):
     return None
 
 
-def build_ask_prompt(question, beliefs_context, tool_history=None):
+def build_ask_prompt(question, beliefs_context, tool_history=None, natural=False):
     """Build the full prompt for LLM synthesis."""
     history_section = ""
     if tool_history:
@@ -135,10 +139,11 @@ def build_ask_prompt(question, beliefs_context, tool_history=None):
         question=question,
         beliefs_context=beliefs_context,
         tool_history=history_section,
+        cite_rule=_CITE_RULE_NATURAL if natural else _CITE_RULE,
     )
 
 
-def build_final_prompt(question, beliefs_context, tool_history=None):
+def build_final_prompt(question, beliefs_context, tool_history=None, natural=False):
     """Build prompt for final synthesis — no tool definition."""
     history_section = ""
     if tool_history:
@@ -154,6 +159,7 @@ def build_final_prompt(question, beliefs_context, tool_history=None):
         question=question,
         beliefs_context=beliefs_context,
         tool_history=history_section,
+        cite_rule=_CITE_RULE_NATURAL if natural else _CITE_RULE,
     )
 
 
@@ -185,6 +191,14 @@ def _strip_belief_metadata(beliefs_context):
         if line.startswith("**Supported by:**"):
             continue
         if line.startswith("**Supports:**"):
+            continue
+        if line.startswith("**Depended on by:**"):
+            continue
+        if line.startswith("**Related nodes:**"):
+            continue
+        if re.match(r'^- \*\*\S+\*\* \((?:IN|OUT)\):', line):
+            continue
+        if line.strip() == "---":
             continue
         out.append(line)
     result = "\n".join(out).strip()
@@ -358,7 +372,7 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
         if not beliefs_context.strip():
             return NO_BELIEFS_MSG
 
-        prompt = build_simple_prompt(question, beliefs_context)
+        prompt = build_simple_prompt(question, beliefs_context, natural=natural)
         print("Synthesizing (simple)...", file=sys.stderr)
         try:
             response = invoke_model(prompt, model=model, timeout=timeout)
@@ -384,9 +398,9 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
 
     for iteration in range(MAX_ITERATIONS):
         if iteration == MAX_ITERATIONS - 1:
-            prompt = build_final_prompt(question, beliefs_context, tool_history)
+            prompt = build_final_prompt(question, beliefs_context, tool_history, natural=natural)
         else:
-            prompt = build_ask_prompt(question, beliefs_context, tool_history)
+            prompt = build_ask_prompt(question, beliefs_context, tool_history, natural=natural)
 
         print(f"Synthesizing (round {iteration + 1}/{MAX_ITERATIONS})...",
               file=sys.stderr)
@@ -417,7 +431,7 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
 
         if iteration == MAX_ITERATIONS - 1:
             print(f"Synthesizing (final)...", file=sys.stderr)
-            prompt = build_final_prompt(question, beliefs_context, tool_history)
+            prompt = build_final_prompt(question, beliefs_context, tool_history, natural=natural)
             try:
                 response = invoke_model(prompt, model=model, timeout=timeout)
             except subprocess.TimeoutExpired:
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -229,6 +229,77 @@ def _search_source_chunks(question, sources_db, top_k=10):
     return "\n\n---\n\n".join(parts)
 
 
+FTS_RAG_PROMPT = """\
+You are a Red Hat domain expert answering questions using retrieved document excerpts.
+
+Below are the most relevant excerpts from Red Hat internal documents, retrieved via
+full-text search. Use them to answer the question. Cite your sources by referencing
+the document filename in [brackets].
+
+If the excerpts don't contain enough information to answer the question, say so honestly.
+Do not fabricate information that isn't in the provided excerpts.
+
+## Retrieved Documents
+
+{context}
+
+## Question
+
+{question}
+
+## Instructions
+
+- Answer the question based on the retrieved documents above
+- Cite sources using [filename] notation
+- If information is insufficient, say what you can and note the gaps
+- Be specific and concise
+"""
+
+MERGE_PROMPT = """\
+You are merging two answers to the same question. Each answer was produced
+independently using a different retrieval method:
+
+- Answer A used a structured belief network with dependency chains
+- Answer B used full-text search over source documents
+
+Produce a single merged answer that:
+- Combines information from both answers
+- When both answers cover the same point, use the more specific/detailed version
+- Preserve all citations (belief IDs in [brackets] from Answer A, [filenames] from Answer B)
+- Do not add information that neither answer contains
+- If the answers contradict each other, note the disagreement
+
+## Question
+
+{question}
+
+## Answer A (Belief Network)
+
+{answer_tms}
+
+## Answer B (Source Documents)
+
+{answer_fts}
+"""
+
+
+def _fts_rag_answer(question, sources_db, model="claude", timeout=300):
+    """Run FTS5 RAG over source document chunks and synthesize an answer."""
+    context = _search_source_chunks(question, sources_db)
+    if not context:
+        return "No relevant documents found for this question."
+    prompt = FTS_RAG_PROMPT.format(context=context, question=question)
+    return invoke_model(prompt, model=model, timeout=timeout).strip()
+
+
+def _merge_answers(question, answer_tms, answer_fts, model="claude", timeout=300):
+    """Merge two answers from different retrieval paths."""
+    prompt = MERGE_PROMPT.format(
+        question=question, answer_tms=answer_tms, answer_fts=answer_fts,
+    )
+    return invoke_model(prompt, model=model, timeout=timeout).strip()
+
+
 MAX_ITERATIONS = 3
 
 NO_BELIEFS_MSG = "No matching beliefs found. Cannot answer from the belief network."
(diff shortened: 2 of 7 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#93
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## Disagreements

- **reasons_lib/api.py:_fts_search** [MEDIUM]: claude: CONCERN, gemini: BLOCK
- **reasons_lib/ask.py:_strip_belief_metadata** [MEDIUM]: claude: CONCERN, gemini: BLOCK

## claude [CONCERN]

### reasons_lib/api.py:_STOP_WORDS
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The stop word list includes "between" twice — once in the prepositions block (line 1329) and again in the final catch-all block (line 1340). This is harmless (frozenset deduplicates), but signals the list was assembled without a final pass. More substantively, some domain-relevant words are included as stop words that could cause false negatives in a belief network context: "not", "no", "need", "must", "should", "before", "after" all carry semantic weight in reasoning. A query like "what must happen before X" would be stripped to just "happen X". No tests validate the stop word filtering behavior.

### reasons_lib/api.py:_fts_search
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The progressive relaxation change (`min_terms = max(1, len(terms) // 2)`) is more aggressive than before — for 6 terms, the old code tried 5 and 4 (removed 1–2 terms), while new code tries 5, 4, 3 (down to half). This is reasonable for natural language queries. However, `_fts_query` joins terms with spaces (implicit AND in FTS5), so the combinatorial explosion is a concern: `combinations(6, 3)` = 20 queries; if there are 8 content terms, `combinations(8, 4)` = 70 queries. This could be slow on large databases. The `import re` is now inside the function body rather than at module level — minor but inconsistent with the module-level `_STOP_WORDS` that uses no import. No tests cover the new relaxation bounds or stop-word filtering.

### reasons_lib/ask.py:_strip_belief_metadata
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The function strips lines starting with specific prefixes but doesn't strip the belief text line itself that often follows `### node-id`. If the export format is `### node-id\nBelief text here\n**Status:** IN`, then removing the header and status lines preserves the text — this works. But if beliefs have text embedded after `### ` on the same line, that text is lost. The regex `r'\n{3,}'` cleanup is good. The function doesn't handle the case where `beliefs_context` could be whitespace-only (though the caller guards against empty). No unit tests exist for this function — edge cases like beliefs with only metadata (no prose) or unusual formatting are unverified.

### reasons_lib/ask.py:_search_source_chunks
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Two issues: (1) No stop-word filtering — unlike `_fts_search` in api.py which filters `_STOP_WORDS`, this function uses all words > 1 char. A natural language question like "What are the specific reasons for this decision?" produces an OR query with many noise terms, potentially returning low-relevance chunks. The inconsistency between the two search paths is concerning. (2) The `conn.close()` is not in a `finally` block — if `cur.execute` raises an exception not caught by the `except` clause (e.g., `KeyboardInterrupt`), the connection leaks. This is minor but inconsistent with the `try/finally` pattern used in `_fts_search`. (3) The `row["section"]` truthiness check on line 226 will fail if section is `None` vs empty string differently, but both are handled correctly by the truthiness check.

### reasons_lib/ask.py:ask (modified)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The `natural` and `sources_db` features are applied in both the `simple` and full (tool-loop) code paths, which is good. However, in the full path (line 297), when `beliefs_context` is falsy (None or empty) and `sources_db` returns results, the context becomes `"\n\n## Source Documents\n\n..."` with a leading double-newline, which is cosmetic but sloppy. In the simple path (line 272), if `beliefs_context` is empty string and sources are found, same issue: `"" + "\n\n## Source Documents\n\n..."`. More importantly, the `no_synth` path (line 256-258) doesn't apply `natural` or `sources_db` at all — a user passing `--no-synth --natural` gets raw metadata. This may be intentional (no-synth = raw output) but isn't documented. No tests exist for any of the new parameter combinations.

### reasons_lib/cli.py:cmd_ask + argparse
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The CLI wiring is correct. `args.full_sources` maps to `sources_db`, `args.natural` maps to `natural`. The argparse definitions use appropriate types (`default=None` for `--full-sources`, `store_true` for `--natural`). The metavar `FTS_DB` gives good UX. The hyphenated `--full-sources` is correctly accessed as `args.full_sources` (argparse auto-converts). Integration is complete — flags reach the `ask()` function.

### Self-Review
**Limitations:** Could not see the full export format produced by `api.search(format="markdown")` to verify that `_strip_belief_metadata` correctly handles all metadata lines. Could not verify the `chunks` and `chunks_fts` table schemas to confirm the SQL in `_search_source_chunks` is valid. No test files were included in the diff, so the "UNTESTED" verdicts are based on absence in the diff — tests may exist elsewhere but were not modified or added. Could not verify whether `_fts_query`'s implicit AND semantics are intentional for the new OR-based `_search_source_chunks` approach.

### Feature Requests
- Include test files that exist for modified source files (e.g., show test_ask.py alongside ask.py changes) to verify coverage claims
- Show the schema or sample output of upstream functions (like `api.search(format="markdown")`) when downstream code parses that output
- Flag when a new function duplicates logic from an existing function (e.g., `_search_source_chunks` reimplements word tokenization without the stop-word filtering from `_fts_search`)

## gemini [BLOCK]

### reasons_lib/api.py:_STOP_WORDS
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The set contains a duplicate entry for `"between"` (lines 1329 and 1340). While sets deduplicate automatically, this indicates a minor lack of polish.

### reasons_lib/api.py:_fts_search
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **Performance Explosion:** The progressive relaxation using `combinations(terms, n)` down to `n = len(terms) // 2` is a severe performance risk. For a 20-term query, searching 10-term combinations involves C(20, 10) = 184,756 SQLite queries, which will hang the process. Capping at `len - 2` or `len - 3` is necessary.
2. **Logic Flaw:** The `len(terms) > 2` guard prevents relaxation for 2-term queries. A search for "foo bar" that fails to find both won't fall back to searching for "foo" or "bar" individually.
3. **Redundant Import:** `import re` is added locally despite already being imported at the module level (line 11).

### reasons_lib/ask.py:_strip_belief_metadata
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **Incomplete Stripping:** The function misses several metadata headers produced by `api.py:_format_markdown`, notably `**Depended on by:**` (api.py:1403), the `---` separator, and the `**Related nodes:**` neighbor list. These structured lines will leak into the "natural" output.
2. **Instruction Conflict:** Stripping belief IDs (`### ID`) creates a direct conflict with the `ASK_PROMPT` which instructs the LLM to "Cite belief IDs in [brackets]". Without IDs in the context, the LLM will likely hallucinate them or fail to follow the citation rule.

### reasons_lib/ask.py:_search_source_chunks
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **Noisy Search:** This function uses `OR` logic on all terms with length > 1 but does NOT filter stop words. A query like "What is the system?" will match nearly every chunk containing "is" or "the", resulting in very low-signal RAG results. It should use the same stop-word list as `api.py`.
2. **Fragile Join:** The `JOIN` on `c.id = chunks_fts.rowid` assumes `id` is an integer rowid; if the chunk ID is a string, this query will return no results.

### reasons_lib/ask.py:ask
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **Inconsistent State:** The `natural` stripping and `sources_db` augmentation are not applied to tool-call results in the multi-pass loop. This leads to mixed context where some beliefs are "natural" and others have full metadata.
2. **Context Loss:** In the tool loop, `beliefs_context` is overwritten by the new search results (`beliefs_context = result`), causing the initial "Source Documents" retrieved via RAG to be lost in subsequent rounds.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The CLI wiring for `--full-sources` and `--natural` is correct, well-documented in the help text, and correctly passes arguments to the library.

### Self-Review
**Limitations:** I do not have the schema for the external `sources_db` to verify the `rowid` join assumption. I was able to verify the metadata leak by comparing `_strip_belief_metadata` with the existing `_format_markdown` implementation in `api.py`.

### Feature Requests
- Include the schema of any external databases referenced in the code (e.g., the RAG/source database).
- Provide existing tests alongside implementation changes to help detect potential performance regressions or behavioral side-effects.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#93
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## Disagreements

- **reasons_lib/ask.py:_strip_belief_metadata** [MEDIUM]: claude: CONCERN, gemini: BLOCK

## claude [CONCERN]

### reasons_lib/llm.py:resolve_model_cmd
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Logic is correct — gemini submodel support via `gemini:<model>` and `--skip-trust` flag addition are sound. However, the existing test `test_resolve_gemini` (test_llm.py:16) asserts `["gemini", "-p", ""]` but the code now returns `["gemini", "--skip-trust", "-p", ""]`. **This test will fail.** The new `gemini:<model>` path has no test at all.

### reasons_lib/llm.py:invoke_model (ollama thinking strip)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The thinking-strip logic (lines 68-71) checks for `"Thinking...\n"` but splits on `"...done thinking.\n"`. If the output contains `"Thinking...\n"` but not `"...done thinking.\n"` (e.g. the model's thinking was interrupted or the format changed), the code silently returns the full output including thinking markers — acceptable degradation. But the `model.startswith("ollama:")` check means this only fires for ollama models even though the marker detection is the real signal. More importantly, there are zero tests for this behavior.

### reasons_lib/api.py:_STOP_WORDS
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Reasonable stop word list for FTS query cleanup. Minor nit: `"between"` appears twice (line 1329 and line 1340) — harmless in a frozenset but worth cleaning up. The words `"specific"`, `"specifically"`, `"particular"`, `"particularly"` are domain-specific additions that make sense for natural language queries against a belief network.

### reasons_lib/api.py:_fts_search
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `re.findall(r'\w+', query)` change (line 1352) is better than `split()` for handling punctuation in natural language queries. However, the progressive relaxation change from `max(0, len(terms) - 3)` to `max(1, len(terms) // 2)` is a behavioral change that could affect existing callers. With the old code, a 5-term query tried combos of size 4 and 3. With the new code (`min_terms = max(1, 5//2) = 2`), it tries combos of size 4, 3, and 2 — more aggressive relaxation, potentially returning false positives. For very long queries (10 terms), old code tried 9 and 8; new code tries 9, 8, 7, 6, 5 — significantly more combinations, with potential combinatorial explosion (`C(10,5) = 252` calls to `_fts_query`). Existing tests may still pass but behavioral edge cases are untested.

### reasons_lib/ask.py:_strip_belief_metadata
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The function uses `line.startswith()` checks, which means it only strips metadata lines at column 0. If the markdown format uses indentation (e.g. nested lists), metadata lines would survive. Also, it strips `### ` headers (which contain the belief IDs like `### node-id [IN]`) but this also strips any `### ` heading the text itself might contain. No tests exist to verify the stripping logic or edge cases.

### reasons_lib/ask.py:_search_source_chunks
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Unlike `_fts_search` in api.py, this function does **not** apply stop word filtering (line 197-198 only filters `len(w) > 1`). A query like "What is the purpose of the system" would generate an FTS query with all those stop words OR'd together, returning essentially random high-frequency chunks. This is inconsistent with the stop-word filtering added to `_fts_search` in the same diff. The `conn.close()` is in the try block but not in a finally — if `fetchall()` raises, the connection leaks (though the GC would eventually close it). No tests.

### reasons_lib/ask.py:FTS_RAG_PROMPT
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The prompt hardcodes "You are a Red Hat domain expert" (line 233). This is a domain-specific assumption baked into a general-purpose library. If this tool is used outside a Red Hat context, the persona is misleading. Should be parameterized or generic.

### reasons_lib/ask.py:ask (dual mode)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The dual-mode recursive call at line 331 passes `simple=simple` but not `natural=natural`. So if a user passes `--dual --natural`, the TMS leg runs without metadata stripping. This seems like an oversight — `natural` should propagate. Also, `--dual` without `--full-sources` silently does nothing (line 329: `if dual and sources_db`), which is confusing — the CLI help says "requires --full-sources" but there's no validation error. The three sequential LLM calls (TMS answer, FTS answer, merge) mean the timeout applies per-call, so total wall time could be up to 3x timeout.

### reasons_lib/ask.py:ask (simple + sources_db path)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The logic correctly handles the case where beliefs are empty but source documents exist — it concatenates them and only returns NO_BELIEFS_MSG if both are empty (line 358). The ordering of `natural` stripping before source document appending is correct (don't strip metadata from source docs).

### reasons_lib/cli.py (new args)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

CLI args are correctly wired to `ask()` parameters. The `--full-sources` uses `default=None` and `metavar="FTS_DB"` which is clear. The `--dual` help text correctly notes the dependency on `--full-sources`, though no runtime validation enforces it.

### Self-Review
**Limitations:** Could not run the test suite to confirm the `test_resolve_gemini` breakage. Did not see the full `test_ask.py` file to verify no other tests break due to signature changes. Could not verify the FTS5 schema of `chunks_fts` / `chunks` tables to confirm the JOIN in `_search_source_chunks` is correct. Did not see how `_search_source_chunks` callers handle None vs empty string return.

### Feature Requests
- Include related test files in full when implementation files are modified, so broken assertions can be identified with certainty
- Show the full function body (not just diff hunks) for all modified functions to verify complete logic
- Flag when a diff modifies a function signature and show all callers of that function
- Detect when constants used in tests (like expected return values) are changed in source but not in tests

## gemini [BLOCK]

### reasons_lib/api.py:_fts_search
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The addition of stop words and the `len(t) > 1` filter improves search quality for long natural language queries by reducing noise. However, it introduces a regression: one-character terms (like "X", "3", or "v") are now entirely filtered out of searches, even if they are not in the stop word list. If a belief is about "Version 3" or "File X", it may become harder to find. Additionally, "between" is redundantly included twice in `_STOP_WORDS`.

### reasons_lib/ask.py:_strip_belief_metadata
**Verdict:** BLOCK
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

This function strips the `### ID` lines from the belief context. However, the LLM prompts (`ASK_PROMPT`, `SIMPLE_ASK_PROMPT`, etc.) explicitly mandate that the model "Cite belief IDs in [brackets] when referencing specific beliefs." When `--natural` is used, the model is instructed to cite IDs that have been removed from its context, which will lead to hallucinations (guessing IDs) or failure to comply with system rules. The prompt should be made dynamic or the stripping logic should preserve IDs in a less prominent way if citations are still required.

### reasons_lib/ask.py:ask
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

In the standard (non-simple) path, if `api.search` returns "No results found.", this string is preserved in `beliefs_context`. When `sources_db` is present, it appends "Source Documents" to this "No results found." message, resulting in a slightly confusing prompt for the LLM. The `simple` path correctly clears this string to an empty string; the standard path should do the same for consistency. Also, the `dual` mode recursive call to `ask` does not propagate the `natural` flag, which might be an oversight if the user expects a natural-language sub-answer before merging.

### reasons_lib/llm.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The updates to Gemini submodel handling (`gemini:<model>`) and the addition of `--skip-trust` are correct and follow the CLI usage. The Ollama "Thinking..." output handling is a helpful quality-of-life fix for models that include reasoning traces in stdout, though it may need future updates if Ollama changes its delimiter strings or uses `<think>` tags.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

New CLI arguments (`--full-sources`, `--natural`, `--dual`) are correctly defined and passed to the `ask` implementation.

### Self-Review
**Limitations:** I do not have access to the `rag_fts.db` (sources_db) schema, but the code's SQL queries imply a standard `chunks` table structure which is common for such RAG implementations. I am also assuming the `gemini` CLI behavior based on `cli_help` documentation.

### Feature Requests
- Include test file alongside implementation changes to verify that new logic (like the progressive relaxation in `_fts_search`) is actually exercised by the suite.
- Provide a way to run a subset of tests easily during the review process.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#93
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/api.py:_STOP_WORDS
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Standard stop word set as a module-level frozenset. Includes common English stop words plus a few domain-specific entries (`specific`, `specifically`, `particular`, `particularly`, `included`, `including`, `within`). The domain-specific additions are reasonable for a natural language query interface. Frozenset is the right choice for O(1) lookup.

### reasons_lib/api.py:_fts_search
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Two issues: (1) The `import re` is inside the function body — should be at module level for consistency and to avoid repeated import overhead (though Python caches it, it's a style concern). (2) The progressive relaxation logic has a subtle issue: `min_terms = max(1, len(terms) // 2)` means for 3 terms, min_terms=1, so it tries all combos of size 2 and size 1. For `len(terms) == 2`, the outer `if not results and len(terms) > 2` skips relaxation entirely, which is correct but differs from the old behavior that only went down to `max(0, len(terms) - 3)`. The budget cap of 50 is well-tested. The fallback from stop-word-filtered terms to raw terms (when all are stop words) is good defensive behavior. One edge: `re.findall(r'\w+', query)` treats underscores as word characters, which is actually appropriate for belief IDs like `prop-bfs` being split on hyphens — wait, `\w+` won't split on hyphens, so `prop-bfs` would be one token `prop` and one `bfs`. Actually no, `-` is not `\w`, so it splits correctly.

### reasons_lib/ask.py:_CITE_RULE / _CITE_RULE_NATURAL
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean extraction of citation rules into constants with a `{cite_rule}` format parameter. All three prompt templates (ASK_PROMPT, FINAL_ASK_PROMPT, SIMPLE_ASK_PROMPT) are updated consistently. The natural mode instruction "Do not cite belief IDs. Answer in plain natural language." is clear and appropriate.

### reasons_lib/ask.py:_strip_belief_metadata
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function strips metadata lines using `startswith` checks. This works for the known format but is fragile — if a belief's text content starts with `**Status:**` (unlikely but possible), it would be incorrectly stripped. The regex `r'^- \*\*\S+\*\* \((?:IN|OUT)\):'` for related node lines is specific enough. More concerning: the function returns `None` when given `None` input (line `assert _strip_belief_metadata(None) is None` in test) but has no explicit `None` check — it relies on the `if not beliefs_context: return beliefs_context` guard, which returns `None` for `None` input. This works but the return type is implicitly `str | None`, which is fine since callers check `if natural and beliefs_context:` before calling. The `---` separator stripping could be overly aggressive if belief text contains `---` on its own line, but that's a markdown convention unlikely to appear in belief text.

### reasons_lib/ask.py:_search_source_chunks
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Uses `OR`-based FTS5 matching which will return results if any term matches. This is a reasonable approach for RAG retrieval but differs from the belief search's AND-then-relax strategy. The function imports `_STOP_WORDS` from `api` — a cross-module dependency that could be extracted to a shared location but is acceptable for now. The `conn.close()` in a `try/finally` is correct. The `conn.row_factory = sqlite3.Row` is set after connect but before any queries, which is fine. One concern: the function creates a new connection each call with no connection pooling, which could be slow for dual-mode where it's called multiple times. Not a bug, but a performance consideration.

### reasons_lib/ask.py:FTS_RAG_PROMPT / MERGE_PROMPT
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Prompts are well-structured. The merge prompt correctly instructs the LLM to note disagreements between answers rather than silently picking one. Both prompts instruct against fabrication. The FTS_RAG_PROMPT uses `[filename]` citation style which differs from the TMS `[belief-id]` style — the merge prompt explicitly says to "Preserve all citations" from both, which is appropriate. No direct unit tests on prompt content (only integration via `ask()`), but the prompts are templates used by tested functions.

### reasons_lib/ask.py:ask (modified)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `dual` path recursively calls `ask()` for the TMS leg, which works but means the TMS leg will also receive `sources_db=None` (not passed), so it won't include source documents — this is intentional since dual mode separates the two retrieval paths. However, the recursive call passes `simple=simple` which means if `simple=False`, the TMS leg gets the full iterative tool-call loop, potentially making 3 LLM calls for the TMS leg alone, plus 1 for FTS RAG, plus 1 for merge = 5 calls total. This is undocumented behavior.

In the simple path: when beliefs are empty but sources exist, `beliefs_context` starts as `""`, then gets sources appended, resulting in `"\n\n## Source Documents\n\n..."`. The `if not beliefs_context.strip()` check correctly handles the case where both are empty. But when beliefs are empty and sources exist, the prompt will contain only source documents under the belief context section, which may confuse the LLM since the prompt says "belief matches."

The `_full_context()` closure captures `beliefs_context` and `sources_suffix` by reference. When `beliefs_context` is reassigned inside the tool-call loop (line where `beliefs_context = result`), `_full_context()` will use the new value. This is correct behavior but relies on closure semantics that could be confusing.

### reasons_lib/cli.py:cmd_ask + argparse
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

CLI arguments `--full-sources`, `--natural`, and `--dual` are correctly wired to the `ask()` function. The `--full-sources` uses `metavar="FTS_DB"` for clarity. No CLI-level tests for these new flags (all tests use the `ask()` function directly), but the wiring is straightforward pass-through. The `--dual` help text correctly notes the `--full-sources` requirement, but there's no argparse-level validation — the ValueError in `ask()` handles it at runtime, which is fine.

### reasons_lib/llm.py:resolve_model_cmd (gemini changes)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adds `--skip-trust` to gemini commands and introduces `gemini:<model>` submodel syntax. The gemini submodel uses `-m` flag (different from claude's `--model`), which matches the gemini CLI conventions. The error message's available list correctly includes the new `gemini:<model>` option.

### reasons_lib/llm.py:invoke_model (ollama thinking strip)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The thinking output stripping uses string matching on `"Thinking...\n"` and `"...done thinking.\n"`. This is fragile — if Ollama changes the format (e.g., different casing, extra spaces, localization), the stripping will break silently. The check `if model.startswith("ollama:")` is correct. The fallback behavior (keeping original output if no `...done thinking.\n` marker is found) is safe. However, if the actual LLM response legitimately contains the string `"Thinking...\n"` (e.g., discussing the thinking feature), the first check would match incorrectly. The `split("...done thinking.\n", 1)` with `maxsplit=1` correctly handles the case where the response itself contains this string after the thinking block.

### tests/test_api.py:TestFtsSearch (new tests)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good coverage of the new FTS search features: stop word filtering, fallback behavior, single-char word filtering, natural language queries, punctuation handling, and the query explosion cap. The `test_long_query_does_not_explode` test uses mock counting to verify the budget cap, asserting `<= 51` (initial query + 50 relaxation budget). This is correct. Tests use the `_fts_search` internal function directly where needed, which is appropriate for unit testing.

### tests/test_ask.py:TestStripBeliefMetadata
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: COVERED
INTEGRATION: N/A
REASONING: Thorough test coverage of the metadata stripping function. Tests each metadata line type individually, plus a comprehensive `test_full_markdown_format` that verifies all stripping in combination. Edge cases (empty input, `None`, blank line collapsing) are covered. The `test_strips_separator` test verifies `---` removal. Good practice testing both individual and combined behavior.
---

### tests/test_ask.py:TestSearchSourceChunks
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: COVERED
INTEGRATION: N/A
REASONING: Tests create a realistic FTS5-indexed SQLite database with multiple chunks. Coverage includes: matching, section headers, no-section case, no-match, single-char filtering, bad DB path, top_k limit, stop word filtering, and all-stop-words fallback. The `test_all_stop_words_falls_back` assertion `assert result == "" or isinstance(result, str)` is weak — it always passes since both branches return a string. This should test whether a result is returned when all terms are stop words, but the test data doesn't have content matching "what", "is", "the" anyway, so it would return empty regardless.
---

### tests/test_ask.py:TestAskSourcesPreserved
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Tests verify that source documents survive across tool-call rounds (critical for the iterative synthesis loop). The `test_sources_survive_tool_call` test uses a counting mock to verify that on the second LLM call, the prompt still contains source documents. The `test_natural_applied_to_tool_results` test verifies metadata is stripped from tool-call results in natural mode. Both test the behavior that was the motivation for the "preserve source documents across tool-call rounds" commit.

### tests/test_ask.py:TestAskDual
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Tests cover the three-call pattern (TMS, FTS RAG, merge), the ValueError when `sources_db` is missing, and natural mode propagation to the TMS leg. The `test_dual_calls_three_times` test verifies exactly 3 LLM invocations. The `test_dual_propagates_natural` test checks that `**Status:**` is absent from the TMS leg prompt, confirming natural mode propagation through the recursive `ask()` call.

### Self-Review
**Limitations:** Could not see the full `api.py` `search()` function to verify the format of markdown output that `_strip_belief_metadata` processes. Could not verify the Gemini CLI's actual flag names (`--skip-trust`, `-m`, `-p`). Did not have access to the `rag_fts.db` schema to verify the chunks table matches what `_search_source_chunks` expects.

### Feature Requests
- Include full function bodies for modified functions (the observation section mentioned this but none were provided)
- Show the schema of referenced database tables when SQL queries are in the diff
- Include the existing test file contents when new test classes are added to verify no conflicts with existing fixtures

## gemini [PASS]

### reasons_lib/api.py: _fts_search
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The transition to `re.findall(r'\w+', query)` for tokenization is a significant improvement over simple splitting, as it handles punctuation gracefully. The introduction of a stop-word filter and a relaxation budget (`_MAX_RELAXATION_QUERIES = 50`) prevents combinatorial explosion on long natural language queries, which was a potential performance bottleneck.

### reasons_lib/ask.py: _strip_belief_metadata
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

This function correctly identifies and removes TMS-specific metadata (headers, status, justifications, etc.) to present a clean natural language context to the LLM. This is essential for the new "natural" mode.

### reasons_lib/ask.py: _search_source_chunks
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Implements a clean RAG retrieval path over an external FTS5 database. Using `OR` for term matching in the source-document context (as opposed to the stricter `AND` logic used for beliefs) is a sound architectural choice for document retrieval.

### reasons_lib/ask.py: ask
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `ask` function has been elegantly extended to support three distinct modes: standard TMS synthesis, RAG-augmented synthesis, and a "dual" path that merges results from both. The recursion used in the `dual` path is correctly implemented with default arguments to avoid infinite loops.

### reasons_lib/llm.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Support for Gemini submodels and the addition of `--skip-trust` aligns with the Gemini CLI environment. The Ollama "thinking" stripper is a practical addition for handling reasoning models (like DeepSeek) that include chain-of-thought in their output.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

CLI arguments (`--full-sources`, `--natural`, `--dual`) are correctly wired to the `ask` function.

### Self-Review
**Limitations:** I assumed the `gemini` CLI tool referenced in `llm.py` follows the established pattern of the Gemini CLI wrapper used in this environment. The stop-word list is hardcoded in `api.py`, which is fine for a library of this scope but may need to be configurable in the future.

PR #97, opened 2026-05-05
Add review-beliefs command (#96)

### LINKED ISSUE #96: review-beliefs: validate derived beliefs against their justifications
## Context

The belief network currently holds 703 IN beliefs (437 derived) at max depth 15. Derived beliefs are proposed by an LLM and accepted after structural validation (antecedents exist and are IN, cycle/depth guards, Jaccard dedup). However, there is no mechanism to verify that the *reasoning* from antecedents to conclusion is semantically sound.

Beliefs in the network are **tentatively true according to the decisions made by the LLM, given the information it had at the time.** This is the correct operating mode for a TMS — beliefs are held because their justifications support them, and they are automatically retracted when support changes. But as the network grows and derivation depth increases, reasoning errors can compound silently.

`check-stale` already validates premises by re-reading source code. This proposal extends that pattern to derived beliefs.

## Proposal

Add a `review-beliefs` (or `review-conclusions`) command that examines derived beliefs and their justification chains. For each belief, an LLM evaluates three axes:

### 1. Valid?

Does this conclusion actually follow from these antecedents? A belief stating "X implies Y" should be logically supported by the beliefs it cites. Flag beliefs where the conclusion is a plausible-sounding leap rather than a sound derivation.

### 2. Sufficient?

Are the antecedents enough to support the conclusion, or is the derivation making an unstated assumption? A belief justified by `[a, b]` that actually requires `[a, b, c]` to hold has an incomplete justification — it should either gain `c` as an antecedent or be flagged for review.

### 3. Necessary?

Are all antecedents actually load-bearing? If a belief is justified by `[a, b, c]` but the conclusion follows from `[a, b]` alone, the justification has unnecessary complexity. Simplifying it makes the dependency graph more accurate and retraction more precise.

## Possible actions from review

| Finding | Action |
|---------|--------|
| Invalid reasoning | Flag for retraction |
| Insufficient justification | Flag for strengthening (add antecedents) or demotion to premise |
| Unnecessary antecedents | Simplify justification |
| All three pass | Mark as reviewed (metadata) |

## Interface sketch

```bash
# Review all derived beliefs
reasons review-beliefs

# Review specific beliefs
reasons review-beliefs belief-id-1 belief-id-2

# Review beliefs at depth > N (where compounding risk is highest)
reasons review-beliefs --min-depth 10

# Review beliefs derived from a specific premise (cascade check)
reasons review-beliefs --depends-on premise-id

# Dry run — report findings without taking action
reasons review-beliefs --dry-run

# Auto-retract invalid beliefs
reasons review-beliefs --auto-retract
```

## Design considerations

- **Cost control**: Reviewing 437 derived beliefs is expensive. Support `--budget N`, `--sample N`, and depth/dependency filters to target review effort where it matters most.
- **Review metadata**: Record review timestamp and result on each belief so reviewed beliefs can be skipped in future runs (like `check-stale` records source info).
- **Model choice**: The reviewing model should ideally differ from the deriving model to avoid shared blind spots. Support `--model` flag.
- **Batch prompting**: Group beliefs with shared antecedents to reduce redundant context in prompts.
- **Complements check-stale**: `check-stale` validates premises against code; `review-beliefs` validates derived beliefs against their justifications. Together they cover the full network.
- **Complements #94**: Issue #94 verifies *answers* against the belief network. This verifies the *beliefs themselves*. Both increase confidence but at different layers.

## Expected impact

- Higher confidence in the belief network as a knowledge source
- Early detection of reasoning errors before they propagate to deeper derivations
- Cleaner justification graphs with unnecessary antecedents removed
- A natural feedback loop: review findings inform retraction, which triggers re-derivation with better justifications

benthomasson:
## Summary

- Adds `reasons review-beliefs` command that uses an LLM to evaluate derived beliefs on three axes: **validity** (does the conclusion follow from antecedents?), **sufficiency** (are antecedents enough?), and **necessity** (are all antecedents load-bearing?)
- New module `reasons_lib/review.py` with batched LLM classification following the `list_negative` pattern
- API layer with filtering by belief IDs, min depth, depends-on, sample size, and access tags
- CLI supports `--dry-run`, `--auto-retract`, `--output`, `--sample`, `--min-depth`, `--depends-on`

## Test plan

- [x] 22 new tests in `tests/test_review.py` (formatting, parsing, batching, API filtering, summary counts)
- [x] Full suite passes (753 passed, 106 skipped)
- [ ] Manual: `reasons review-beliefs --sample 5 --dry-run`
- [ ] Manual: `reasons review-beliefs --min-depth 10 --dry-run`

Closes #96

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1690,6 +1690,89 @@ def list_negative(
         }
 
 
+def review_beliefs(
+    belief_ids: list[str] | None = None,
+    model: str = "claude",
+    timeout: int = 300,
+    min_depth: int | None = None,
+    depends_on: str | None = None,
+    sample: int | None = None,
+    visible_to: list[str] | None = None,
+    db_path: str = DEFAULT_DB,
+) -> dict:
+    """Review derived beliefs for validity, sufficiency, and necessity.
+
+    Uses an LLM to evaluate whether each derived belief's reasoning
+    from antecedents to conclusion is sound.
+
+    Returns: {"results": [...], "reviewed": int, "invalid": int,
+              "insufficient": int, "unnecessary": int, "total_derived": int}
+    """
+    from .derive import _get_depth
+    from .review import review_beliefs as _review
+
+    result = export_network(db_path=db_path)
+    nodes = result.get("nodes", {})
+
+    all_derived = {
+        k: v for k, v in nodes.items()
+        if v.get("truth_value") == "IN"
+        and v.get("justifications")
+        and len(v["justifications"]) > 0
+    }
+    total_derived = len(all_derived)
+
+    candidates = dict(all_derived)
+
+    if belief_ids:
+        candidates = {k: v for k, v in candidates.items() if k in belief_ids}
+
+    if visible_to is not None:
+        tags = set(visible_to)
+        candidates = {
+            k: v for k, v in candidates.items()
+            if not v.get("metadata", {}).get("access_tags")
+            or all(t in tags for t in v["metadata"]["access_tags"])
+        }
+
+    if min_depth is not None:
+        memo = {}
+        candidates = {
+            k: v for k, v in candidates.items()
+            if _get_depth(k, nodes, all_derived, memo) >= min_depth
+        }
+
+    if depends_on:
+        candidates = {
+            k: v for k, v in candidates.items()
+            if any(
+                depends_on in j.get("antecedents", [])
+                for j in v.get("justifications", [])
+            )
+        }
+
+    if sample is not None and len(candidates) > sample:
+        import random
+        sampled_keys = random.sample(sorted(candidates.keys()), sample)
+        candidates = {k: candidates[k] for k in sampled_keys}
+
+    review_ids = sorted(candidates.keys())
+    results = _review(nodes, belief_ids=review_ids, model=model, timeout=timeout)
+
+    invalid = sum(1 for r in results if not r.get("valid", True))
+    insufficient = sum(1 for r in results if not r.get("sufficient", True))
+    unnecessary = sum(1 for r in results if not r.get("necessary", True))
+
+    return {
+        "results": results,
+        "reviewed": len(review_ids),
+        "invalid": invalid,
+        "insufficient": insufficient,
+        "unnecessary": unnecessary,
+        "total_derived": total_derived,
+    }
+
+
 def _rewrite_dependents(net, old_id: str, new_id: str):
     """Rewrite justifications that reference old_id to point at new_id.
 
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -940,6 +940,76 @@ def cmd_list_negative(args):
           f"({result['candidates']} candidates from {result['total']} IN nodes)")
 
 
+def cmd_review_beliefs(args):
+    model = getattr(args, "model", None) or "claude"
+    result = api.review_beliefs(
+        belief_ids=args.ids or None,
+        model=model,
+        timeout=args.timeout,
+        min_depth=args.min_depth,
+        depends_on=args.depends_on,
+        sample=args.sample,
+        visible_to=_parse_visible_to(args),
+        db_path=args.db,
+    )
+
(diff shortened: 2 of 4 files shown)


benthomasson:
## Code Review (Claude)

**Gate: BLOCK** (fixed in a5e1de6)

### Findings

| File | Verdict | Notes |
|------|---------|-------|
| `reasons_lib/review.py` (new) | PASS | Well-structured, follows `list_negative` pattern. Only first justification reviewed per belief (reasonable for v1). |
| `review.py:parse_review_response` | CONCERN | Greedy regex `r"\[.*\]"` could miss JSON if LLM wraps response in prose containing `[brackets]`. Unlikely given prompt instructions but fragile. |
| `reasons_lib/api.py:review_beliefs` | PASS | Clean filtering pipeline. Correct design: `visible_to` filters reviewed beliefs, not context nodes. |
| `reasons_lib/cli.py:cmd_review_beliefs` | ~~BLOCK~~ **FIXED** | Called `api.retract()` instead of `api.retract_node()`. Fixed in a5e1de6. |
| `cli.py` argparse wiring | PASS | Matches issue interface sketch. |
| `tests/test_review.py` (new) | CONCERN | 22 tests cover core module well. Missing: `--auto-retract` path, `--output` writing, `--dry-run` behavior, `visible_to` filter. |
| `--depends-on` semantics | CONCERN | Only checks direct antecedents, not transitive. Issue says "cascade check" implying transitive. Reasonable for v1. |

### Fixed

- **`api.retract()` → `api.retract_node()`** — would have raised `AttributeError` at runtime on `--auto-retract`. Fixed in a5e1de6.

### Open concerns (non-blocking)

1. **Regex fragility** in `parse_review_response` — greedy `\[.*\]` with `re.DOTALL` could fail if LLM response has prose `[brackets]` before the JSON array. Consider scanning for `[` positions and trying `json.loads` at each.
2. **`--depends-on` is shallow** — only checks direct antecedents, not transitive dependency chain. Could be enhanced later.
3. **Test gaps** — no tests for `--auto-retract`, `--output`, `--dry-run`, `visible_to` filter paths.

Gemini review failed (workspace trust config issue — not a code problem).

benthomasson:
## Note: --any mode justification review

The current implementation reviews each justification independently, which works well for manually-added alternative justifications (different arguments with different antecedents). However, `--any` mode nodes have a different structure — one SL per antecedent, each with a single premise:

```
SL([A]) — IN if A is IN
SL([B]) — IN if B is IN
SL([C]) — IN if C is IN
```

For these, the meaningful review question is "is any of A, B, or C sufficient to support this claim?" rather than three separate "does this follow from A alone?" questions. The current approach is not wrong but asks the LLM a slightly awkward question for `--any` nodes.

Future improvement: detect `--any`-style justifications (multiple SLs each with exactly one antecedent) and format them as a single OR group rather than separate numbered justifications.

benthomasson:
## Code Review Round 3 (Claude + Gemini)

**Gate: CONCERN / PASS** — no BLOCKs.

### Verdicts

| File | Claude | Gemini |
|------|--------|--------|
| `reasons_lib/review.py` | PASS | PASS |
| `reasons_lib/api.py:review_beliefs` | CONCERN | PASS |
| `reasons_lib/cli.py:cmd_review_beliefs` | PASS | PASS |
| `reasons_lib/cli.py` (argparser) | PASS | PASS |
| `tests/test_review.py` | PASS | PASS |

### Remaining concerns (non-blocking, all future enhancements)

1. **`visible_to` access tag leak** — antecedent text from restricted nodes appears in the LLM prompt even when `visible_to` excludes them. By design: you can't review a belief without seeing its antecedents.
2. **`reviewed` count vs actual** — counts beliefs sent for review, not ones the LLM returned results for. Minor accuracy issue when batches fail.
3. **Metadata persistence** — no way to mark beliefs as reviewed and skip them in future runs. Important for large networks.
4. **`--budget` not implemented** — `--sample` covers the same need for v1.
5. **`--depends-on` is shallow** — direct antecedents only, not transitive. Decided: single-level is correct for reviewing individual reasoning steps.

PR #99, opened 2026-05-06
Add contradictions command for LLM-powered nogood detection

benthomasson:
## Summary

- Adds `reasons contradictions` command that sends batches of IN beliefs to an LLM to find sets that cannot all be true simultaneously (nogoods)
- Follows the same 3-layer pattern as `review-beliefs`: module (`contradictions.py`) → API (`detect_contradictions()`) → CLI subcommand
- Detected contradictions can be auto-applied via `--auto-apply` for dependency-directed backtracking, or previewed with `--dry-run`

## Test plan

- [x] 25 new tests in `tests/test_contradictions.py` covering formatting, parsing, batching, API, and CLI layers
- [x] Full test suite passes (787 passed, 0 failed)
- [ ] Manual test: `reasons contradictions --sample 50 --dry-run`
- [ ] Manual test: `reasons contradictions --auto-apply --sample 10`

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1779,7 +1779,6 @@ def detect_contradictions(
     timeout: int = 300,
     sample: int | None = None,
     auto_apply: bool = False,
-    visible_to: list[str] | None = None,
     db_path: str = DEFAULT_DB,
 ) -> dict:
     """Detect contradictions between IN beliefs via LLM analysis.
@@ -1789,7 +1788,7 @@ def detect_contradictions(
     """
     from .contradictions import detect_contradictions as _detect
 
-    result = export_network(visible_to=visible_to, db_path=db_path)
+    result = export_network(db_path=db_path)
     nodes = result.get("nodes", {})
 
     candidates = {
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1773,6 +1773,77 @@ def review_beliefs(
     }
 
 
+def detect_contradictions(
+    belief_ids: list[str] | None = None,
+    model: str = "claude",
+    timeout: int = 300,
+    sample: int | None = None,
+    auto_apply: bool = False,
+    visible_to: list[str] | None = None,
+    db_path: str = DEFAULT_DB,
+) -> dict:
+    """Detect contradictions between IN beliefs via LLM analysis.
+
+    Returns: {"contradictions": [...], "checked": int, "found": int,
+              "applied": int, "total_in": int}
+    """
+    from .contradictions import detect_contradictions as _detect
+
+    result = export_network(db_path=db_path)
+    nodes = result.get("nodes", {})
+
+    candidates = {
+        k: v for k, v in nodes.items()
+        if v.get("truth_value") == "IN"
+    }
+    total_in = len(candidates)
+
+    if belief_ids:
+        candidates = {k: v for k, v in candidates.items() if k in belief_ids}
+
+    if visible_to is not None:
+        tags = set(visible_to)
+        candidates = {
+            k: v for k, v in candidates.items()
+            if not v.get("metadata", {}).get("access_tags")
+            or all(t in tags for t in v["metadata"]["access_tags"])
+        }
+
+    if sample is not None and len(candidates) > sample:
+        import random
+        sampled_keys = random.sample(sorted(candidates.keys()), sample)
+        candidates = {k: candidates[k] for k in sampled_keys}
+
+    check_ids = sorted(candidates.keys())
+    contradictions = _detect(nodes, belief_ids=check_ids, model=model,
+                            timeout=timeout)
+
+    applied = 0
+    applied_details = []
+    if auto_apply:
+        for c in contradictions:
+            try:
+                nogood_result = add_nogood(c["claims"], db_path=db_path)
+                applied += 1
+                applied_details.append({
+                    "id": c["id"],
+                    "nogood_id": nogood_result.get("nogood_id"),
+                    "changed": nogood_result.get("changed", []),
+                })
+            except Exception as e:
+                print(f"  WARN: failed to apply {c['id']}: {e}",
+                      file=__import__("sys").stderr)
+
+    return {
+        "contradictions": contradictions,
+        "checked": len(check_ids),
+        "found": len(contradictions),
+        "applied": applied,
+        "applied_details": applied_details,
+        "total_in": total_in,
+    }
+
+
 def _rewrite_dependents(net, old_id: str, new_id: str):
     """Rewrite justifications that reference old_id to point at new_id.
 
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1789,7 +1789,7 @@ def detect_contradictions(
     """
     from .contradictions import detect_contradictions as _detect
 
-    result = export_network(db_path=db_path)
+    result = export_network(visible_to=visible_to, db_path=db_path)
     nodes = result.get("nodes", {})
 
     candidates = {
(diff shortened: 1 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#99
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/contradictions.py (new file)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The module is well-structured and follows existing patterns (matches `review_beliefs` in structure). Two concerns:

1. **Cross-batch contradictions are invisible.** Beliefs are shuffled then split into batches of 50. A contradiction between belief A (batch 1) and belief B (batch 2) will never be detected because the LLM only sees one batch at a time. The `random.shuffle` mitigates this probabilistically across runs, and the commit message explicitly calls this out ("probabilistic cross-batch coverage"), so this is a known design tradeoff — but there's no documentation in the code or CLI help about this limitation. Users may assume all pairs are checked.

2. **No deduplication of detected nogoods.** If the same pair `(A, B)` is detected across multiple runs or batches, `all_results.extend(results)` will produce duplicates. The `auto_apply` path in `api.py` will then call `add_nogood` for the same pair multiple times. This may be harmless if `add_nogood` is idempotent, but it's worth noting.

The regex parsing (`_NOGOOD_PATTERN`) is solid — it correctly handles multi-line body content and multiple NOGOOD sections. The 200-char text truncation is a good guard against token bloat.

### reasons_lib/contradictions.py:parse_contradiction_response
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Handles all expected cases: multiple NOGOODs, filtering against valid IDs, minimum 2-claim requirement, malformed responses ("No contradictions detected."). The backtick-stripping (`c.strip().strip("\`")`) is a smart touch since the prompt formats IDs in backticks and the LLM may echo them back. Tests cover all these paths thoroughly.

### reasons_lib/api.py:detect_contradictions
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Two issues:

1. **`auto_apply` silently swallows errors.** When `add_nogood` raises, the exception is printed to stderr but the contradiction is not included in `applied_details`. The caller has no way to distinguish "0 applied because none found" from "0 applied because all failed." The return dict should include a `failed` count or list.

2. **Lazy import of `random` inside a conditional.** Line 1802 imports `random` inside `if sample is not None`, while `contradictions.py` imports `random` at module level. This is inconsistent but not a bug — it just looks accidental. Minor.

The filtering logic is correct: first filter to IN beliefs, then optionally filter by `belief_ids`, then optionally sample. The `sorted()` before sampling ensures deterministic results with a fixed seed (though no seed is set).

### reasons_lib/cli.py:cmd_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean CLI handler following the pattern established by `cmd_review_beliefs`. The `--dry-run` flag correctly prevents auto-apply by computing `auto_apply = args.auto_apply and not args.dry_run`. The `--output` flag writes a clean markdown report. The argument parser is properly wired into the subcommand dispatch dict. One minor note: there's no `--visible-to` flag here unlike `review-beliefs`, which was explicitly removed per commit `becd379` — this is intentional.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Parser is correctly placed between `review-beliefs` and `list` in the subcommand registration. All arguments have sensible defaults (`timeout=300`, `model=None` falling back to "claude"). The dispatch dict entry at line 1442 correctly maps `"contradictions"` to `cmd_contradictions`.

### tests/test_contradictions.py (new file)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Thorough test coverage across all three layers: unit tests for the parser/formatter, integration tests for the API function, and CLI end-to-end tests. Key scenarios covered: batching behavior, empty networks, timeout passthrough, batch failures (graceful degradation), IN-only filtering, sampling, auto-apply, dry-run prevention, markdown output, and the "no contradictions" path. The mocking strategy (patching `reasons_lib.llm.subprocess.run`) is correct — it mocks at the right level to test the actual integration without hitting a real LLM.

One minor concern: `test_batches_beliefs` asserts `mock_run.call_count == 2` for 3 IN beliefs with batch_size=2, but because `random.shuffle` is called on belief_ids, the batch composition is non-deterministic. The count assertion is stable (always 2 batches regardless of shuffle order), so this is fine.

### Self-Review
**Limitations:** Could not verify whether `add_nogood` is idempotent when called with the same node pair twice (relevant to the duplicate detection concern). Could not see the full `review_beliefs` implementation to compare patterns more thoroughly. Did not verify whether the `invoke_model` subprocess call passes `input` as a kwarg (the test asserts `mock_run.call_args[1]["input"]` which assumes keyword argument — this was verified by the subagent).

### Feature Requests
- Show the full implementation of closely related functions (e.g., `add_nogood`, `review_beliefs`) when reviewing a new function that follows the same pattern — enables pattern-consistency checks
- Flag when mocked test assertions depend on internal implementation details (like `call_args[1]["input"]`) that could break if the underlying library changes its calling convention

## gemini [CONCERN]

### reasons_lib/api.py:detect_contradictions
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function uses `sys.stderr` at line 1819 to report failures during `auto_apply`, but `sys` is not imported within the function scope. While it may be imported at the top of the file, other dependencies in this function (like `random` and `_detect`) are imported locally, suggesting a pattern of localizing imports for new dependencies. If `sys` is missing at the top level, this will cause a crash when an error occurs. Additionally, printing to `stderr` from a library function is generally less desirable than using logging or returning error details in the response.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The CLI implementation is complete and follows existing patterns. It correctly wires the new `contradictions` command, handles optional sampling and auto-application, and provides both console and markdown output.

### reasons_lib/contradictions.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The core logic for LLM-based detection is sound. It includes proper batching to handle large networks, a well-defined prompt to minimize false positives, and robust regex-based parsing of the LLM output. The use of `random.shuffle` on batches helps surface different potential contradictions across multiple runs.

### tests/test_contradictions.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The test suite is excellent, covering unit-level parsing logic (including edge cases like single-claim nogoods or invalid IDs) and integration-level API and CLI behavior. Mocking is appropriately used to simulate LLM responses.

### Self-Review
**Limitations:** I did not have visibility into the top of `reasons_lib/api.py` to confirm if `sys` is imported globally. I am also assuming `add_nogood` and `export_network` are available in the `api.py` global scope as they are called without prefixes.

### Feature Requests
- Include the top-level imports of modified files in the diff to verify dependency availability.
- Provide a summary of existing library patterns (e.g., "always use logging" vs "printing to stderr is acceptable") to better evaluate library/API changes.

PR #104, opened 2026-05-06
Add update command to edit belief text in place

### LINKED ISSUE #103: Add update command to edit belief text in place
## Problem

There is no way to update a belief's text without either:
1. supersede -- creates a new node ID and rewires dependents (overkill for text fixes)
2. Direct SQL -- bypasses the API layer
3. Retract + re-add -- fails because the node already exists in retracted state

## Use case

When a premise is imprecise (e.g. "commits at the end" should be "commits or rolls back at the end"), the fix is a text edit, not a new belief. The current workflow requires a direct DB update which is error-prone and skips validation.

## Suggested interface

```
reasons update <id> --text "new text"
```

Should update the text of an existing node without changing its truth value, justifications, or dependents. Should work on both IN and OUT nodes.

## Context

Encountered while resolving contradiction pgapi-commit-vs-whatif-rollback -- needed to fix an imprecise premise but had no command to do it.

benthomasson:
## Summary

- Adds `reasons update <id> <text>` command to modify a node's text without superseding or re-creating it
- Also supports `--source` and `--source-url` flags for updating source metadata
- Follows existing API/CLI patterns: `update_node()` in api.py, `cmd_update()` in cli.py

Closes #103

## Test plan

- [x] 4 API tests: text update, source update, nonexistent raises, justifications preserved
- [x] 2 CLI tests: successful update, nonexistent error
- [x] Full test suite passes (792 passed, 0 failed)

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -632,6 +632,34 @@ def supersede(old_id: str, new_id: str, db_path: str = DEFAULT_DB) -> dict:
         return net.supersede(old_id, new_id)
 
 
+def update_node(
+    node_id: str,
+    text: str | None = None,
+    source: str | None = None,
+    source_url: str | None = None,
+    db_path: str = DEFAULT_DB,
+) -> dict:
+    """Update a node's text, source, or source_url in place.
+
+    Returns: {"node_id": str, "updated_fields": list[str]}
+    """
+    with _with_network(db_path, write=True) as net:
+        if node_id not in net.nodes:
+            raise KeyError(f"Node '{node_id}' not found")
+        node = net.nodes[node_id]
+        updated = []
+        if text is not None:
+            node.text = text
+            updated.append("text")
+        if source is not None:
+            node.source = source
+            updated.append("source")
+        if source_url is not None:
+            node.source_url = source_url
+            updated.append("source_url")
+        return {"node_id": node_id, "updated_fields": updated}
+
+
 def challenge(
     target_id: str,
     reason: str,
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -318,11 +318,15 @@ def cmd_supersede(args):
 
 
 def cmd_update(args):
+    if not any([args.text, args.source, args.source_url]):
+        print("Error: at least one of --text, --source, or --source-url required",
+              file=sys.stderr)
+        sys.exit(1)
     try:
         result = api.update_node(
             args.node_id, text=args.text,
-            source=getattr(args, "source", None),
-            source_url=getattr(args, "source_url", None),
+            source=args.source,
+            source_url=args.source_url,
             db_path=args.db,
         )
     except KeyError as e:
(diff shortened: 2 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#104
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### reasons_lib/api.py:update_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean implementation. Uses `_with_network(write=True)` which ensures changes persist to SQLite via `store.save()`. The `is not None` checks correctly distinguish "not provided" from "set to empty string". Raises `KeyError` for missing nodes, consistent with the pattern used by `supersede`, `challenge`, etc. Does not touch truth value, justifications, or dependents — exactly as the issue requires.

### reasons_lib/cli.py:cmd_update
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The validation `not any([args.text, args.source, args.source_url])` is truthiness-based, not None-based. If a user passes `--text ""` (empty string), `any()` treats it as falsy, so the CLI rejects it — but the API layer would accept it (it checks `is not None`). This is a minor inconsistency: the CLI guard is stricter than the API. In practice, setting text to empty string is probably not a real use case, but it's a semantic mismatch between the two layers. The `--text` flag approach (rather than positional) is the right call since source-only updates are supported.

### reasons_lib/cli.py (parser registration + dispatch)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Parser registered at the right location with `--text`, `--source`, `--source-url` flags. Handler mapped in the `commands` dict. Follows the exact same pattern as `supersede`, `challenge`, etc.

### tests/test_api.py:TestUpdateNode
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Five tests cover the key scenarios: text update, source update, nonexistent node error, justification preservation after update, and updating an OUT node. The `test_preserves_justifications` test is particularly good — it verifies the core invariant from the issue ("without changing truth value, justifications, or dependents"). Missing: no test for `source_url` update or for updating multiple fields simultaneously, but the existing coverage is solid for the core behavior.

### tests/test_cli.py:TestCmdUpdate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Four CLI tests cover: text update end-to-end (including `show` verification), nonexistent node error, source-only update, and no-flags validation. Good coverage of error paths. Uses `run_cli` helper consistently with the rest of the test file.

### Self-Review
**Limitations:** I verified persistence by reading the `_with_network` context manager and `storage.save()` — the full save path looks correct (it saves all node attributes including text/source/source_url). I did not verify whether `source_hash` should be cleared or updated when `source` changes — this could cause `check-stale` to report false results if the source path changes but the old hash remains. This is an edge case worth considering but not a blocker.

### Feature Requests
- Include the storage layer's save method in the diff context when API functions modify node attributes — helps verify persistence correctness without manual lookup.

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

Model invocation failed: Model gemini timed out after 300s

PR #116, opened 2026-05-08
Preserve justification chains for OUT beliefs on import/sync

### LINKED ISSUE #117: Import/sync strips justification chains from OUT beliefs
## Problem

When importing beliefs from one `reasons.db` into another via `import-agent` or `sync-agent`, OUT beliefs have their justifications stripped. This means:

- `list-gated` in the target DB can't show gated beliefs from the source
- `explain` can't trace why an imported belief is OUT
- `what-if retract` can't simulate fixing source-internal blockers from the target

## Root cause

Two code paths in `import_agent.py` discard justifications for OUT beliefs:

1. `_normalize_json` (line 144): `if is_out: raw_justs = []`
2. `_build_justifications` (line 222): `if claim["is_out"]: return []`

Additionally, `_import_claims` and `_sync_claims` force-retract all OUT beliefs regardless of whether they're retracted premises or gated (justification invalid).

## Impact

In the redhat-expert project, 6 cluster DBs have 1,455 OUT beliefs with full justification chains. After import to the top-level DB, all gating structure was lost — `list-gated` showed only the 6 agent kill-switch pairs instead of 141 real blockers gating 308 beliefs.

## Fix

PR #116 — preserve justifications for OUT beliefs and only force-retract true premise retractions (OUT with no justifications).

benthomasson:
## Summary

Fixes #117

- Import and sync previously stripped justifications from OUT beliefs, making gated beliefs appear as simple retractions in the target DB
- This lost the "would be IN if X were fixed" information, breaking `list-gated`, `explain`, and `what-if` at the top level
- Now only true premise retractions (OUT with no justifications) are force-retracted
- Gated beliefs (OUT because justification is invalid) keep their justification chains

## Test plan

- [x] Re-imported all 6 clusters into top-level redhat-expert DB
- [x] `list-gated` now shows 141 blockers gating 308 beliefs (was 6 kill-switch pairs)
- [x] `explain` traces through cluster justification chains
- [x] Belief counts unchanged (11,933 IN / 1,659 OUT) — no truth value changes, just preserved structure
- [x] Retracted premises (no justifications) still properly OUT

### DIFF
--- a/reasons_lib/import_agent.py
+++ b/reasons_lib/import_agent.py
@@ -382,8 +382,11 @@ def _sync_claims(network, agent_name, claims, source_path, nogoods):
             node.metadata["imported_from"] = source_path
 
             if is_out and not claim["raw_justifications"]:
-                if not _justifications_match(node.justifications, []):
-                    _update_node_justifications(network, node_id, [])
+                new_justs = _build_justifications(
+                    claim, prefix, inactive_id, agent_name
+                )
+                if not _justifications_match(node.justifications, new_justs):
+                    _update_node_justifications(network, node_id, new_justs)
                     changed = True
                 retract_after.append(node_id)
             elif is_out:
--- a/reasons_lib/import_agent.py
+++ b/reasons_lib/import_agent.py
@@ -141,17 +141,14 @@ def _normalize_json(data, only_in=False):
         is_out = ndata.get("truth_value") == "OUT"
         meta = dict(ndata.get("metadata", {}))
 
-        if is_out:
-            raw_justs = []
-        else:
-            raw_justs = []
-            for j in ndata.get("justifications", []):
-                raw_justs.append({
-                    "type": j.get("type", "SL"),
-                    "antecedents": list(j.get("antecedents", [])),
-                    "outlist": [o for o in j.get("outlist", []) if o in node_ids],
-                    "label": j.get("label"),
-                })
+        raw_justs = []
+        for j in ndata.get("justifications", []):
+            raw_justs.append({
+                "type": j.get("type", "SL"),
+                "antecedents": list(j.get("antecedents", [])),
+                "outlist": [o for o in j.get("outlist", []) if o in node_ids],
+                "label": j.get("label"),
+            })
 
         normalized.append({
             "id": nid,
@@ -219,9 +216,6 @@ def _topo_sort_claims(claims):
 
 def _build_justifications(claim, prefix, inactive_id, agent_name):
     """Build Justification objects from a normalized claim."""
-    if claim["is_out"]:
-        return []
-
     justs = []
     for rj in claim["raw_justifications"]:
         antecedents = [f"{prefix}{a}" for a in rj["antecedents"]]
@@ -309,7 +303,7 @@ def _import_claims(network, agent_name, claims, source_path, nogoods):
         )
         imported += 1
 
-        if claim["is_out"]:
+        if claim["is_out"] and not claim["raw_justifications"]:
             retract_after.append(node_id)
 
     nogoods_imported = _import_nogoods(network, prefix, nogoods)
@@ -387,11 +381,18 @@ def _sync_claims(network, agent_name, claims, source_path, nogoods):
                     node.metadata[k] = v
             node.metadata["imported_from"] = source_path
 
-            if is_out:
+            if is_out and not claim["raw_justifications"]:
                 if not _justifications_match(node.justifications, []):
                     _update_node_justifications(network, node_id, [])
                     changed = True
                 retract_after.append(node_id)
+            elif is_out:
+                new_justs = _build_justifications(
+                    claim, prefix, inactive_id, agent_name
+                )
+                if not _justifications_match(node.justifications, new_justs):
+                    _update_node_justifications(network, node_id, new_justs)
+                    changed = True
             else:
                 new_justs = _build_justifications(
                     claim, prefix, inactive_id, agent_name
--- a/reasons_lib/import_agent.py
+++ b/reasons_lib/import_agent.py
@@ -307,6 +307,7 @@ def _import_claims(network, agent_name, claims, source_path, nogoods):
             retract_after.append(node_id)
         elif claim["is_out"]:
             network.nodes[node_id].metadata["_retracted"] = True
+            network.nodes[node_id].truth_value = "OUT"
 
     nogoods_imported = _import_nogoods(network, prefix, nogoods)
 
@@ -405,6 +406,9 @@ def _sync_claims(network, agent_name, claims, source_path, nogoods):
                 if not node.metadata.get("_retracted"):
                     node.metadata["_retracted"] = True
                     changed = True
+                if node.truth_value != "OUT":
+                    node.truth_value = "OUT"
+                    changed = True
             else:
                 new_justs = _build_justifications(
                     claim, prefix, inactive_id, agent_name
@@ -455,6 +459,7 @@ def _sync_claims(network, agent_name, claims, source_path, nogoods):
                 # matching the existing-node sync path which relies on
                 # _retracted to keep OUT-with-justifications nodes locked.
                 network.nodes[node_id].metadata["_retracted"] = True
+                network.nodes[node_id].truth_value = "OUT"
 
     beliefs_removed = 0
     removed_ids = local_agent_ids - remote_ids
--- a/reasons_lib/import_agent.py
+++ b/reasons_lib/import_agent.py
@@ -528,8 +528,10 @@ def import_agent(
     'agent_name:inactive' (IN when active is OUT). Imported beliefs have
     inactive in their outlist — retracting active cascades everything OUT.
 
-    Beliefs that are OUT/STALE in the source are imported as bare premises
-    with no justification, so recompute_all cannot resurrect them.
+    Beliefs that are OUT/STALE in the source are marked with _retracted
+    metadata so recompute_all cannot resurrect them. JSON imports preserve
+    justifications on OUT nodes (enabling future resurrection when the remote
+    flips to IN); markdown imports strip them (bare premise path).
     """
     claims = _normalize_markdown(beliefs_text, only_in)
     nogoods = _normalize_nogoods_markdown(nogoods_text)
--- a/reasons_lib/import_agent.py
+++ b/reasons_lib/import_agent.py
@@ -440,8 +440,13 @@ def _sync_claims(network, agent_name, claims, source_path, nogoods):
             )
             beliefs_added += 1
 
-            if is_out:
+            if is_out and not claim["raw_justifications"]:
                 retract_after.append(node_id)
+            elif is_out:
+                # Set _retracted directly instead of using retract_after,
+                # matching the existing-node sync path which relies on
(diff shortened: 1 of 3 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#116
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### `reasons_lib/import_agent.py:_normalize_json`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The old code had an unnecessary `if is_out: raw_justs = [] / else: raw_justs = []` branch that produced the same empty list in both arms before populating it only in the `else` case. Removing the branch and always preserving justifications is correct — the truth value should be computed from justification validity, not by stripping justifications based on the source's current truth value. The outlist filtering (`if o in node_ids`) is correctly preserved.

### `reasons_lib/import_agent.py:_build_justifications`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Removing the early `if claim["is_out"]: return []` guard is the correct counterpart to the `_normalize_json` change. The `inactive_id` gating in the outlist is still applied uniformly, which is correct — all imported justifications should be gated on the agent being active.

### `reasons_lib/import_agent.py:_import_claims`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The condition change from `if claim["is_out"]` to `if claim["is_out"] and not claim["raw_justifications"]` is correct. OUT nodes without justifications are explicit retractions (premises) and should be retracted after import. OUT nodes with justifications should keep their chains — the TMS will naturally compute them as OUT if their antecedents are unsatisfied, and as IN if they become satisfied later. This enables restoration without re-import.

### `reasons_lib/import_agent.py:_sync_claims` (existing node path)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Two concerns:

**1. Missing `_retracted` handling for OUT-with-justifications (line 389-395).** The `elif is_out:` branch updates justifications but does not clear `_retracted` metadata. Since `_retracted` is a "sticky OUT" flag that causes `recompute_all()` to skip the node (confirmed in `network.py:771`), a node that was previously retracted and now has valid justifications would remain stuck OUT forever — `recompute_all` won't touch it, and it's not in `retract_after` or `assert_after`. Compare with the `else` branch (line 404) which correctly clears `_retracted`. The fix depends on desired semantics: if OUT-with-justifications should allow natural resurrection via TMS, `_retracted` should be cleared; if the remote's OUT status should be sticky, it should be added to `retract_after`.

**2. Duplicated justification-building logic (lines 389-395 vs 397-402).** The `elif is_out:` and `else:` branches have identical code for building and comparing justifications. This isn't a bug, but it's a missed opportunity to combine them — the only difference is the `_retracted`/`assert_after` handling afterward.

### `reasons_lib/import_agent.py:_sync_claims` (new node path, line 439)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Line 439 still uses the original `if is_out:` check without the `not claim["raw_justifications"]` guard. This means NEW nodes that are OUT-with-justifications during sync get added to `retract_after` unconditionally — they'll be explicitly retracted at line 464-471, which sets `_retracted=True`. This is inconsistent with the existing-node path (lines 389-395) which preserves justifications for OUT nodes without adding them to `retract_after`. This discrepancy means the same OUT-with-justifications belief behaves differently depending on whether it's a first sync or an update sync.

### Self-Review
**Limitations:** Could not see test files in the diff to verify whether new tests were added for the OUT-with-justifications scenarios. The explore agent confirmed no such tests exist, but the test files may have been modified in commits not shown in this diff. Could not verify the full `recompute_all` implementation to confirm `_retracted` skip behavior beyond what the explore agent reported.

### Feature Requests
- Include related test file diffs alongside implementation changes when they exist in the same commit
- Show full function bodies for modified functions (the diff hunks cut off surrounding context that's critical for understanding flow)
- Flag inconsistencies between parallel code paths (import vs sync, new-node vs existing-node) automatically

## gemini [CONCERN]

### reasons_lib/import_agent.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The changes allow 'OUT' claims to retain justifications during JSON import/sync, which is good for graph integrity. However, the logic in `_import_claims` (line 305) and `_sync_claims` (line 391) only appends to `retract_after` if `not claim["raw_justifications"]`. This means if a node is 'OUT' in the source but has justifications that happen to be satisfied locally, it will incorrectly come up 'IN' after import, violating the "remote wins" principle and causing inconsistency with Markdown imports (which still strip justifications for 'OUT' nodes via `_normalize_markdown`). Forcing a retraction via `retract_after` is necessary to honor the source truth value while still preserving the structural justifications. Additionally, `_normalize_markdown` should be updated for parity with `_normalize_json` to also preserve justifications for 'OUT' claims.

### reasons_lib/import_agent.py:_normalize_json
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly updated to preserve `raw_justifications` regardless of the `is_out` status, which allows for a more lossless import from JSON sources.

### reasons_lib/import_agent.py:_build_justifications
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly updated to build justifications even for 'OUT' claims, supporting the goal of preserving reasoning structure.

### reasons_lib/import_agent.py:_sync_claims
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Inconsistent logic: at line 391, it only retracts existing 'OUT' nodes if they have no justifications. However, at line 434, it always appends new 'OUT' nodes to `retract_after` regardless of justifications. The check `and not claim["raw_justifications"]` at line 391 should be removed to ensure the remote truth value is always honored.

### Self-Review
**Limitations:** I verified the TMS behavior in `reasons_lib/network.py` and confirmed that `retract()` sets `_retracted: True` without clearing justifications. This confirms that the proposed fix (always retracting 'OUT' nodes) would correctly preserve both the truth value and the reasoning structure. I was not able to run the tests to see the failure case for JSON sync of an 'OUT' node that becomes 'IN' locally, but the logic analysis strongly suggests this failure mode.

### Feature Requests
- Automatically highlight inconsistencies between similar functions (like `_normalize_json` and `_normalize_markdown`).
- Provide the ability to run specific tests and see the output during the review process.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#116
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [BLOCK]

### ERROR
**Verdict:** BLOCK

Model invocation failed: Model claude timed out after 300s

## gemini [CONCERN]

### reasons_lib/import_agent.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The changes successfully enable preservation of justifications for `OUT` nodes during JSON import, but the implementation has several logic gaps that could lead to incorrect truth values.

1.  **Missing Lock in `_import_claims`**: If a node is `OUT` in the source but has justifications, it is not added to `retract_after` (line 305). If its justifications happen to be satisfied locally, the node will end up `IN` despite the remote saying `OUT`. It should be added to `retract_after` regardless of whether it has justifications.
2.  **Inconsistent `_sync_claims` for New Nodes**: For new nodes that are `OUT` with justifications, the code sets `_retracted` metadata directly (line 448) but does not ensure the `truth_value` is `OUT`. If `add_node` computed it as `IN`, it will stay `IN` but locked. Using `retract_after` would be safer as it handles both the flag and the truth value.
3.  **Broken `_sync_claims` for Existing Nodes**: If an existing node transitions from `IN` to `OUT` on the remote and has justifications, it is neither added to `retract_after` nor is `_retracted` set (lines 388-397). The code assumes `_retracted` is already set, which is false if the node was previously `IN`.
4.  **Markdown Parity**: `_normalize_markdown` was not updated to match `_normalize_json`. While markdown is a simpler format, this creates a behavioral divergence where JSON imports preserve justifications for `OUT` nodes but markdown imports still strip them.

### tests/test_import_agent.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The test updates correctly reflect the new behavior where `OUT` nodes keep their justifications (specifically the default inactive gate). However, the test suite lacks a case for an `OUT` node that *has* justifications in the source, which would have caught the correctness issues identified in `import_agent.py`.

### Self-Review
**Limitations:** I verified the `network.py` implementation to confirm that `_retracted` is the correct mechanism for locking `OUT` nodes. The review focuses on the logical consistency of how this flag is applied across different import/sync paths. I did not have a specific issue or spec, so I judged based on the internal consistency of the RMS logic.

### Feature Requests
- Include a test case with an `OUT` node that has valid justifications in the source to verify it stays `OUT` locally.
- Ensure `_normalize_markdown` also preserves justifications if parity is desired.
- Consolidate the "locking" logic in `import_agent.py` to use `retract_after` (or a shared helper) for all `is_out` cases.

benthomasson:
## Code Review (Round 4) -- Claude only (Gemini trust directory failure)

**Gate: PASS** (9 pass, 1 concern)

### Changes reviewed
- `_normalize_json`: Removed dead branch, justifications now flow through -- **PASS**
- `_build_justifications`: Removed early-return for OUT, callers decide -- **PASS**
- `_import_claims`: Fixed `truth_value=OUT` after `add_node` -- **PASS** (minor concern: misleading propagation log entry)
- `_sync_claims` existing OUT path: Idempotent `_retracted` + `truth_value=OUT` -- **PASS**
- `_sync_claims` IN path clears `_retracted`: Enables resurrection -- **PASS**
- `_sync_claims` new OUT path: Mirrors import pattern -- **PASS**
- Docstring update: Accurately describes new behavior -- **PASS**
- Updated test assertions (supersession, JSON blocker) -- **PASS**
- `test_out_with_satisfied_justifications_stays_out`: Key regression test -- **PASS**
- `TestSyncOutWithJustifications` (3 tests): Full lifecycle coverage -- **PASS**

### Remaining concern (cosmetic)
`add_node()` computes and logs truth as IN before the override to OUT. The propagation log will show a misleading IN entry. Not a correctness bug -- could add a log entry when forcing OUT, or suppress initial computation for known-OUT nodes.

### Summary
All correctness issues from prior reviews are resolved. The `truth_value=OUT` fix addresses the round-3 concern about `_retracted` not forcing truth value. Docstring updated. Sync tests added for IN-to-OUT, OUT-to-IN, and idempotency paths.

PR #125, opened 2026-05-09
Add remove-justification command

### LINKED ISSUE #124: Add remove-justification command
## Problem

There is no way to remove a specific justification from a node. `retract` marks the node OUT but does not remove justifications — if a new justification is added, old ones still exist and can bring the node back IN.

This makes it impossible to replace an ungated justification with a gated one (using `--unless`), because the old ungated justification continues to satisfy independently.

## Use Case

When converting a belief to a GATE belief (adding an outlist condition), the workflow should be:
1. Add new justification with `--unless` outlist
2. Remove the old ungated justification

Without step 2, the gate has no effect.

## Proposed Solution

Add a `remove-justification` subcommand:

```bash
reasons remove-justification <node_id> --index N
```

Where `--index` refers to the justification index as shown by `reasons show <node_id>` (0-based). After removal, propagate truth values since the node may flip OUT.

Should also handle:
- Error if node has only one justification (would leave it unjustified — suggest `convert-to-premise` or `retract` instead)
- Error if index is out of range
- Display the justification being removed for confirmation

benthomasson:
## Summary

- Add `remove-justification` command to remove individual justifications by 0-based index
- Correctly cleans up dependents index (only removes refs not used by remaining justifications)
- Propagates truth value changes after removal
- Guards against removing from premises, single-justification nodes, and out-of-range indices
- 12 tests covering network, API, and CLI layers

Closes #124

## Test plan

- [x] `uv run pytest tests/test_remove_justification.py -v` — 12 tests pass
- [x] Full suite: 837 passed, 107 skipped
- [ ] Manual: `reasons show <node>` to see indices, `reasons remove-justification <node> <index>`

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -609,6 +609,14 @@ def convert_to_premise(node_id: str, db_path: str = DEFAULT_DB) -> dict:
         return net.convert_to_premise(node_id)
 
 
+def remove_justification(
+    node_id: str, index: int, db_path: str = DEFAULT_DB
+) -> dict:
+    """Remove a single justification by index and propagate."""
+    with _with_network(db_path, write=True) as net:
+        return net.remove_justification(node_id, index)
+
+
 def summarize(
     summary_id: str,
     text: str,
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -79,6 +79,27 @@ def cmd_add_justification(args):
         sys.exit(1)
 
 
+def cmd_remove_justification(args):
+    try:
+        result = api.remove_justification(
+            node_id=args.node_id,
+            index=args.index,
+            db_path=args.db,
+        )
+        removed = result["removed"]
+        ants = ", ".join(removed["antecedents"])
+        label = f" [{removed['label']}]" if removed["label"] else ""
+        print(f"Removed justification {args.index} from {result['node_id']}")
+        print(f"  Was: {removed['type']}({ants}){label}")
+        print(f"  Truth value: {result['old_truth_value']} → {result['new_truth_value']}")
+        print(f"  Remaining justifications: {result['remaining']}")
+        if result["changed"]:
+            print(f"  Cascade: {', '.join(result['changed'])}")
+    except (KeyError, ValueError, IndexError) as e:
+        print(f"Error: {e}", file=sys.stderr)
+        sys.exit(1)
+
+
 def _print_cascade(result):
     """Print cascade results, splitting went_out from went_in."""
     went_out = result.get("went_out", [])
@@ -1316,6 +1337,11 @@ def main():
     p.add_argument("--label", help="Justification label")
     p.add_argument("-n", "--namespace", help="Namespace prefix")
 
+    # remove-justification
+    p = sub.add_parser("remove-justification", help="Remove a justification by index")
+    p.add_argument("node_id", help="Node to remove justification from")
+    p.add_argument("index", type=int, help="0-based justification index (see 'show' output)")
+
     # retract
     p = sub.add_parser("retract", help="Retract a node (mark OUT + cascade)")
     p.add_argument("node_id", help="Node to retract")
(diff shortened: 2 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#125
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [PASS]

### `reasons_lib/network.py:remove_justification`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The method correctly follows the established pattern from `convert_to_premise` — validates inputs, pops the justification, selectively cleans dependents (only removing from `dependents` sets when the antecedent/outlist node no longer appears in any remaining justification), recomputes truth via `_compute_truth`, and propagates via BFS `_propagate`. The guard clauses match the issue spec: errors for missing node (KeyError), premise node (ValueError), single-justification node (ValueError with guidance to use `convert-to-premise` or `retract`), and out-of-range index (IndexError). Storage persistence is correct — the context manager in `_with_network` saves the network after mutation, and `storage.save()` clears and rewrites justifications, so the popped justification is simply not persisted.

### `reasons_lib/api.py:remove_justification`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Thin wrapper matching the established pattern — uses `_with_network(db_path, write=True)` context manager and delegates to `net.remove_justification()`. Consistent with every other mutating API function (`convert_to_premise`, `add_justification`, etc.).

### `reasons_lib/cli.py:cmd_remove_justification`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

CLI handler correctly reads `args.node_id`, `args.index`, `args.db`, and formats output showing the removed justification's type/antecedents/label, truth value transition, remaining count, and cascade. Exception handling catches KeyError, ValueError, and IndexError — matching the three exception types raised by `network.remove_justification`. The subparser is registered with positional `node_id` and `index` (typed as `int`), and the command is wired into the dispatch dict. The `--db` flag is inherited from the global parser.

### `reasons_lib/cli.py` (parser registration)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Parser correctly defines `node_id` as positional and `index` as positional with `type=int`. Placed logically after `add-justification`. Command entry added to dispatch dict.

### `tests/test_remove_justification.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test coverage across three layers (Network, API, CLI). Tests cover: removing one of two justifications (basic case), removal causing OUT transition, cascade propagation to dependents, dependent cleanup for antecedents no longer referenced, shared-antecedent preservation (critical edge case — if `a` appears in both justifications, removing one must NOT remove `a` from dependents), outlist dependent cleanup, and all four error cases (premise, single justification, bad index, missing node). The CLI test uses `subprocess.run` with `uv run reasons` and verifies both stdout output and the persisted state via `api.show_node`. 187 lines, 13 test methods — thorough.

### Self-Review
**Limitations:** Relied on subagent exploration of full method bodies rather than reading them directly — though the subagent returned complete code for all critical methods (`_compute_truth`, `_propagate`, `convert_to_premise`, `add_justification`, storage save/load), giving high confidence. Could not verify whether CLAUDE.md or other docs need updating to reflect the new command (the CLAUDE.md command list in the diff doesn't show changes, so it may need a separate update).

### Feature Requests
- Include the full file contents of closely related methods (e.g., `convert_to_premise`, `add_justification`) alongside the diff to make pattern-comparison easier without needing exploration
- Flag when project documentation (CLAUDE.md command lists, README) references the modified module but isn't updated in the PR

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

PR #129, opened 2026-05-10
Add cluster-list command for semantic belief clustering

benthomasson:
## Summary

- Adds `cluster-list` command that groups beliefs by semantic similarity using sentence-transformers + KMeans
- Supports `--format text|json|markdown` for doc generation pipelines
- Adds `list_clusters()` to `cluster.py` (reuses existing embed/KMeans flow) and `api.py`
- 3 new tests in `test_cluster.py`, full suite passes (858 passed, 131 skipped)

## Test plan

- [x] `uv run pytest tests/test_cluster.py -v` — all 12 tests pass (3 new)
- [x] `uv run pytest tests/ -x -q` — 858 passed, 131 skipped
- [x] Smoke tested against ftl-reasons-expert reasons.db (575 IN beliefs → 20 clusters)
- [x] Verified `--format json` produces valid parseable JSON
- [x] Verified `--format markdown` produces doc-ready output with headers

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2133,6 +2133,7 @@ def _rewrite_dependents(net, old_id: str, new_id: str):
 def list_clusters(
     status: str = "IN",
     n_clusters: int | None = None,
+    seed: int | None = None,
     embedding_model: str | None = None,
     visible_to: list[str] | None = None,
     db_path: str = DEFAULT_DB,
@@ -2155,6 +2156,7 @@ def list_clusters(
     return _list_clusters(
         beliefs,
         n_clusters=n_clusters,
+        seed=seed,
         model_name=embedding_model or DEFAULT_MODEL,
     )
 
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2130,6 +2130,35 @@ def _rewrite_dependents(net, old_id: str, new_id: str):
         old_node.dependents.discard(dep_id)
 
 
+def list_clusters(
+    status: str = "IN",
+    n_clusters: int | None = None,
+    embedding_model: str | None = None,
+    visible_to: list[str] | None = None,
+    db_path: str = DEFAULT_DB,
+) -> dict:
+    """Cluster beliefs by semantic similarity and return full assignments."""
+    from .cluster import list_clusters as _list_clusters, DEFAULT_MODEL
+
+    with _with_network(db_path) as net:
+        beliefs = {}
+        for nid, n in sorted(net.nodes.items()):
+            if status and n.truth_value != status:
+                continue
+            if visible_to is not None and not _is_visible(n, visible_to):
+                continue
+            beliefs[nid] = n.text
+
+    if not beliefs:
+        return {"clusters": [], "n_clusters": 0, "embedding_model": embedding_model or DEFAULT_MODEL}
+
+    return _list_clusters(
+        beliefs,
+        n_clusters=n_clusters,
+        model_name=embedding_model or DEFAULT_MODEL,
+    )
+
+
 def deduplicate(
     threshold: float = 0.5,
     auto: bool = False,
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -731,6 +731,7 @@ def cmd_cluster_list(args):
     result = api.list_clusters(
         status=args.status,
         n_clusters=args.n_clusters,
+        seed=args.seed,
         embedding_model=args.embedding_model,
         visible_to=_parse_visible_to(args),
         db_path=args.db,
@@ -1672,6 +1673,8 @@ def main():
                    help="Filter by truth value (default: IN)")
     p.add_argument("--n-clusters", type=int, default=None,
                    help="Override automatic cluster count")
+    p.add_argument("--seed", type=int, default=None,
+                   help="Random seed for reproducible clustering")
     p.add_argument("--embedding-model", default=None,
                    help="Sentence-transformers model (default: all-MiniLM-L6-v2)")
     p.add_argument("--visible-to", metavar="TAG,TAG",
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -763,7 +763,8 @@ def cmd_cluster_list(args):
         for b in cluster["beliefs"]:
             marker = "+" if args.status == "IN" else "-"
             print(f"  [{marker}] {b['id']}")
-            print(f"      {b['text'][:100]}")
+            text = b['text'][:100] + "..." if len(b['text']) > 100 else b['text']
+            print(f"      {text}")
 
     print(f"\n{result['n_clusters']} cluster(s), {total} beliefs")
     print(f"Model: {result['embedding_model']}")
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -740,7 +740,7 @@ def cmd_cluster_list(args):
         print("No beliefs to cluster.")
         return
 
-    fmt = getattr(args, "format", "text")
+    fmt = args.format
 
     if fmt == "json":
         print(json.dumps(result, indent=2))
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -726,6 +726,48 @@ def cmd_ask(args):
     print(result)
 
 
+def cmd_cluster_list(args):
+    _require_sqlite(args, "cluster-list")
+    result = api.list_clusters(
+        status=args.status,
+        n_clusters=args.n_clusters,
+        embedding_model=args.embedding_model,
+        visible_to=_parse_visible_to(args),
+        db_path=args.db,
+    )
+
+    if not result["clusters"]:
+        print("No beliefs to cluster.")
+        return
+
+    fmt = getattr(args, "format", "text")
+
+    if fmt == "json":
+        print(json.dumps(result, indent=2))
+        return
+
+    if fmt == "markdown":
+        for i, cluster in enumerate(result["clusters"], 1):
+            size = len(cluster["beliefs"])
+            print(f"\n## Cluster {i} ({size} belief{'s' if size != 1 else ''})\n")
+            for b in cluster["beliefs"]:
+                print(f"- **{b['id']}**: {b['text']}")
+        return
+
+    total = 0
+    for i, cluster in enumerate(result["clusters"], 1):
+        size = len(cluster["beliefs"])
+        total += size
+        print(f"\nCluster {i} ({size} belief{'s' if size != 1 else ''}):")
+        for b in cluster["beliefs"]:
+            marker = "+" if args.status == "IN" else "-"
+            print(f"  [{marker}] {b['id']}")
+            print(f"      {b['text'][:100]}")
+
+    print(f"\n{result['n_clusters']} cluster(s), {total} beliefs")
+    print(f"Model: {result['embedding_model']}")
+
+
 def cmd_deduplicate(args):
     _require_sqlite(args, "deduplicate")
     if args.accept:
@@ -1624,6 +1666,19 @@ def main():
     p.add_argument("--accept", metavar="FILE",
                    help="Apply a reviewed dedup plan file")
 
+    # cluster-list
+    p = sub.add_parser("cluster-list", help="List semantic similarity clusters")
+    p.add_argument("--status", choices=["IN", "OUT"], default="IN",
+                   help="Filter by truth value (default: IN)")
+    p.add_argument("--n-clusters", type=int, default=None,
+                   help="Override automatic cluster count")
+    p.add_argument("--embedding-model", default=None,
+                   help="Sentence-transformers model (default: all-MiniLM-L6-v2)")
+    p.add_argument("--visible-to", metavar="TAG,TAG",
+                   help="Only show nodes whose access_tags are a subset of these tags")
+    p.add_argument("--format", choices=["text", "json", "markdown"], default="text",
+                   help="Output format (default: text)")
+
     # namespaces
     p = sub.add_parser("list-gated", help="List OUT nodes blocked by IN outlist nodes")
     p.add_argument("--visible-to", metavar="TAG,TAG", help="Only show nodes whose access_tags are a subset of these tags")
@@ -1740,6 +1795,7 @@ def main():
         "lookup": cmd_lookup,
         "ask": cmd_ask,
         "deduplicate": cmd_deduplicate,
+        "cluster-list": cmd_cluster_list,
         "list": cmd_list,
         "list-gated": cmd_list_gated,
         "list-negative": cmd_list_negative,
--- a/reasons_lib/cluster.py
+++ b/reasons_lib/cluster.py
@@ -81,7 +81,7 @@ def _auto_k(n_beliefs, n_clusters=None, max_k=20):
     if n_clusters is not None:
         k = max(n_clusters, 1)
     else:
-        k = len(range(n_beliefs)) // 5
+        k = n_beliefs // 5
         k = min(k, max_k)
         k = max(k, 2)
     return min(k, n_beliefs)
--- a/reasons_lib/cluster.py
+++ b/reasons_lib/cluster.py
@@ -141,3 +141,62 @@ def cluster_beliefs(beliefs, budget, seed=None, n_clusters=None,
         "cluster_sizes": cluster_sizes,
         "embedding_model": cache.model_name,
     }
+
+
+def list_clusters(beliefs, n_clusters=None, cache=None, model_name=DEFAULT_MODEL):
+    """Cluster beliefs and return full cluster assignments.
+
+    Args:
+        beliefs: {node_id: text} dict
+        n_clusters: override automatic cluster count
+        cache: optional ClusterCache for embedding reuse
+        model_name: sentence-transformers model name
+
+    Returns:
+        {"clusters": [{"id": int, "beliefs": [{"id": str, "text": str}]}],
+         "n_clusters": int, "embedding_model": str}
+    """
+    _require_cluster_deps()
+
+    if len(beliefs) <= 3:
+        return {
+            "clusters": [{"id": 0, "beliefs": [
+                {"id": nid, "text": text} for nid, text in sorted(beliefs.items())
+            ]}],
+            "n_clusters": 1,
+            "embedding_model": model_name,
+        }
+
+    if cache is None:
+        cache = ClusterCache(model_name)
+
+    ids, embeddings = cache.embed(beliefs)
+
+    if n_clusters is None:
+        k = len(beliefs) // 5
+        k = min(k, 20)
+        k = max(k, 2)
+    else:
+        k = max(n_clusters, 1)
+
+    k = min(k, len(beliefs))
+
+    km = KMeans(n_clusters=k, random_state=42, n_init=10)
+    labels = km.fit_predict(embeddings)
+
+    groups = {}
+    for i, nid in enumerate(ids):
+        groups.setdefault(int(labels[i]), []).append(
+            {"id": nid, "text": beliefs[nid]}
+        )
+
+    clusters = [
+        {"id": label, "beliefs": members}
+        for label, members in sorted(groups.items(), key=lambda x: -len(x[1]))
+    ]
+
+    return {
+        "clusters": clusters,
+        "n_clusters": k,
+        "embedding_model": cache.model_name,
+    }
--- a/reasons_lib/cluster.py
+++ b/reasons_lib/cluster.py
@@ -76,6 +76,17 @@ class ClusterCache:
         return ids, np.array(embeddings)
 
 
+def _auto_k(n_beliefs, n_clusters=None, max_k=20):
+    """Compute cluster count from belief count or explicit override."""
+    if n_clusters is not None:
+        k = max(n_clusters, 1)
+    else:
+        k = len(range(n_beliefs)) // 5
+        k = min(k, max_k)
+        k = max(k, 2)
+    return min(k, n_beliefs)
+
+
 def cluster_beliefs(beliefs, budget, seed=None, n_clusters=None,
                     cache=None, model_name=DEFAULT_MODEL):
     """Cluster beliefs and sample across cluster boundaries.
@@ -106,14 +117,7 @@ def cluster_beliefs(beliefs, budget, seed=None, n_clusters=None,
 
     ids, embeddings = cache.embed(beliefs)
 
-    if n_clusters is None:
-        k = len(beliefs) // 5
-        k = min(k, budget // 3, 20)
-        k = max(k, 2)
-    else:
-        k = max(n_clusters, 1)
-
-    k = min(k, len(beliefs))
+    k = _auto_k(len(beliefs), n_clusters, max_k=min(budget // 3, 20))
 
     km = KMeans(n_clusters=k, random_state=seed, n_init=10)
     labels = km.fit_predict(embeddings)
@@ -143,12 +147,14 @@ def cluster_beliefs(beliefs, budget, seed=None, n_clusters=None,
     }
 
 
-def list_clusters(beliefs, n_clusters=None, cache=None, model_name=DEFAULT_MODEL):
+def list_clusters(beliefs, n_clusters=None, seed=None, cache=None,
+                  model_name=DEFAULT_MODEL):
     """Cluster beliefs and return full cluster assignments.
 
     Args:
         beliefs: {node_id: text} dict
         n_clusters: override automatic cluster count
+        seed: random seed for KMeans
         cache: optional ClusterCache for embedding reuse
         model_name: sentence-transformers model name
 
@@ -172,16 +178,9 @@ def list_clusters(beliefs, n_clusters=None, cache=None, model_name=DEFAULT_MODEL
 
     ids, embeddings = cache.embed(beliefs)
 
-    if n_clusters is None:
-        k = len(beliefs) // 5
-        k = min(k, 20)
-        k = max(k, 2)
-    else:
-        k = max(n_clusters, 1)
+    k = _auto_k(len(beliefs), n_clusters)
 
-    k = min(k, len(beliefs))
-
-    km = KMeans(n_clusters=k, random_state=42, n_init=10)
+    km = KMeans(n_clusters=k, random_state=seed, n_init=10)
     labels = km.fit_predict(embeddings)
 
     groups = {}
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1,6 +1,6 @@
 """Tests for the functional Python API."""
 
-from unittest.mock import patch, MagicMock
+from unittest.mock import patch
 
 import pytest
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1,6 +1,6 @@
 """Tests for the functional Python API."""
 
-from unittest.mock import patch
+from unittest.mock import patch, MagicMock
 
 import pytest
 
@@ -577,3 +577,52 @@ class TestUpdateNode:
         node = api.show_node("a", db_path=db_path)
         assert node["text"] == "Updated while OUT"
         assert node["truth_value"] == "OUT"
+
+
+class TestListClusters:
+
+    @pytest.fixture
+    def db_with_beliefs(self, tmp_path):
+        db = str(tmp_path / "test.db")
+        api.init_db(db_path=db)
+        for i in range(10):
+            api.add_node(f"in-{i}", f"Active belief {i}", db_path=db)
+        for i in range(5):
+            api.add_node(f"out-{i}", f"Retracted belief {i}", db_path=db)
+            api.retract_node(f"out-{i}", db_path=db)
+        return db
+
+    def test_filters_by_status(self, db_with_beliefs):
+        mock_result = {"clusters": [{"id": 0, "beliefs": []}], "n_clusters": 1, "embedding_model": "test"}
+        with patch("reasons_lib.cluster.list_clusters", return_value=mock_result) as mock_lc:
+            api.list_clusters(status="IN", db_path=db_with_beliefs)
+            beliefs_arg = mock_lc.call_args[0][0]
+            assert all(k.startswith("in-") for k in beliefs_arg)
+            assert len(beliefs_arg) == 10
+
+    def test_filters_out_status(self, db_with_beliefs):
+        mock_result = {"clusters": [{"id": 0, "beliefs": []}], "n_clusters": 1, "embedding_model": "test"}
+        with patch("reasons_lib.cluster.list_clusters", return_value=mock_result) as mock_lc:
+            api.list_clusters(status="OUT", db_path=db_with_beliefs)
+            beliefs_arg = mock_lc.call_args[0][0]
+            assert all(k.startswith("out-") for k in beliefs_arg)
+            assert len(beliefs_arg) == 5
+
+    def test_empty_network(self, tmp_path):
+        db = str(tmp_path / "empty.db")
+        api.init_db(db_path=db)
+        result = api.list_clusters(db_path=db)
+        assert result["clusters"] == []
+        assert result["n_clusters"] == 0
+
+    def test_passes_seed(self, db_with_beliefs):
+        mock_result = {"clusters": [], "n_clusters": 0, "embedding_model": "test"}
+        with patch("reasons_lib.cluster.list_clusters", return_value=mock_result) as mock_lc:
+            api.list_clusters(seed=42, db_path=db_with_beliefs)
+            assert mock_lc.call_args[1]["seed"] == 42
+
+    def test_passes_n_clusters(self, db_with_beliefs):
+        mock_result = {"clusters": [], "n_clusters": 0, "embedding_model": "test"}
+        with patch("reasons_lib.cluster.list_clusters", return_value=mock_result) as mock_lc:
+            api.list_clusters(n_clusters=3, db_path=db_with_beliefs)
+            assert mock_lc.call_args[1]["n_clusters"] == 3
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -9,7 +9,7 @@ import pytest
 try:
     from reasons_lib.cluster import (
         cluster_beliefs, list_clusters, ClusterCache, _require_cluster_deps,
-        HAS_CLUSTER_DEPS,
+        _auto_k, HAS_CLUSTER_DEPS,
     )
 except ImportError:
     HAS_CLUSTER_DEPS = False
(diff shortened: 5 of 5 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#129
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **reasons_lib/cluster.py:list_clusters** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### reasons_lib/cluster.py:list_clusters
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Logic is sound — embeds beliefs, runs KMeans, groups results. The `<= 3` small-set short-circuit is a reasonable heuristic. Two minor concerns: (1) Hardcoded `random_state=42` instead of accepting a `seed` parameter like the sibling `cluster_beliefs()` function does — inconsistent API surface within the same module. (2) The automatic `k = len(beliefs) // 5` logic is duplicated from `cluster_beliefs()` (lines 109-112 vs 175-178) rather than extracted into a shared helper. Neither is blocking but both are worth noting.

### reasons_lib/api.py:list_clusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Follows the established pattern exactly — `_with_network` context manager, `_is_visible` filtering, status filtering by `n.truth_value`. The lazy import of the cluster module is consistent with how `cluster_beliefs` is imported elsewhere. Empty-beliefs edge case returns a well-formed dict matching the normal return shape. The one gap is that there are no direct tests for this API wrapper — filtering by status, visible_to filtering, and the empty-network path are only tested indirectly through the CLI or not at all.

### reasons_lib/cli.py:cmd_cluster_list
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Follows the same structure as `cmd_deduplicate` — calls `_require_sqlite`, dispatches to API, handles three output formats. The text format marker logic (`"+" if args.status == "IN" else "-"`) is correct for the two allowed status values. Text truncation at 100 chars is reasonable. Parser registration and command dispatch wiring are both present. No CLI-level tests exist, but this is consistent with the codebase — other CLI commands like `cmd_deduplicate` also lack direct CLI tests.

### reasons_lib/cli.py (argparse registration)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Arguments are properly typed (`--n-clusters` as int, `--status` with choices, `--format` with choices). Defaults are sensible. The parser is registered before the `namespaces` parser, and the command is added to the dispatch dict. All good.

### tests/test_cluster.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Three tests cover the key paths: normal clustering with all-beliefs-returned assertion, `n_clusters` override, and small-set short-circuit. The `test_list_clusters_returns_all_beliefs` test properly asserts that every input belief appears in exactly one cluster (set equality). The `skip_no_cluster` decorator is applied consistently. Import updated correctly to include `list_clusters`.

### Self-Review
**Limitations:** Could not verify whether the `_parse_visible_to` helper correctly handles the `--visible-to` argument format for this command. Could not see existing tests for `api.deduplicate()` to gauge whether the absence of `api.list_clusters()` tests is anomalous or consistent. Did not run the test suite to confirm the new tests pass.

### Feature Requests
- Include the full source of sibling functions (e.g., `cluster_beliefs`) inline so reviewers can verify pattern consistency without spawning subagents
- Show test coverage reports or at least which API functions have corresponding test files
- Flag duplicated logic blocks across functions in the same module (the auto-k calculation is copy-pasted)

## gemini [PASS]

### reasons_lib/api.py:list_clusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The function correctly implements the top-level API for semantic clustering. It handles database access via the established `_with_network` pattern, applies visibility filters consistently with other API functions, and uses local imports to avoid loading heavy machine learning dependencies when they aren't needed.

### reasons_lib/cli.py:cmd_cluster_list
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The CLI implementation is complete and follows existing patterns for command registration and output formatting (supporting text, JSON, and Markdown). It correctly handles empty results and provides a helpful summary at the end.

### reasons_lib/cluster.py:list_clusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The core clustering logic is robust. It includes a fallback for small datasets (<= 3 beliefs) and uses a reasonable heuristic for the number of clusters when not specified. The use of `KMeans` with a fixed `random_state` ensures stable results.

### tests/test_cluster.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New test cases cover the primary functionality, the cluster count override, and the small-set fallback. The use of the `skip_no_cluster` decorator ensures tests don't fail in environments without the optional dependencies.

### Self-Review
**Limitations:** I verified the presence and signatures of helper functions like `_is_visible` and `_with_network` via grep, as they were used but not defined in the provided diff hunks. I did not run the code as no environment was provided, but the logic is straightforward and follows established project patterns.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#129
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **reasons_lib/cli.py:cmd_cluster_list** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### reasons_lib/cluster.py:_auto_k
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

`len(range(n_beliefs)) // 5` on line 84 is a roundabout way to write `n_beliefs // 5`. `range()` creates an object only to immediately take its length, returning the same integer. This is dead weight — likely a copy artifact. The refactor itself is behavior-preserving for `cluster_beliefs`: the original `min(k, budget // 3, 20)` is equivalent to `min(k, min(budget // 3, 20))` which is what the new call `_auto_k(..., max_k=min(budget // 3, 20))` produces. The `_auto_k` helper has no direct unit tests — it is only tested indirectly through `cluster_beliefs` and `list_clusters` tests. A direct test would catch edge cases (e.g., `n_beliefs=0`, `n_clusters > n_beliefs`).

### reasons_lib/cluster.py:list_clusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Function follows the same embed-then-KMeans pattern as `cluster_beliefs` but returns all beliefs with their cluster assignments instead of sampling. The early return for `<= 3` beliefs is sound (KMeans with k=2 on 3 points is wasteful). Output structure is clean: clusters sorted by descending size, beliefs contain both id and text. Uses `_auto_k` without the `budget // 3` cap, which is correct since there's no budget concept here. Three tests cover the main paths: normal clustering, explicit n_clusters override, and the small-set early return.

### reasons_lib/api.py:list_clusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Properly uses `_with_network` context manager for read-only access, filters by status and visibility using established helpers (`_is_visible`). The lazy import of `cluster.list_clusters` follows the existing pattern (e.g., `deduplicate` does the same). Empty-beliefs early return is correct and avoids calling into cluster code with no data. The `seed` parameter is not exposed through this API layer — acceptable for a listing command where deterministic ordering isn't critical, but worth noting. No unit test at the `api.list_clusters` level, though the CLI integration test path covers it indirectly.

### reasons_lib/cli.py:cmd_cluster_list
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The text format output on line 759 hardcodes the marker based on `args.status` (`"+"` for IN, `"-"` otherwise). But `args.status` defaults to `"IN"`, which is fine — the issue is that `args.status` can only be `"IN"` or `"OUT"` per the argparse choices, so the ternary is always correct. Text output truncates belief text to 100 chars (line 760) which is reasonable. One concern: there are no CLI-level tests (e.g., testing `cmd_cluster_list` with mock args or via subprocess). The `--format` argument is accessed via `getattr(args, "format", "text")` at line 741, but argparse always sets it since there's a `default="text"`, so the `getattr` fallback is unnecessary (though harmless). The command is properly wired in the dispatch dict at line 1798.

### reasons_lib/cli.py (argparse setup)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Argparse definition is clean. `--n-clusters` with `type=int` correctly maps to `args.n_clusters` (argparse converts hyphens to underscores). `--embedding-model` similarly maps correctly. `--visible-to` follows the existing pattern used by other commands. All arguments are optional with sensible defaults.

### tests/test_cluster.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Three new tests cover the key paths for `list_clusters`: (1) all beliefs appear in output, (2) explicit `n_clusters` override is respected, (3) small-set early return produces a single cluster. Tests are properly decorated with `@skip_no_cluster` for environments without ML deps. Assertions check structural properties (set membership, cluster count) rather than order-dependent properties, so the lack of `seed` parameter in tests is not a flakiness risk.

### Self-Review
**Limitations:** Could not verify whether existing `cluster_beliefs` tests still pass after the `_auto_k` refactor without running the test suite. Could not see all callers of `cluster_beliefs` to confirm no other code relies on the old inline k-computation. Did not verify the full `_with_network` context manager behavior (write vs read-only) since I only saw a summary.

### Feature Requests
- Include the ability to run the test suite as part of review to verify refactors don't break existing tests
- Show full function bodies for modified functions in the observations section, not just diff hunks

## gemini [PASS]

### reasons_lib/api.py:list_clusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function correctly implements belief filtering by status and visibility before delegating to the clustering logic. It uses existing patterns like `_with_network` and `_is_visible`. The empty belief check ensures a consistent return type even when no nodes match the criteria.

### reasons_lib/cli.py:cmd_cluster_list
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The command handler is properly registered and wired to the API. It supports multiple output formats (text, JSON, markdown) and correctly handles the case where no beliefs are found. The use of `_parse_visible_to` and `_require_sqlite` follows existing project conventions.

### reasons_lib/cluster.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The refactoring of `_auto_k` successfully centralizes the cluster count logic for both sampling and listing. The `list_clusters` implementation is sound, handling small datasets gracefully by bypassing clustering. The sorting of clusters by size descending is a helpful UX improvement.

### tests/test_cluster.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New tests provide good coverage for the `list_clusters` functionality, including edge cases like small datasets and parameter overrides. The use of the `skip_no_cluster` decorator ensures tests are skipped appropriately in environments without optional dependencies.

### Self-Review
**Limitations:** I did not perform a live execution of the clustering logic as it requires `sentence-transformers`, but the unit tests and logic review confirm correctness assuming the underlying libraries behave as expected.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#129
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/cluster.py:`_auto_k`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Clean extraction of the k-computation logic. Behavior is equivalent to the inlined version in `cluster_beliefs` — verified that `min(k, budget // 3, 20)` maps to `max_k=min(budget // 3, 20)` then `min(k, max_k)`. Edge cases are handled: `n_clusters=0` clamps to 1, final `min(k, n_beliefs)` prevents k > n. No direct unit test for `_auto_k` itself — the edge cases (n_beliefs < 2, n_clusters exceeding n_beliefs) are only indirectly tested through `list_clusters` and `cluster_beliefs`.

### reasons_lib/cluster.py:`list_clusters`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correct parallel to `cluster_beliefs` that returns full assignments instead of sampling. The small-set threshold (<=3) matches `cluster_beliefs`. Uses `_auto_k` with default `max_k=20` (no budget concept here, which makes sense). Clusters sorted by descending size is a good UX choice. Three tests cover the main paths: all-beliefs assignment, n_clusters override, and small-set fallback.

### reasons_lib/api.py:`list_clusters`
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The API function correctly filters by status and visibility, then delegates to the cluster module. However, there is no API-level test — the tests in `test_cluster.py` only exercise `cluster.list_clusters` directly, skipping the filtering logic in `api.list_clusters`. A test verifying that status filtering and `visible_to` filtering work correctly before beliefs reach the clustering layer would be valuable. The `seed` parameter is accepted by `cluster.list_clusters` but not exposed in the API — fine for a listing command where reproducibility is less critical, but worth noting.

### reasons_lib/cli.py:`cmd_cluster_list`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Clean CLI handler with three output formats. The text format truncates belief text to 100 chars, which is reasonable. The `marker` is derived from `args.status` rather than per-belief status, but since beliefs are pre-filtered this is correct. The parser is properly registered in the dispatch table. No CLI-level tests, but this is consistent with how other commands in this codebase appear to be tested (via API tests rather than CLI integration tests).

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Parser arguments are well-defined with appropriate types, defaults, and help text. The command is registered in the dispatch dict. Placement in the parser block (after deduplicate, before namespaces) is logical.

### tests/test_cluster.py (new tests)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Three tests cover the key paths: all-belief assignment invariant (every belief appears in exactly one cluster), n_clusters override, and small-set fallback. The `skip_no_cluster` decorator correctly gates on optional dependencies. Missing: no test for `_auto_k` edge cases directly (e.g., n_beliefs=1, n_clusters > n_beliefs), and no test asserting cluster sort order (descending by size).

### Self-Review
**Limitations:** Could not verify `_parse_visible_to`, `_is_visible`, or `_with_network` implementations — assumed they work correctly based on existing usage in the codebase. Could not see existing API-level tests to confirm whether the lack of a test for `api.list_clusters` is a pattern violation or consistent with the project. Did not run the test suite to verify the `_auto_k` refactoring doesn't break existing `cluster_beliefs` tests.

### Feature Requests
- Include full function bodies for helper functions referenced by the diff (e.g., `_parse_visible_to`, `_is_visible`) so reviewers can verify argument passing
- Show existing test patterns for sibling features (e.g., `test_api.py` for `deduplicate`) to gauge whether missing API tests are a real gap or consistent with project convention

## gemini [PASS]

### reasons_lib/api.py:list_clusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The implementation correctly leverages `_with_network` and `_is_visible` to collect beliefs while respecting access controls. Sorting nodes by ID ensures a deterministic input to the clustering algorithm. Integration is standard for the project's API layer.

### reasons_lib/cli.py:cmd_cluster_list
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The command is correctly registered and handles multiple output formats (text, json, markdown) appropriately. It includes required SQLite backend validation and visibility tag parsing. One minor suggestion is to add a `--seed` argument for reproducibility, as KMeans is stochastic by default.

### reasons_lib/cluster.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The refactoring of `_auto_k` centralizes the cluster count logic effectively. The new `list_clusters` function follows the established pattern of embedding then clustering, and correctly handles small sets (<= 3 beliefs) with an early return. The use of `n_init=10` in KMeans is good practice for stability.

### tests/test_cluster.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New tests cover the core clustering logic, including the early return for small sets and the explicit cluster count override. The use of the `skip_no_cluster` marker maintains compatibility for environments without the optional dependencies.

### Self-Review
**Limitations:** I could not verify the end-to-end execution of the CLI command as I do not have the optional dependencies (`sentence-transformers`, `scikit-learn`) installed in the review environment. However, the logic matches existing patterns in `derive.py` which are known to work. I also noticed that while the core logic is tested, the API and CLI wrappers lack integration tests in `test_api.py` and `test_cli.py`, though this is common for thin wrappers in this project.

### Feature Requests
- Add a `--seed` argument to the `cluster-list` CLI command and `api.list_clusters` function to allow for reproducible clustering results, consistent with the `derive` command.
- Consider adding a basic integration test in `tests/test_cli.py` that mocks the `api.list_clusters` return to verify the CLI formatting logic.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#129
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **reasons_lib/cluster.py:_auto_k** [LOW]: claude: CONCERN, gemini: PASS
- **reasons_lib/api.py:list_clusters** [LOW]: claude: CONCERN, gemini: PASS
- **reasons_lib/cli.py:cmd_cluster_list** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### reasons_lib/cluster.py:_auto_k
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The extraction of `_auto_k` is a good refactor, but it introduces a subtle behavioral change to `cluster_beliefs`. Previously, the auto path used `k = min(k, budget // 3, 20)` — the budget constraint was part of the `min` chain. Now `cluster_beliefs` calls `_auto_k(len(beliefs), n_clusters, max_k=min(budget // 3, 20))`, which applies `max_k` correctly. However, `list_clusters` calls `_auto_k(len(beliefs), n_clusters)` with the default `max_k=20`, which means for `list_clusters` there's no budget-based cap. This is intentional since `list_clusters` has no budget concept — but worth noting that the two callers get different `max_k` behavior. The edge case `_auto_k(1)` returns `min(max(2, ...), 1) = 1`, which is correct since the final `min(k, n_beliefs)` caps it. Good.

### reasons_lib/cluster.py:list_clusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean implementation that mirrors `cluster_beliefs` structure. The small-set early return (≤3 beliefs → single cluster) is reasonable. Clusters are sorted by descending size, which is a good default for display. Uses `int(labels[i])` to avoid numpy int keys in the output dict — correct. Returns all beliefs unlike `cluster_beliefs` which samples — appropriate for a listing command.

### reasons_lib/api.py:list_clusters
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function uses `_with_network(db_path)` without `write=True`, which is correct since it only reads. However, the empty-beliefs early return constructs a result with `"clusters": []` and `"n_clusters": 0`, while a non-empty path delegates to `_list_clusters` which returns `"n_clusters": k` (the actual KMeans k, not `len(clusters)`). This is consistent since the empty case genuinely has 0 clusters. One concern: the `status` parameter accepts any string — passing `status=""` would skip the filter (matching nothing since truth_value is always "IN" or "OUT"), but passing `status=None` would also skip via `if status and ...`. The CLI constrains to "IN"/"OUT", so this is only a concern for programmatic API users. Minor: the lazy import `from .cluster import list_clusters as _list_clusters, DEFAULT_MODEL` is fine for optional-dependency isolation.

### reasons_lib/cli.py:cmd_cluster_list
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The CLI handler works but has no tests — the test file only covers `api.list_clusters` and `cluster.list_clusters`, not the CLI formatting logic. The text format output uses `args.status` to pick the marker (`+` for IN, `-` for OUT), which is a nice touch. However, the text format truncates belief text to 100 chars (`b['text'][:100]`) without any truncation indicator (no `...`), which could confuse users whose beliefs are cut off mid-word. The JSON output uses `json.dumps(result, indent=2)` which is fine. The markdown format doesn't print the model or cluster count summary that the text format does — minor inconsistency. The `--format` argument uses `args.format` directly (line 745), which works because argparse stores it as `args.format` — this is fine since `format` isn't a reserved argparse attribute.

### reasons_lib/cli.py (argparse registration)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The subparser is registered correctly between `deduplicate` and `list-gated`, maintaining logical ordering. The command dispatch map at line 1801 correctly maps `"cluster-list"` to `cmd_cluster_list`. All argument types are appropriate (`int` for `--n-clusters` and `--seed`, `choices` for `--status` and `--format`).

### Self-Review
**Limitations:** Could not verify whether `_parse_visible_to` handles edge cases (empty string, whitespace). Did not see the full `cmd_deduplicate` to compare error handling patterns. No CLI integration tests exist in the diff to verify end-to-end argument parsing. The `MagicMock` import addition wasn't verified against all test methods to confirm it's unused.

### Feature Requests
- Include the full function body of neighboring/related functions when a refactor extracts shared logic (e.g., show old `cluster_beliefs` inline alongside new `_auto_k`)
- Flag unused imports added in diffs (the `MagicMock` import appears unused)
- Include CLI integration test files if they exist, to verify command dispatch coverage

## gemini [PASS]

### reasons_lib/api.py:list_clusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function correctly filters beliefs by status and visibility before delegating to the clustering logic. It handles the empty network case gracefully.

### reasons_lib/cli.py:cmd_cluster_list
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly integrates with the API and provides multiple output formats (text, JSON, Markdown). It uses existing CLI helpers like `_require_sqlite` and `_parse_visible_to` consistently.

### reasons_lib/cluster.py:_auto_k
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Successfully refactors the cluster count logic into a reusable helper with sensible defaults and bounds. Used by both the existing `cluster_beliefs` and the new `list_clusters`.

### reasons_lib/cluster.py:list_clusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Implementation is sound, correctly handling small belief sets and sorting clusters by size for better readability. Correctly enforces dependency checks.

### tests/test_api.py:TestListClusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Provides good coverage for filtering by status, handling empty networks, and passing parameters like seed and cluster count.

### tests/test_cluster.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive tests for the new clustering logic, including edge cases like small sets, reproducibility with seeds, and the `_auto_k` helper logic.

### Self-Review
**Limitations:** I was initially unable to find the definitions of `_require_sqlite` and `_parse_visible_to` due to grep patterns, but eventually located them by reading the file context; they are existing helpers used correctly in the new code.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#129
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### `reasons_lib/cluster.py:_auto_k`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The extraction of `_auto_k` is a good refactor, but it **changes behavior** for `cluster_beliefs`. Previously, the auto-k formula was `k = min(k, budget // 3, 20)` — budget constrained the cluster count. Now `cluster_beliefs` calls `_auto_k(len(beliefs), n_clusters, max_k=min(budget // 3, 20))` which computes the same thing, so this is actually correct. However, `list_clusters` calls `_auto_k(len(beliefs), n_clusters)` with default `max_k=20` — no budget constraint, which is intentional since `list_clusters` has no budget concept. Correct on closer inspection.

### `reasons_lib/cluster.py:list_clusters`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean implementation. Returns all beliefs grouped by cluster, sorted by cluster size descending. The early-exit for `<= 3` beliefs returning a single cluster is consistent with the small-set handling pattern. Uses `int(labels[i])` to convert numpy int to Python int for JSON serialization — good. The `n_clusters` field in the return reflects the `k` passed to KMeans, which matches the actual cluster count.

### `reasons_lib/api.py:list_clusters`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function uses `_with_network(db_path)` without `write=True`, which is correct (read-only operation). The belief filtering loop matches existing patterns. However, the `status` parameter defaults to `"IN"` but also accepts empty string — `if status and n.truth_value != status` means passing `status=""` would include **all** beliefs regardless of truth value, but the CLI only allows `"IN"` or `"OUT"`. This implicit behavior isn't documented or tested, and could surprise callers using the API directly. Minor: the `embedding_model` parameter isn't passed to `_list_clusters` when it equals `DEFAULT_MODEL` by coincidence — it uses `embedding_model or DEFAULT_MODEL`, so `None` resolves correctly.

### `reasons_lib/cli.py:cmd_cluster_list`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Three output formats (json, markdown, text) are well-implemented. The text format truncates belief text at 100 chars which is practical. The marker logic (`"+"` for IN, `"-"` for OUT) correctly reflects the filtered status. The `json.dumps(result, indent=2)` works because `list_clusters` in cluster.py already converts numpy ints to Python ints. No CLI-level tests exist (e.g., testing output formatting or argument parsing), but the API layer is tested. This is consistent with how `cmd_deduplicate` is also untested at the CLI level.

### `reasons_lib/cli.py` (argparse setup + dispatch)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The subparser is properly registered with appropriate argument types and defaults. The command is wired into the dispatch dict. The `--format` argument correctly uses `choices=["text", "json", "markdown"]`. The `--visible-to` uses the existing `_parse_visible_to(args)` helper. One nit: `--status` doesn't allow filtering for all beliefs (both IN and OUT together), but this is a reasonable constraint for v1.

### `tests/test_api.py:TestListClusters`
VERDICT: CONCERN
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: PARTIAL
INTEGRATION: N/A
REASONING: Tests mock `reasons_lib.cluster.list_clusters` to avoid needing ML dependencies, which is a sound pattern. Coverage for status filtering (IN/OUT), empty network, seed passthrough, and n_clusters passthrough are all good. However: (1) No test for `visible_to` filtering even though it's a parameter. (2) No test for `embedding_model` passthrough. (3) The `test_passes_seed` and `test_passes_n_clusters` tests mock at `reasons_lib.cluster.list_clusters` but `api.py` imports as `from .cluster import list_clusters as _list_clusters` — let me verify the mock target is correct... The mock patches `reasons_lib.cluster.list_clusters` which is the module-level name, and the import `from .cluster import list_clusters as _list_clusters` binds at import time, so **the mock target should be `reasons_lib.api._list_clusters`** or alternatively the import needs to happen at call time. Wait — looking again, the import is *inside* the function body (`from .cluster import list_clusters as _list_clusters`), so it re-imports each call. Patching `reasons_lib.cluster.list_clusters` will work correctly because the `from .cluster import` inside the function will pick up the patched version. This is correct.

---

### `tests/test_cluster.py` (new tests)
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: COVERED
INTEGRATION: N/A
REASONING: Good coverage of `list_clusters`: all beliefs returned, n_clusters override, small set fallback, seed reproducibility. The `_auto_k` unit tests cover defaults, override, max_k constraint, and the n_beliefs clamping (override > n_beliefs returns n_beliefs). The test `test_auto_k_with_max_k` asserts `_auto_k(100, max_k=30) == 20` which validates that the default `max_k=20` still caps even when a higher max_k is passed — this is correct because the function logic is `min(k, max_k)` and `100 // 5 = 20` which is already <= 30.

---

### `reasons_lib/cluster.py:cluster_beliefs` (refactored)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The refactoring to use `_auto_k` preserves the original behavior exactly. Old: `k = len(beliefs) // 5; k = min(k, budget // 3, 20); k = max(k, 2); k = min(k, len(beliefs))`. New: `_auto_k(len(beliefs), n_clusters, max_k=min(budget // 3, 20))` which computes: `k = n_beliefs // 5; k = min(k, max_k); k = max(k, 2); return min(k, n_beliefs)`. The only semantic difference is that the old code applied `min(k, budget // 3, 20)` as a single expression while the new code uses `max_k=min(budget // 3, 20)` then applies `min(k, max_k)` — mathematically identical. Existing tests in `test_cluster.py` cover `cluster_beliefs`.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Did not check whether the existing `cluster_beliefs` tests still pass after the `_auto_k` refactor (verified logic equivalence manually instead). Could not verify the `_parse_visible_to` helper behavior since I didn't read it, but it's an existing function used by other commands.

### Feature Requests
- Include existing test files in full when they're modified, so I can verify existing test assertions still hold against refactored code
- Show callers/importers of modified functions to verify nothing breaks upstream
- Auto-run the test suite as part of review and include results

## gemini [PASS]

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The `list_clusters` function is correctly implemented, adhering to the project's API patterns (returning dicts, using `_with_network` context manager, and respecting `_is_visible` access controls). It handles the empty belief set case gracefully. The use of a local import for `cluster.list_clusters` is a sound choice to keep heavy dependencies (sentence-transformers) optional until needed.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** UNTESTED
**Integration:** WIRED

The `cluster-list` command is properly integrated into the CLI. It supports multiple output formats (text, JSON, markdown) and correctly passes arguments (status, n_clusters, seed, embedding_model, visible_to) to the API. The formatting logic for the text and markdown outputs is clear and includes useful details like cluster size and total belief count. While there are no specific CLI-level tests in the diff, the command relies on the well-tested API and cluster modules.

### reasons_lib/cluster.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The new `list_clusters` function and the `_auto_k` helper are well-designed. `_auto_k` centralizes the logic for determining the number of clusters, which was previously duplicated or slightly different. The implementation handles small belief sets (<= 3) specially to avoid unnecessary clustering overhead. The use of `int(labels[i])` for group keys ensures JSON serializability. The refactoring of `cluster_beliefs` to use `_auto_k` improves maintainability without changing behavior.

### tests/test_api.py:TestListClusters
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

These tests provide good coverage for the `api.list_clusters` function, specifically verifying filtering logic (status) and parameter passing to the underlying cluster library using mocks.

### tests/test_cluster.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive tests for the core clustering logic, including reproducibility with seeds, automatic k calculation, and handling of small belief sets. The use of `@skip_no_cluster` correctly handles environments without the optional dependencies.

### Self-Review
**Limitations:** I was able to verify the existence of internal helper functions like `_is_visible` and `_with_network` via `grep_search`. I confirmed that `_require_cluster_deps` is called in both `cluster_beliefs` and the new `list_clusters`. The diff includes test files which confirm high coverage of the new logic. No significant context was missing.

PR #130, opened 2026-05-10
Add --semantic mode to deduplicate

benthomasson:
## Summary

- Adds `--semantic` flag to `deduplicate` that uses embedding cosine similarity (via `ClusterCache` from `cluster.py`) instead of Jaccard similarity on ID tokens
- Catches semantically similar beliefs with different names (e.g., "input validation at boundaries" vs "boundary input checking")
- Adds `--embedding-model` flag to override the default sentence-transformers model
- Existing Jaccard mode unchanged — `--semantic` is opt-in

## Test plan

- [x] 4 new tests in `TestDeduplicateSemantic`: finds similar text, skips dissimilar, auto-retracts, Jaccard mode unchanged
- [x] Full suite: 871 passed, 131 skipped
- [x] Manual test against expert repo: 30 semantic clusters (65 beliefs) at threshold 0.85

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2183,9 +2183,8 @@ def deduplicate(
     Returns: {"clusters": list[dict], "retracted": list[str]}
     """
     if semantic:
-        from .cluster import ClusterCache, DEFAULT_MODEL, _require_cluster_deps
+        from .cluster import ClusterCache, DEFAULT_MODEL
         import numpy as np
-        _require_cluster_deps()
     else:
         from .derive import _tokenize_id, _jaccard
 
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2192,6 +2192,9 @@ def deduplicate(
         in_nodes = [(nid, n) for nid, n in sorted(net.nodes.items())
                     if n.truth_value == "IN"]
 
+        if not in_nodes:
+            return {"clusters": [], "retracted": []}
+
         if semantic:
             beliefs = {nid: n.text for nid, n in in_nodes}
             cache = ClusterCache(embedding_model or DEFAULT_MODEL)
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2164,30 +2164,46 @@ def list_clusters(
 def deduplicate(
     threshold: float = 0.5,
     auto: bool = False,
+    semantic: bool = False,
+    embedding_model: str | None = None,
     db_path: str = DEFAULT_DB,
 ) -> dict:
-    """Find clusters of IN beliefs with similar IDs (likely duplicates).
+    """Find clusters of IN beliefs with similar IDs or text (likely duplicates).
 
-    Uses Jaccard similarity on ID tokens to detect beliefs that say the
-    same thing under slightly different IDs.
+    Uses Jaccard similarity on ID tokens by default, or embedding cosine
+    similarity when semantic=True.
 
     Args:
-        threshold: Minimum Jaccard similarity to consider a pair (default: 0.5)
+        threshold: Minimum similarity to consider a pair (default: 0.5)
         auto: If True, retract all but the most-connected belief in each cluster
+        semantic: If True, use embedding cosine similarity instead of ID tokens
+        embedding_model: Sentence-transformers model (semantic mode only)
         db_path: Path to database
 
     Returns: {"clusters": list[dict], "retracted": list[str]}
     """
-    from .derive import _tokenize_id, _jaccard
+    if semantic:
+        from .cluster import ClusterCache, DEFAULT_MODEL, _require_cluster_deps
+        import numpy as np
+        _require_cluster_deps()
+    else:
+        from .derive import _tokenize_id, _jaccard
 
     with _with_network(db_path, write=auto) as net:
         in_nodes = [(nid, n) for nid, n in sorted(net.nodes.items())
                     if n.truth_value == "IN"]
 
-        # Build token sets once
-        tokens = {nid: _tokenize_id(nid) for nid, _ in in_nodes}
+        if semantic:
+            beliefs = {nid: n.text for nid, n in in_nodes}
+            cache = ClusterCache(embedding_model or DEFAULT_MODEL)
+            ids, embeddings = cache.embed(beliefs)
+            norms = np.linalg.norm(embeddings, axis=1, keepdims=True)
+            norms = np.where(norms == 0, 1, norms)
+            normed = embeddings / norms
+            sim_matrix = normed @ normed.T
+        else:
+            tokens = {nid: _tokenize_id(nid) for nid, _ in in_nodes}
 
-        # Union-find to group similar beliefs
         parent = {nid: nid for nid, _ in in_nodes}
 
         def find(x):
@@ -2201,10 +2217,16 @@ def deduplicate(
             if ra != rb:
                 parent[ra] = rb
 
-        for i, (nid_a, _) in enumerate(in_nodes):
-            for nid_b, _ in in_nodes[i + 1:]:
-                if _jaccard(tokens[nid_a], tokens[nid_b]) >= threshold:
-                    union(nid_a, nid_b)
+        if semantic:
+            for i in range(len(ids)):
+                for j in range(i + 1, len(ids)):
+                    if sim_matrix[i, j] >= threshold:
+                        union(ids[i], ids[j])
+        else:
+            for i, (nid_a, _) in enumerate(in_nodes):
+                for nid_b, _ in in_nodes[i + 1:]:
+                    if _jaccard(tokens[nid_a], tokens[nid_b]) >= threshold:
+                        union(nid_a, nid_b)
 
         # Collect clusters (only groups of 2+)
         from collections import defaultdict
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -797,7 +797,7 @@ def cmd_deduplicate(args):
         threshold=args.threshold,
         auto=args.auto,
         semantic=args.semantic,
-        embedding_model=getattr(args, "embedding_model", None),
+        embedding_model=args.embedding_model,
         db_path=args.db,
     )
 
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -796,6 +796,8 @@ def cmd_deduplicate(args):
     result = api.deduplicate(
         threshold=args.threshold,
         auto=args.auto,
+        semantic=args.semantic,
+        embedding_model=getattr(args, "embedding_model", None),
         db_path=args.db,
     )
 
@@ -1660,7 +1662,11 @@ def main():
     # deduplicate
     p = sub.add_parser("deduplicate", help="Find and optionally retract duplicate IN beliefs")
     p.add_argument("--threshold", type=float, default=0.5,
-                   help="Jaccard similarity threshold for ID tokens (default: 0.5)")
+                   help="Similarity threshold (Jaccard for ID mode, cosine for --semantic; default: 0.5)")
+    p.add_argument("--semantic", action="store_true",
+                   help="Use embedding similarity instead of ID token similarity")
+    p.add_argument("--embedding-model", default=None,
+                   help="Sentence-transformers model (default: all-MiniLM-L6-v2)")
     p.add_argument("--auto", action="store_true",
                    help="Automatically retract duplicates (keeps one per cluster)")
     p.add_argument("-o", "--output", default="proposed-dedup.md",
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -673,9 +673,10 @@ class TestDeduplicateSemantic:
     def test_semantic_auto_retracts(self, db_with_similar):
         result = api.deduplicate(threshold=0.5, semantic=True, auto=True,
                                   db_path=db_with_similar)
-        assert len(result["retracted"]) >= 1
         retracted_set = set(result["retracted"])
-        assert not ({"input-validation-at-boundaries", "boundary-input-checking"} <= retracted_set)
+        similar_pair = {"input-validation-at-boundaries", "boundary-input-checking"}
+        assert len(retracted_set & similar_pair) == 1
+        assert "database-query-performance" not in retracted_set
 
     def test_jaccard_mode_unchanged(self, db_with_similar):
         result = api.deduplicate(threshold=0.5, semantic=False, db_path=db_with_similar)
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -678,6 +678,13 @@ class TestDeduplicateSemantic:
         assert len(retracted_set & similar_pair) == 1
         assert "database-query-performance" not in retracted_set
 
+    def test_semantic_empty_network(self, tmp_path):
+        db = str(tmp_path / "empty.db")
+        api.init_db(db_path=db)
+        result = api.deduplicate(threshold=0.5, semantic=True, db_path=db)
+        assert result["clusters"] == []
+        assert result["retracted"] == []
+
     def test_jaccard_mode_unchanged(self, db_with_similar):
         result = api.deduplicate(threshold=0.5, semantic=False, db_path=db_with_similar)
         assert result["retracted"] == []
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -626,3 +626,57 @@ class TestListClusters:
         with patch("reasons_lib.cluster.list_clusters", return_value=mock_result) as mock_lc:
             api.list_clusters(n_clusters=3, db_path=db_with_beliefs)
             assert mock_lc.call_args[1]["n_clusters"] == 3
+
+
+try:
+    from reasons_lib.cluster import HAS_CLUSTER_DEPS
+except ImportError:
+    HAS_CLUSTER_DEPS = False
+
+skip_no_cluster = pytest.mark.skipif(
+    not HAS_CLUSTER_DEPS,
+    reason="sentence-transformers and scikit-learn not installed"
+)
+
+
+@skip_no_cluster
+class TestDeduplicateSemantic:
+
+    @pytest.fixture
+    def db_with_similar(self, tmp_path):
+        db = str(tmp_path / "sim.db")
+        api.init_db(db_path=db)
+        api.add_node("input-validation-at-boundaries",
+                      "The system validates all inputs at system boundaries",
+                      db_path=db)
+        api.add_node("boundary-input-checking",
+                      "Input validation occurs at system edges and boundaries",
+                      db_path=db)
+        api.add_node("database-query-performance",
+                      "Database queries are optimized for read-heavy workloads",
+                      db_path=db)
+        return db
+
+    def test_semantic_finds_similar_text(self, db_with_similar):
+        result = api.deduplicate(threshold=0.5, semantic=True, db_path=db_with_similar)
+        assert len(result["clusters"]) >= 1
+        cluster_ids = {b["id"] for b in result["clusters"][0]["beliefs"]}
+        assert "input-validation-at-boundaries" in cluster_ids
+        assert "boundary-input-checking" in cluster_ids
+
+    def test_semantic_skips_dissimilar(self, db_with_similar):
+        result = api.deduplicate(threshold=0.8, semantic=True, db_path=db_with_similar)
+        for cluster in result["clusters"]:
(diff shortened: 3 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#130
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### reasons_lib/api.py:deduplicate
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

1. **`_require_cluster_deps()` called twice**: `ClusterCache.__init__` already calls `_require_cluster_deps()` internally (cluster.py:37). The explicit call at api.py line ~2190 is redundant but harmless — it just means the error is raised before constructing the cache object rather than inside it.

2. **Empty `in_nodes` with semantic mode**: If there are 0 or 1 IN nodes, the semantic path constructs a `ClusterCache` and calls `cache.embed({})` or embeds a single belief, then builds `sim_matrix` — this works but does unnecessary embedding work. The Jaccard path naturally produces no clusters. Minor efficiency concern only.

3. **`numpy` imported at function scope under conditional**: `import numpy as np` inside `if semantic:` is fine for lazy loading, but the `np` name is then used later in the same function scope inside another `if semantic:` block. This works in Python but relies on the two `if semantic:` blocks always being entered together, which they are since `semantic` doesn't change. Acceptable but somewhat fragile — if someone refactored the blocks independently they'd get a `NameError`.

4. **Cosine similarity is correct**: The manual normalization + dot product is a valid cosine similarity computation. The `np.where(norms == 0, 1, norms)` guard against zero-norm vectors is correct.

5. **O(n^2) pairwise comparison**: Both paths use O(n^2) comparison. For semantic mode this is fine since embedding is the bottleneck. No regression.

### reasons_lib/cli.py:cmd_deduplicate
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

1. **`getattr(args, "embedding_model", None)` is unnecessary**: argparse converts `--embedding-model` to `args.embedding_model` and it has `default=None`, so `args.embedding_model` always exists. Plain `args.embedding_model` would be clearer. The `getattr` suggests uncertainty about whether the attribute exists, which could mask real bugs.

2. **`--accept` path ignores `--semantic`**: When `--accept` is used (applying a reviewed dedup plan), the function returns early without consulting `args.semantic`. This is probably correct — the plan file is mode-agnostic — but it means `reasons deduplicate --semantic --accept plan.md` silently ignores `--semantic`. Could confuse users who think they need both flags.

3. **No CLI test for the new flags**: The diff only shows API-level tests. No test verifies that `--semantic` is correctly parsed and forwarded from the CLI.

### reasons_lib/cli.py (argparse changes)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The argument definitions are clean. `--semantic` as `store_true` and `--embedding-model` with `default=None` follow established patterns in this CLI. Help text is accurate.

### Self-Review
**Limitations:** Could not verify whether `ClusterCache.embed()` handles an empty dict correctly (0 IN nodes edge case). Could not see the full test suite to check if CLI-level tests exist elsewhere for deduplicate. Could not verify whether `_rewrite_dependents` behaves correctly when the "kept" node comes from semantic similarity vs ID similarity (the merge logic is mode-agnostic, which seems right, but couldn't verify the full function).

### Feature Requests
- Include the full body of related helper functions called from modified code (e.g., `_rewrite_dependents`, `_with_network`) to verify integration correctness
- Show existing CLI-level tests for the same subcommand to assess CLI test coverage gaps
- Flag when test assertions use negative logic (`assert not X`) that could silently pass on bugs

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#130
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### reasons_lib/api.py:deduplicate
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The semantic path works correctly end-to-end, but there are two issues: (1) **Conditional imports at function top are evaluated eagerly even though they appear inside the function** — this is fine for correctness but the `import numpy as np` inside the `if semantic:` branch means numpy is only imported when needed, which is good. However, the `else` branch imports `_tokenize_id` and `_jaccard` even when `semantic=True` is used — this is not a bug since the `if semantic:` guard prevents execution, but structurally the conditional import at the top is a bit unusual. (2) **More substantively**: the `ids` list from `cache.embed()` may not be in the same order as `in_nodes`. The `embed()` method returns `sorted(beliefs.keys())` order. The union-find `parent` dict is keyed by `nid` from `in_nodes`, but the similarity loop indexes by `ids[i]` and `ids[j]`. Since `parent` contains all in_nodes' IDs, and `ids` is a subset of those same IDs (same set, different order), the `union(ids[i], ids[j])` calls will work correctly — the parent dict has all needed keys. This is correct but the implicit coupling between `ids` ordering from `embed()` and the `parent` dict is fragile. A comment would help.

### reasons_lib/cli.py:cmd_deduplicate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

CLI args `--semantic` and `--embedding-model` are properly defined and passed through to `api.deduplicate()`. The help text for `--threshold` is updated to mention both modes. The `--accept` path correctly does not need these new params since plan application is mode-agnostic.

### reasons_lib/cli.py:main (argparse)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Parser definitions are clean. `--embedding-model` uses `default=None` which correctly triggers the fallback to `DEFAULT_MODEL` in the API layer. The `store_true` for `--semantic` is the right pattern.

### tests/test_api.py:TestDeduplicateSemantic
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Good test coverage of the happy path, but `test_semantic_auto_retracts` has a suspicious assertion: `assert not ({"input-validation-at-boundaries", "boundary-input-checking"} <= retracted_set)` asserts that *not both* were retracted (correct — auto mode should keep one). But it doesn't assert that *exactly one* was retracted from that pair. The test only checks `len(result["retracted"]) >= 1`, which could pass even if the wrong node was retracted (e.g., the unrelated "database-query-performance" node). A stronger assertion would verify that exactly one of the two similar beliefs was retracted and the dissimilar one was not. Also missing: test for `embedding_model` parameter passthrough, and test behavior when `semantic=True` but cluster deps are not installed (the `skipif` handles this at the test level, but there's no test that the API raises a clear error).

### tests/test_cli.py:TestDeduplicateSemantic
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Two CLI-level tests covering the semantic flag and threshold interaction. `test_semantic_finds_similar` verifies the CLI outputs "Cluster" text, and `test_semantic_with_threshold` verifies high threshold produces "No duplicate". These are good smoke tests. Missing: test with `--embedding-model` flag, test combining `--semantic --auto`.

### Self-Review
**Limitations:** Could not see the full `cluster.py` `embed()` implementation to verify ordering guarantees. Could not verify whether `sim_matrix` diagonal handling matters (diagonal is 1.0, so each node is similar to itself, but since `i < j` loop skips diagonal, this is fine). Did not verify the `_with_network` context manager's write behavior for the `auto=True` semantic path.

### Feature Requests
- Include the full implementation of called functions (e.g., `ClusterCache.embed()`) when they're used in new code paths, so reviewers can verify contract alignment without separate lookups
- Show the complete function body around diff hunks, not just the changed lines — the retraction logic after the clustering section is important context

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#130
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/api.py:deduplicate
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The semantic path manually recomputes cosine similarity (normalize + matmul) despite `ClusterCache` already existing in `cluster.py` — this duplicates logic and is a maintainability concern. More importantly, there's an ordering assumption: `cache.embed(beliefs)` returns `ids` in a particular order, and the code assumes `ids[i]` corresponds to `in_nodes` entries. This is safe *today* because `embed()` preserves dict insertion order, but it's fragile. The conditional import of `numpy` at function scope (only when `semantic=True`) is fine but means a typo or numpy absence would only surface at runtime in semantic mode — no static analysis catches it. The division-by-zero guard on norms (`np.where(norms == 0, 1, norms)`) is correct. The union-find is duplicated between the two branches rather than shared — the `find`/`union` definitions and cluster collection are identical, only the similarity loop differs. This could be a single loop with a pluggable similarity function.

### reasons_lib/api.py — conditional imports
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `if semantic:` / `else:` import split at lines 2185-2189 means both code paths import their dependencies lazily, which is good for optional deps. However, placing `import numpy as np` inside a conditional means type checkers and linters won't catch usage errors. Minor — acceptable for optional-dependency patterns.

### reasons_lib/cli.py:cmd_deduplicate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`args.semantic` and `args.embedding_model` are correctly threaded through to `api.deduplicate()`. The `--accept` early-return path silently ignores `--semantic`, which is functionally correct (accept applies a pre-reviewed plan regardless of how it was generated) but could confuse users. A one-line warning when `--semantic` is combined with `--accept` would be a nice UX improvement but isn't a blocker.

### reasons_lib/cli.py — argparse definitions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`--semantic` (store_true) and `--embedding-model` (string, default None) are properly defined. The help text for `--threshold` was updated to mention both modes. argparse auto-converts `--embedding-model` to `args.embedding_model`. Clean.

### tests/test_api.py:TestDeduplicateSemantic
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Good coverage of the happy path — finds similar text, skips dissimilar, auto-retracts correctly, and verifies Jaccard mode is unaffected. The `test_semantic_auto_retracts` correctly asserts exactly one of the similar pair is retracted. However, missing edge cases: (1) no IN beliefs, (2) all beliefs identical (threshold 1.0), (3) single belief (no pairs to compare), (4) what happens when `semantic=True` but cluster deps are missing — the test skips entirely rather than verifying a clean error message. The `skip_no_cluster` guard is duplicated between `test_api.py` and `test_cli.py` — could be a shared fixture.

### tests/test_cli.py:TestDeduplicateSemantic
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

CLI-level tests verify `--semantic` produces cluster output and `--threshold 0.95` correctly yields no duplicates for dissimilar beliefs. Good end-to-end coverage. Missing: test for `--semantic --auto` through CLI, and test for `--semantic --embedding-model <model>` flag passthrough.

### Self-Review
**Limitations:** I verified the full function body and related modules via subagent, so I have good coverage of the integration. I did not verify whether `ClusterCache.embed()` guarantees insertion-order preservation of the returned `ids` list — the ordering assumption in the semantic similarity loop is the main correctness risk I couldn't fully verify from the diff alone. I also couldn't run the test suite to confirm the tests pass with actual embeddings.

### Feature Requests
- Show the full function body for every modified function in observations, not just diff hunks — the auto-retract logic at the bottom of `deduplicate()` was critical context that wasn't in the diff
- Flag when conditional imports mean a code path is untestable by static analysis
- Detect duplicated code patterns between branches of an if/else (the union-find logic is identical in both paths)

## gemini [CONCERN]

### reasons_lib/api.py:deduplicate
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The function will crash with a `ValueError` (axis 1 out of bounds) if the network contains no `IN` nodes and `semantic=True` is used. This happens because `ClusterCache.embed({})` returns an empty 1D numpy array, and `np.linalg.norm(embeddings, axis=1)` is called on it. A guard like `if not in_nodes: return {"clusters": [], "retracted": []}` should be added at the start of the function. Additionally, the `import numpy as np` inside the `if semantic:` block might mask the descriptive `ImportError` from `_require_cluster_deps()` if `numpy` itself is missing, though this is a minor issue.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The CLI correctly exposes the new flags (`--semantic`, `--embedding-model`) and routes them to the API. Help text is updated and consistent with other clustering commands.

### tests/test_api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new tests correctly verify semantic similarity clustering, threshold behavior, and auto-retraction. The use of `skip_no_cluster` ensures tests don't fail in environments without optional ML dependencies.

### tests/test_cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

CLI tests successfully verify the end-to-end flow of the semantic deduplication feature, including threshold filtering and output formatting.

### Self-Review
**Limitations:** I could not verify the exact output of `sentence_transformers.SentenceTransformer.encode` for an empty list in this specific environment, but based on standard `numpy` behavior, the downstream `np.linalg.norm` call on an empty array with `axis=1` is a known crash point.

### Feature Requests
- Automatically provide a reproduction script or "mental-run" capability for edge cases like empty networks.
- Flag potential optional dependency issues more explicitly when imports are nested within conditional blocks.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#130
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### `reasons_lib/api.py:deduplicate`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Logic is sound. Conditional imports keep ML deps optional. Zero-norm guard (`np.where(norms == 0, 1, norms)`) correctly prevents division by zero — zero-vector embeddings produce 0 cosine similarity, so they won't spuriously cluster. Empty-network early return is correct. Union-find is shared between modes and `ids` from `cache.embed` maps to the same key space as `parent`, so the union calls are safe. O(n²) pairwise comparison matches the existing Jaccard path — acceptable for a belief network's typical scale.

### `reasons_lib/cli.py` (parser + `cmd_deduplicate`)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Straightforward wiring. `--semantic` is `store_true`, `--embedding-model` defaults to `None` (deferred to `DEFAULT_MODEL` in api.py). Argparse hyphen-to-underscore conversion for `embedding_model` is handled correctly. Help text accurately describes both modes.

### `tests/test_api.py:TestDeduplicateSemantic`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Five tests cover the key scenarios: similar-text clustering, dissimilar exclusion at high threshold, auto-retraction (asserts exactly one of a pair retracted), empty network, and Jaccard-mode regression. The `skip_no_cluster` gate properly handles environments without ML deps. One minor note: `test_jaccard_mode_unchanged` only asserts `retracted == []` which is trivially true without `auto=True` — checking `result["clusters"]` would strengthen it, but it's not wrong as-is since it's a smoke test that the non-semantic path doesn't break.

### `tests/test_cli.py:TestDeduplicateSemantic`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Two end-to-end CLI tests verify the `--semantic` flag works through the full stack and that `--threshold` with a high value produces no clusters. Correctly gated behind `HAS_CLUSTER_DEPS`.

### `reasons_lib/api.py` (shared default threshold)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The same default threshold (0.5) is used for both Jaccard and cosine similarity, but these metrics have different distributions. Jaccard 0.5 requires substantial token overlap; cosine 0.5 on sentence embeddings is a much lower bar (many unrelated sentences can hit 0.5). This could lead to over-clustering in semantic mode with the default threshold. Consider a different default for semantic mode (e.g., 0.7–0.8), or at minimum document the difference in the help text. Not blocking since users can override it, and the help text does mention both modes.

### Self-Review
**Limitations:** Could not see the implementation of `ClusterCache.embed()` to verify the order and completeness of returned `ids` relative to input `beliefs` dict. Could not see the full `deduplicate` function (only diff hunks + context) to verify the retraction logic after the union-find is unchanged. Could not verify that `_tokenize_id` splits on hyphens as assumed when reasoning about `test_jaccard_mode_unchanged`.

### Feature Requests
- Include full function bodies for modified functions, not just diff context — the retraction logic after the union-find is cut off at the diff boundary
- Show the implementation of called functions (e.g., `ClusterCache.embed`) when they're central to the correctness argument
- Flag when the same skip/import pattern is duplicated across test files (the `HAS_CLUSTER_DEPS` try/except appears in both test files — could be a shared fixture)

## gemini [PASS]

### reasons_lib/api.py:deduplicate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation of semantic deduplication is correct and follows the established pattern of the codebase. It uses the `ClusterCache` to handle embeddings efficiently and performs vector normalization to compute cosine similarity via matrix multiplication. The union-find logic ensures that groups of similar beliefs are correctly clustered. The integration of `_rewrite_dependents` before retraction ensures that the TMS integrity is maintained by preserving reasoning chains. The default behavior (Jaccard similarity on IDs) remains unchanged, ensuring backward compatibility.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The CLI changes correctly expose the new semantic deduplication features to the user. The `--semantic` flag and `--embedding-model` option are properly threaded through to the underlying API call. The help text has been updated to clarify that the threshold applies to both modes (Jaccard and cosine).

### tests/test_api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New tests cover semantic similarity detection, threshold sensitivity, and automatic retraction. The tests also verify that the default Jaccard mode remains unaffected by semantic similarity when IDs are distinct. The use of `@skip_no_cluster` correctly handles environments without the optional clustering dependencies.

### tests/test_cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

CLI tests verify that the `--semantic` flag is functional and produces expected output when duplicates are found. They also test the threshold override.

### Self-Review
**Limitations:** I could not see the implementation of `_rewrite_dependents`, but its usage in `api.py` is consistent with other deduplication logic in the project. The `ClusterCache.embed` method was verified in `cluster.py` to ensure it returns ordered IDs and embeddings, which is critical for the similarity matrix indexing.

### Feature Requests
- Include the definition of `_rewrite_dependents` or the `Network` class when reviewing graph-mutating functions in `api.py`.

PR #131, opened 2026-05-10
Add --semantic mode to detect-contradictions

benthomasson:
## Summary

- Adds `--semantic` flag to `contradictions` that clusters beliefs by embedding similarity before sending each cluster to the LLM
- Topically related beliefs are analyzed together instead of in random batches, increasing the chance of catching contradictions
- Adds `--embedding-model` flag to override the default sentence-transformers model
- Existing random-batch mode unchanged — `--semantic` is opt-in

## Test plan

- [x] 3 new tests in `TestDetectContradictionsSemantic`: cluster grouping, empty network, single-belief skip
- [x] Full suite: 877 passed, 131 skipped
- [x] Existing contradiction tests unaffected

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2056,6 +2056,7 @@ def detect_contradictions(
     auto_apply: bool = False,
     semantic: bool = False,
     embedding_model: str | None = None,
+    output_path: str | None = None,
     db_path: str = DEFAULT_DB,
 ) -> dict:
     """Detect contradictions between IN beliefs via LLM analysis.
@@ -2063,6 +2064,9 @@ def detect_contradictions(
     When semantic=True, beliefs are clustered by embedding similarity
     before sending to the LLM, so topically related beliefs are
     analyzed together.
+
+    When output_path is set, results are written incrementally after
+    each batch/cluster completes.
 
     Returns: {"contradictions": [...], "checked": int, "found": int,
               "applied": int, "total_in": int}
@@ -2092,10 +2096,12 @@ def detect_contradictions(
     if semantic:
         contradictions = _detect_semantic(nodes, belief_ids=check_ids,
                                           model=model, timeout=timeout,
-                                          embedding_model=embedding_model)
+                                          embedding_model=embedding_model,
+                                          output_path=output_path)
     else:
         contradictions = _detect(nodes, belief_ids=check_ids, model=model,
-                                timeout=timeout)
+                                timeout=timeout,
+                                output_path=output_path)
 
     applied = 0
     applied_details = []
@@ -2121,6 +2127,95 @@ def detect_contradictions(
         "applied_details": applied_details,
         "total_in": total_in,
     }
+
+
+def write_contradiction_plan(contradictions: list[dict], output_path: str,
+                             append: bool = False) -> str:
+    """Write (or append to) a contradiction plan file for human review.
+
+    Format is parseable by parse_contradiction_plan(). Each nogood lists
+    claims and analysis. Change [APPLY] to [SKIP] or delete entries to
+    exclude them.
+    """
+    path = Path(output_path)
+    mode = "a" if append else "w"
+    with open(path, mode) as f:
+        if not append:
+            f.write("# Contradiction Plan\n\n")
+            f.write("Review each nogood below. Delete any you want to skip,\n")
+            f.write("or change APPLY to SKIP. Then run:\n")
+            f.write(f"  reasons contradictions --accept {output_path}\n\n")
+            f.write("---\n\n")
+
+        for c in contradictions:
+            severity = c.get("severity", "")
+            f.write(f"### NOGOOD {c['id']} [APPLY]\n")
+            if severity:
+                f.write(f"- Severity: {severity}\n")
+            f.write(f"- Claims: {', '.join(c['claims'])}\n")
+            if c.get("analysis"):
+                f.write(f"- Analysis: {c['analysis']}\n")
+            f.write("\n")
+
+    return str(path)
+
+
+def parse_contradiction_plan(plan_text: str) -> list[dict]:
+    """Parse a contradiction plan file into actionable entries.
+
+    Returns list of {"id": str, "claims": list[str]} for APPLY-marked entries.
+    """
+    import re
+    entries = []
+    current_id = None
+    current_claims = []
+
+    for line in plan_text.splitlines():
+        m = re.match(r"###\s+NOGOOD\s+(\S+)\s+\[(APPLY|SKIP)\]", line)
+        if m:
+            if current_id and current_claims:
+                entries.append({"id": current_id, "claims": current_claims})
+            current_id = m.group(2) == "APPLY" and m.group(1) or None
+            current_claims = []
+            continue
+
+        if current_id and line.strip().startswith("- Claims:"):
+            claims_str = line.split(":", 1)[1].strip()
+            current_claims = [c.strip().strip("`") for c in claims_str.split(",")
+                              if c.strip()]
+
+    if current_id and current_claims:
+        entries.append({"id": current_id, "claims": current_claims})
+
+    return entries
+
+
+def apply_contradiction_plan(
+    plan: list[dict],
+    db_path: str = DEFAULT_DB,
+) -> dict:
+    """Apply a reviewed contradiction plan: record nogoods and backtrack.
+
+    Args:
+        plan: list of {"id": str, "claims": list[str]} from parse_contradiction_plan
+        db_path: Path to database
+
+    Returns: {"applied": int, "nogoods": list[dict], "errors": list[str]}
+    """
+    nogoods = []
+    errors = []
+    for entry in plan:
+        try:
+            result = add_nogood(entry["claims"], db_path=db_path)
+            nogoods.append({
+                "id": entry["id"],
+                "nogood_id": result.get("nogood_id"),
+                "changed": result.get("changed", []),
+                "backtracked_to": result.get("backtracked_to"),
+            })
+        except Exception as e:
+            errors.append(f"Failed to apply {entry['id']}: {e}")
+    return {"applied": len(nogoods), "nogoods": nogoods, "errors": errors}
 
 
 def _rewrite_dependents(net, old_id: str, new_id: str):
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2175,7 +2175,7 @@ def parse_contradiction_plan(plan_text: str) -> list[dict]:
         if m:
             if current_id and current_claims:
                 entries.append({"id": current_id, "claims": current_claims})
-            current_id = m.group(2) == "APPLY" and m.group(1) or None
+            current_id = m.group(1) if m.group(2) == "APPLY" else None
             current_claims = []
             continue
 
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2054,14 +2054,22 @@ def detect_contradictions(
     timeout: int = 300,
     sample: int | None = None,
     auto_apply: bool = False,
+    semantic: bool = False,
+    embedding_model: str | None = None,
     db_path: str = DEFAULT_DB,
 ) -> dict:
     """Detect contradictions between IN beliefs via LLM analysis.
+
+    When semantic=True, beliefs are clustered by embedding similarity
+    before sending to the LLM, so topically related beliefs are
+    analyzed together.
 
     Returns: {"contradictions": [...], "checked": int, "found": int,
               "applied": int, "total_in": int}
     """
     from .contradictions import detect_contradictions as _detect
+    if semantic:
+        from .contradictions import detect_contradictions_semantic as _detect_semantic
 
     result = export_network(db_path=db_path)
     nodes = result.get("nodes", {})
@@ -2081,8 +2089,13 @@ def detect_contradictions(
         candidates = {k: candidates[k] for k in sampled_keys}
 
     check_ids = sorted(candidates.keys())
-    contradictions = _detect(nodes, belief_ids=check_ids, model=model,
-                            timeout=timeout)
+    if semantic:
+        contradictions = _detect_semantic(nodes, belief_ids=check_ids,
+                                          model=model, timeout=timeout,
+                                          embedding_model=embedding_model)
+    else:
+        contradictions = _detect(nodes, belief_ids=check_ids, model=model,
+                                timeout=timeout)
 
     applied = 0
     applied_details = []
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -1327,16 +1327,40 @@ def cmd_review_beliefs(args):
 
 def cmd_contradictions(args):
     _require_sqlite(args, "detect-contradictions")
+
+    if args.accept:
+        accept_path = Path(args.accept)
+        if not accept_path.exists():
+            print(f"File not found: {accept_path}", file=sys.stderr)
+            sys.exit(1)
+        plan = api.parse_contradiction_plan(accept_path.read_text())
+        if not plan:
+            print("No APPLY entries found in plan file.")
+            return
+        result = api.apply_contradiction_plan(plan, db_path=args.db)
+        for err in result["errors"]:
+            print(f"  ERROR: {err}", file=sys.stderr)
+        if result["nogoods"]:
+            print(f"Applied {result['applied']} nogood(s):")
+            for n in result["nogoods"]:
+                changed = n.get("changed", [])
+                print(f"  {n['id']}: nogood={n.get('nogood_id', '?')}, "
+                      f"changed {len(changed)} node(s)")
+        else:
+            print("No nogoods to apply.")
+        return
+
     model = getattr(args, "model", None) or "claude"
-    auto_apply = args.auto_apply and not args.dry_run
+    output = args.output
     result = api.detect_contradictions(
         belief_ids=args.ids or None,
         model=model,
         timeout=args.timeout,
         sample=args.sample,
-        auto_apply=auto_apply,
+        auto_apply=args.auto_apply,
         semantic=args.semantic,
         embedding_model=args.embedding_model,
+        output_path=output if not args.auto_apply else None,
         db_path=args.db,
     )
 
@@ -1356,27 +1380,15 @@ def cmd_contradictions(args):
     print(f"\nChecked {result['checked']} of {result['total_in']} IN beliefs")
     print(f"  Found: {result['found']}  Applied: {result['applied']}")
 
-    if args.output:
-        with open(args.output, "w") as f:
-            f.write("# Contradiction Detection Findings\n\n")
-            f.write(f"Checked {result['checked']} beliefs, "
-                    f"found {result['found']} contradictions.\n\n")
-            for c in contradictions:
-                f.write(f"### NOGOOD {c['id']}\n")
-                f.write(f"- Claims: {', '.join(c['claims'])}\n")
-                if c.get("analysis"):
-                    f.write(f"- Analysis: {c['analysis']}\n")
-                if c.get("severity"):
-                    f.write(f"- Severity: {c['severity']}\n")
-                f.write("\n")
-        print(f"\nWrote findings to {args.output}")
-
-    if auto_apply and result.get("applied_details"):
+    if args.auto_apply and result.get("applied_details"):
         print(f"\nApplied {result['applied']} nogood(s):")
         for d in result["applied_details"]:
             changed = d.get("changed", [])
             print(f"  {d.get('id', '?')}: nogood={d.get('nogood_id', '?')}, "
                   f"changed {len(changed)} node(s)")
+    elif not args.auto_apply:
+        print(f"\nWrote {output} — review, then run:")
+        print(f"  reasons contradictions --accept {output}")
 
 
 def cmd_namespaces(args):
@@ -1737,16 +1749,16 @@ def main():
                    help="LLM timeout in seconds (default: 300)")
     p.add_argument("--sample", type=int, default=None,
                    help="Randomly sample N beliefs to check")
-    p.add_argument("--dry-run", action="store_true",
-                   help="Show findings without applying nogoods")
     p.add_argument("--auto-apply", action="store_true",
                    help="Auto-apply detected nogoods via dependency-directed backtracking")
     p.add_argument("--semantic", action="store_true",
                    help="Group beliefs by semantic similarity before LLM analysis")
     p.add_argument("--embedding-model", default=None,
                    help="Sentence-transformers model (default: all-MiniLM-L6-v2)")
-    p.add_argument("-o", "--output", default=None,
-                   help="Write proposals to markdown file")
+    p.add_argument("-o", "--output", default="proposed-contradictions.md",
+                   help="Output file for contradiction plan (default: proposed-contradictions.md)")
+    p.add_argument("--accept", metavar="FILE",
+                   help="Apply a reviewed contradiction plan file")
 
     # list
     p = sub.add_parser("list", help="List nodes with filters")
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -1335,6 +1335,8 @@ def cmd_contradictions(args):
         timeout=args.timeout,
         sample=args.sample,
         auto_apply=auto_apply,
+        semantic=args.semantic,
+        embedding_model=args.embedding_model,
         db_path=args.db,
     )
 
@@ -1739,6 +1741,10 @@ def main():
                    help="Show findings without applying nogoods")
     p.add_argument("--auto-apply", action="store_true",
                    help="Auto-apply detected nogoods via dependency-directed backtracking")
+    p.add_argument("--semantic", action="store_true",
+                   help="Group beliefs by semantic similarity before LLM analysis")
+    p.add_argument("--embedding-model", default=None,
+                   help="Sentence-transformers model (default: all-MiniLM-L6-v2)")
     p.add_argument("-o", "--output", default=None,
                    help="Write proposals to markdown file")
 
--- a/reasons_lib/contradictions.py
+++ b/reasons_lib/contradictions.py
@@ -110,8 +110,18 @@ def parse_contradiction_response(response, valid_ids=None):
     return results
 
 
+def _flush_results(results, output_path, header_written):
+    """Write results to plan file incrementally."""
+    if not output_path or not results:
+        return header_written
+    from .api import write_contradiction_plan
+    write_contradiction_plan(results, output_path, append=header_written)
+    return True
+
+
 def detect_contradictions(nodes, belief_ids=None, model="claude", timeout=300,
-                          batch_size=CONTRADICTION_BATCH_SIZE):
+                          batch_size=CONTRADICTION_BATCH_SIZE,
+                          output_path=None):
     """Detect contradictions between IN beliefs via LLM.
 
     Args:
@@ -121,6 +131,7 @@ def detect_contradictions(nodes, belief_ids=None, model="claude", timeout=300,
         model: LLM model to use.
         timeout: LLM timeout in seconds.
         batch_size: Number of beliefs per LLM call.
+        output_path: If set, write results incrementally to this file.
 
     Returns: list of contradiction dicts.
     """
@@ -145,6 +156,7 @@ def detect_contradictions(nodes, belief_ids=None, model="claude", timeout=300,
     valid_ids = set(belief_ids)
     all_results = []
     total_batches = (len(belief_ids) + batch_size - 1) // batch_size
+    header_written = False
 
     for i in range(0, len(belief_ids), batch_size):
         batch = belief_ids[i:i + batch_size]
@@ -160,6 +172,8 @@ def detect_contradictions(nodes, belief_ids=None, model="claude", timeout=300,
             results = parse_contradiction_response(response,
                                                    valid_ids=valid_ids)
             all_results.extend(results)
+            header_written = _flush_results(results, output_path,
+                                            header_written)
         except Exception as e:
             print(f"  WARN: batch {batch_num} failed: {e}", file=sys.stderr)
 
@@ -167,7 +181,8 @@ def detect_contradictions(nodes, belief_ids=None, model="claude", timeout=300,
 
 
 def detect_contradictions_semantic(nodes, belief_ids=None, model="claude",
-                                   timeout=300, embedding_model=None):
+                                   timeout=300, embedding_model=None,
+                                   output_path=None):
     """Detect contradictions by clustering beliefs semantically before LLM analysis.
 
     Groups beliefs by semantic similarity so topically related beliefs
@@ -179,6 +194,7 @@ def detect_contradictions_semantic(nodes, belief_ids=None, model="claude",
         model: LLM model to use.
         timeout: LLM timeout in seconds.
         embedding_model: Sentence-transformers model name.
+        output_path: If set, write results incrementally to this file.
 
     Returns: list of contradiction dicts.
     """
@@ -212,6 +228,7 @@ def detect_contradictions_semantic(nodes, belief_ids=None, model="claude",
     valid_ids = set(belief_ids)
     all_results = []
     clusters = result["clusters"]
+    header_written = False
 
     for ci, cluster in enumerate(clusters, 1):
         cluster_ids = [b["id"] for b in cluster["beliefs"]]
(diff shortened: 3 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#131
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **tests/test_contradictions.py:TestDetectContradictionsSemantic** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### reasons_lib/contradictions.py:detect_contradictions_semantic
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Logic is sound — filters to IN beliefs, clusters via `list_clusters`, skips single-belief clusters, batches within clusters using existing `CONTRADICTION_BATCH_SIZE`, and reuses the same prompt/parse pipeline as the non-semantic path. One subtle point: `list_clusters` puts all beliefs into a single cluster when `len(beliefs) <= 3` (cluster.py:167), so the `len(cluster_ids) < 2` skip at line 218 only fires when 0 or 1 beliefs are passed — correct. The dedup concern is minor: if a belief appears in multiple clusters (it can't with KMeans hard assignment), duplicates would be caught by `parse_contradiction_response`'s valid_ids filter anyway. Tests cover empty network and single-belief skip, but don't verify that contradictions found within a cluster are actually returned (only checks that the LLM was called with the right IDs). A test with a mock LLM returning a NOGOOD and asserting it appears in results would strengthen coverage.

### reasons_lib/api.py:detect_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `semantic` and `embedding_model` parameters are threaded through cleanly. The conditional import of `_detect_semantic` only when `semantic=True` avoids loading sentence-transformers when not needed — good. The `embedding_model` parameter is only passed to the semantic path, which is correct since the non-semantic path doesn't use embeddings. No API test exercises the new `semantic=True` path through `detect_contradictions` (the tests call the inner function directly), but the CLI integration test gap is minor since the plumbing is straightforward passthrough.

### reasons_lib/cli.py:cmd_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

CLI args `--semantic` and `--embedding-model` are defined and passed through to `api.detect_contradictions`. The argparse setup is correct (`store_true` for the flag, `default=None` for the model string). No CLI-level test exercises `--semantic`, but existing tests for `cmd_contradictions` cover the non-semantic path and the new args are pure passthrough with no branching logic in the CLI layer.

### tests/test_contradictions.py:TestDetectContradictionsSemantic
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `test_semantic_groups_by_cluster` test has a fragile assertion. It checks `mock_run.call_args_list` using `call[1].get("input", "")` — this assumes `subprocess.run` is called with `input` as a keyword arg, which depends on `invoke_model`'s internal implementation. If that implementation changes (e.g., passes input positionally), the assertion silently passes with empty strings. More importantly, the test doesn't verify that a contradiction *detected* by the mock LLM is actually returned in the result list — it only checks the LLM was invoked. The `test_semantic_single_belief_skips_llm` test is clever and correct: with 1 IN belief, `list_clusters` returns a single cluster with 1 belief, which is skipped by the `len(cluster_ids) < 2` guard. The `@skip_no_cluster` decorator with the `HAS_CLUSTER_DEPS` import-guard pattern is well done — tests degrade gracefully when optional deps are missing.

### Self-Review
**Limitations:** Could not see the full `invoke_model` implementation to verify how `subprocess.run` is called (positional vs keyword `input`), which affects whether the test assertion on `call[1].get("input", "")` is robust. Did not run the tests to confirm they pass on the current branch.

### Feature Requests
- Include the implementation of functions called by the diff (e.g., `invoke_model`, `list_clusters`) inline in the observations section so reviewers can verify integration without extra lookups
- Show test execution results (pass/fail/skip counts) alongside the diff

## gemini [PASS]

### reasons_lib/api.py:detect_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function is correctly updated to support semantic analysis. It maintains backward compatibility with existing calls while providing an opt-in path for semantic clustering. The internal sampling and filtering logic correctly prepares the node set regardless of the analysis method used.

### reasons_lib/cli.py:cmd_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly wires the new CLI arguments to the API layer.

### reasons_lib/cli.py:main
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Properly exposes the `--semantic` and `--embedding-model` flags to the user, with appropriate help text.

### reasons_lib/contradictions.py:detect_contradictions_semantic
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Implements a sound strategy for scaling contradiction detection. By clustering beliefs semantically, it increases the likelihood that the LLM will encounter related but contradictory claims within the same context window. The implementation handles batching within clusters and correctly ignores clusters too small to contain contradictions.

### tests/test_contradictions.py:TestDetectContradictionsSemantic
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The test suite effectively validates the new functionality. It covers the success path (verifying that related items are grouped in prompts), empty network scenarios, and the single-belief optimization. The use of conditional skipping based on optional dependencies (`sentence-transformers`, `scikit-learn`) ensures the tests don't fail in environments without those libraries.

### Self-Review
**Limitations:** I did not have the full content of `reasons_lib/cluster.py` to verify the exact return type of `list_clusters`, though the tests confirm that the implementation in `contradictions.py` consumes it correctly. I also assumed `CONTRADICTION_BATCH_SIZE` and `CONTRADICTION_PROMPT` are defined as module-level constants as they were not in the diff hunks but were used in the new functions.

### Feature Requests
- Include full file context or at least module-level constants for the modified files to verify all references.
- Show signatures of functions imported from other modules within the project.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#131
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **reasons_lib/contradictions.py:detect_contradictions_semantic** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### reasons_lib/api.py:detect_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Clean routing — when `semantic=True`, imports and calls `detect_contradictions_semantic` with the `embedding_model` kwarg threaded correctly. The conditional import inside the `if semantic` block is fine (avoids importing cluster deps when not needed). The rest of the function (sampling, auto_apply, result assembly) is shared across both paths, which is correct. The only gap is that the API-level `detect_contradictions` with `semantic=True` is not tested directly — the tests exercise the underlying `detect_contradictions_semantic` function from `contradictions.py` directly instead.

### reasons_lib/cli.py:cmd_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

`args.semantic` and `args.embedding_model` are correctly forwarded to `api.detect_contradictions`. The argparse definitions use sensible defaults (`store_true` for `--semantic`, `None` for `--embedding-model`). No CLI-level test exercises `--semantic`, but this is a thin passthrough and the integration risk is low.

### reasons_lib/contradictions.py:detect_contradictions_semantic
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function works correctly for the happy path, but has a subtle behavioral difference from the non-semantic path: the original `detect_contradictions` shuffles `belief_ids` before batching (`random.shuffle(belief_ids)` — visible in the full file), which randomizes which beliefs land in the same batch. The semantic version does not shuffle within clusters, so ordering within a cluster is deterministic (sorted by ID from `sorted(candidates.keys())` upstream). This is arguably better for reproducibility, but it's a difference worth noting.

More substantively: when a single cluster exceeds `CONTRADICTION_BATCH_SIZE` (50), it gets subdivided into sequential slices. Beliefs at position 49 and 50 will never be compared to each other within the same LLM call. The original function has the same limitation for beliefs in different batches, but the semantic mode's value proposition is "group related beliefs together" — so splitting a large topical cluster defeats part of the purpose. Consider overlapping windows or at least documenting this limitation.

Additionally, the function silently returns `[]` when `list_clusters` puts every belief into its own cluster (all singletons), since clusters with `< 2` beliefs are skipped. This is correct behavior but could be surprising if the embedding model produces poor clusters.

### tests/test_contradictions.py:TestDetectContradictionsSemantic
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Four tests cover the key paths — clustering happens, contradictions are returned, empty networks return `[]`, and single-belief networks skip the LLM. However:

1. `test_semantic_groups_by_cluster` asserts that both `auth-login-validates` and `auth-login-skips-validation` appear in prompts, but doesn't verify they appear in the *same* prompt/batch. The assertion checks `" ".join(all_prompts)` which concatenates all calls — so this test would pass even if the two beliefs were sent in separate batches to separate LLM calls. This means the test doesn't actually verify the clustering behavior it claims to test.

2. `test_semantic_single_belief_skips_llm` — only one IN belief exists, but `list_clusters` is still called (embedding computation happens). The test asserts `mock_run.assert_not_called()` which verifies the LLM subprocess isn't called, but the sentence-transformers model still runs. The test name says "skips LLM" which is accurate, but it would be worth also testing that a single-element cluster (not single-element network) is skipped.

3. The `skip_no_cluster` decorator correctly gates on `HAS_CLUSTER_DEPS`, so these tests won't produce false failures in environments without the ML deps.

### Self-Review
**Limitations:** Could not verify the full `detect_contradictions` original function to confirm the shuffle behavior I referenced — relying on the explore agent's report. Could not see the full `list_clusters` implementation to verify whether it can produce degenerate clusterings (all singletons, one giant cluster) that would affect the semantic path. Could not verify whether `parse_contradiction_response` handles edge cases in claim IDs that might differ between semantic and non-semantic paths.

### Feature Requests
- Include the full body of functions modified in the diff (not just changed lines) so reviewers can verify surrounding logic
- Show the implementation of called functions (e.g., `list_clusters`, `parse_contradiction_response`) when they're critical to understanding correctness
- Flag when test assertions don't actually verify the behavior described by the test name (semantic verification of test intent)

## gemini [PASS]

### reasons_lib/api.py:detect_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function signature was updated to include `semantic` and `embedding_model` parameters. It correctly delegates to the new `detect_contradictions_semantic` implementation when requested. The use of local imports for the semantic implementation is a good practice to keep the heavy dependencies (sentence-transformers) optional.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `contradictions` command now supports `--semantic` and `--embedding-model` flags. The defaults are consistent with the rest of the system (e.g., `all-MiniLM-L6-v2` for the embedding model).

### reasons_lib/contradictions.py:detect_contradictions_semantic
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation correctly leverages `list_clusters` to group beliefs. It handles cluster-level batching and skips clusters with fewer than two beliefs (as a contradiction requires at least two claims). The logic respects the `IN` truth-value requirement and properly filters `belief_ids`.

### tests/test_contradictions.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New tests cover semantic grouping, found contradictions, and edge cases (empty network, single belief). The tests correctly use `@skip_no_cluster` to handle environments without the required ML dependencies.

### Self-Review
**Limitations:** I assumed that `CONTRADICTION_BATCH_SIZE` and `CONTRADICTION_PROMPT` were correctly defined in `contradictions.py` based on their usage in the existing `detect_contradictions` function. I verified this by reading the file.

### Feature Requests
- It might be useful to allow overlapping batches within a cluster if a cluster exceeds `CONTRADICTION_BATCH_SIZE`, as the current implementation might still miss contradictions between beliefs at the boundaries of batches.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#131
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **reasons_lib/api.py:parse_contradiction_plan** [LOW]: claude: CONCERN, gemini: PASS
- **reasons_lib/contradictions.py:detect_contradictions_semantic** [LOW]: claude: CONCERN, gemini: PASS
- **reasons_lib/cli.py:cmd_contradictions** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### reasons_lib/api.py:detect_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean addition of `semantic`, `embedding_model`, and `output_path` parameters with appropriate forwarding. The conditional dispatch to `_detect_semantic` vs `_detect` is straightforward. The `output_path` is suppressed when `auto_apply` is true (line 1363 of cli.py), which is correct — auto-apply and plan-file are mutually exclusive workflows.

### reasons_lib/api.py:write_contradiction_plan
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean file writer with append mode for incremental saves. The format is parseable by `parse_contradiction_plan`. The header includes usage instructions pointing to `--accept`. Tests verify both fresh write and append modes.

### reasons_lib/api.py:parse_contradiction_plan
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The parser has a subtle issue: it uses `m.group(2) == "APPLY" and m.group(1) or None` which relies on Python truthiness. If a nogood ID happens to be `"0"` or any other falsy string, the `and/or` expression would incorrectly produce `None`. Should use a ternary `m.group(1) if m.group(2) == "APPLY" else None` instead. Additionally, backtick-stripping in claims parsing (`.strip("\`")`) is asymmetric — `write_contradiction_plan` never writes backticks, so this is dead code that might mask a format mismatch. Tests don't cover edge cases like multi-line claims or malformed entries with missing claims lines.

### reasons_lib/api.py:apply_contradiction_plan
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple loop over plan entries calling `add_nogood`. Error handling is reasonable — individual failures are collected without aborting the entire plan. The return structure matches what cli.py expects. Tested with a real DB round-trip in `test_apply_contradiction_plan`.

### reasons_lib/contradictions.py:_flush_results
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Small helper that handles incremental writes correctly. Uses lazy import from `.api` to avoid circular dependency (both `api.py` and `contradictions.py` already use lazy cross-imports). The `header_written` flag ensures the header is only written once. Not directly unit-tested but exercised through integration tests (`test_output_writes_plan`).

### reasons_lib/contradictions.py:detect_contradictions_semantic
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Two concerns: (1) Beliefs that end up in singleton clusters (< 2 members) are silently skipped, meaning they're never checked for contradictions against any other beliefs. If clustering puts an outlier belief alone, it gets no contradiction checking at all — this is a coverage gap that could miss cross-cluster contradictions. (2) No deduplication of results across clusters — if a belief appears in multiple clusters (unlikely with hard clustering but possible depending on the clustering impl), the same contradiction could be reported twice. Both are acceptable tradeoffs if documented, but the silent skipping of singletons should at least be logged.

### reasons_lib/cli.py:cmd_contradictions
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `--dry-run` flag was removed but there's no migration guidance. More importantly, when `--auto-apply` is **not** set, the CLI unconditionally prints "Wrote {output} — review, then run:" (line 1390), but this message prints even when no contradictions are found — the early return on line 1369-1370 prevents reaching it for the zero-contradiction case, but the plan file is still written incrementally by the detection layer and may contain just a header. Also, `--output` now defaults to `"proposed-contradictions.md"` instead of `None`, meaning every non-`--auto-apply` run writes a file to the current directory even if the user didn't ask for it. This is a behavior change from the old default of `None`.

### reasons_lib/cli.py (argparse changes)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Removing `--dry-run` is a breaking change for existing users/scripts. The old `--dry-run` prevented `--auto-apply` from taking effect. The new workflow replaces this with the plan file approach, which is better, but any existing automation using `--dry-run` will fail with an unrecognized argument error. The default `--output` of `"proposed-contradictions.md"` means the command always creates a file — this side effect on a default path could surprise users.

### tests/test_contradictions.py:TestDetectContradictionsSemantic
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good test coverage of the semantic path: verifies clustering groups related beliefs, verifies found contradictions are returned, verifies empty/single-belief edge cases skip the LLM. The `skipif` guard on `HAS_CLUSTER_DEPS` is appropriate for optional dependency handling.

### tests/test_contradictions.py:TestContradictionPlan
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Thorough test coverage of the plan write/parse/apply/CLI cycle. Tests cover: format correctness, append mode, APPLY/SKIP filtering, full CLI round-trip, missing file error, and empty plan. The `test_accept_cli_flow` test is a good end-to-end integration test.

### Self-Review
**Limitations:** Could not see the full `cluster.py` implementation to verify whether beliefs can appear in multiple clusters. Could not verify whether existing CI or scripts depend on `--dry-run`. Did not see the full test runner output to confirm all existing tests still pass with the `--dry-run` removal.

### Feature Requests
- Include the full file for modified functions (not just diff hunks) — particularly useful for verifying early-return paths and control flow around the changed code
- Show callers of modified/removed CLI flags to assess breaking change impact
- Flag removed CLI arguments/flags as potential breaking changes automatically

## gemini [CONCERN]

### reasons_lib/api.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The logic for writing, parsing, and applying contradiction plans is sound. However, there is a missing import for `Path` in this file. The code uses `path = Path(output_path)` but `Path` (from `pathlib`) is not imported in the provided diff or the beginning of the file (verified via `grep`). This will cause a `NameError` at runtime.

### reasons_lib/api.py:detect_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Properly integrates the new semantic and incremental output parameters. It maintains backward compatibility for the `auto_apply` logic while enabling the new workflow.

### reasons_lib/api.py:parse_contradiction_plan
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The parsing logic is robust against extra text and correctly filters for entries marked `[APPLY]`. The regex and state management correctly handle multiple entries and varying claim formats.

### reasons_lib/cli.py:cmd_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The CLI command successfully wires together the new detection and application flows. The removal of the redundant `--dry-run` flag in favor of the default-to-plan-file behavior is a good UX improvement.

### reasons_lib/contradictions.py:detect_contradictions_semantic
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

This is a well-implemented feature. By clustering beliefs before LLM analysis, the tool can focus on topically related nodes, which is where contradictions are most likely to occur. The incremental flush logic ensures results are not lost during long-running semantic analysis.

### reasons_lib/contradictions.py:_flush_results
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly handles incremental writing to the plan file. It uses a lazy import of `write_contradiction_plan` to avoid top-level circular dependencies with `api.py`.

### tests/test_contradictions.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new tests provide good coverage for the semantic clustering, the plan file format, and the full "detect -> review -> accept" cycle.

### Self-Review
**Limitations:** I verified the missing `Path` import by checking the existing imports in `reasons_lib/api.py` using `grep`. The rest of the review is based on the provided diff and full function bodies.

### Feature Requests
- Ensure `pathlib.Path` is imported in `reasons_lib/api.py`.
- Consider moving `write_contradiction_plan` and `parse_contradiction_plan` to a dedicated `reasons_lib/plans.py` to avoid the circular dependency between `api.py` and `contradictions.py`.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#131
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/api.py:detect_contradictions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new `semantic`, `embedding_model`, and `output_path` parameters are correctly threaded through to the underlying `_detect` and `_detect_semantic` functions. The conditional import of `_detect_semantic` only when `semantic=True` avoids pulling in heavy dependencies unnecessarily. The `output_path` is correctly suppressed when `auto_apply` is True at the CLI layer (not here — this function just passes it through, which is correct).

### reasons_lib/api.py:write_contradiction_plan
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The write order is Severity then Claims, but `parse_contradiction_plan` only looks for `- Claims:` lines under the current heading. This works because the parser scans all lines. However, if a user edits the plan and a claim ID contains a comma (e.g., `my,claim`), both the writer (`', '.join`) and parser (`split(",")`) will produce incorrect results. This is a latent edge case — belief IDs with commas are likely invalid, but there's no validation or documentation of this constraint. Minor: the `output_path` is embedded literally in the plan file header instruction (`reasons contradictions --accept {output_path}`), which could produce confusing instructions if the path contains spaces or special characters.

### reasons_lib/api.py:parse_contradiction_plan
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The parser has a subtle issue with the SKIP→APPLY transition. When a SKIP entry is encountered, `current_id` is set to `None` via the ternary, but `current_claims` is reset to `[]`. If a SKIP entry is followed by an APPLY entry, the APPLY entry's claims are correctly captured because `current_claims` was reset. This is correct. However, if the last entry in the file is SKIP, the final `if current_id and current_claims` guard correctly prevents it from being added. The `strip("`")` in the claims parser handles backtick-wrapped IDs — good defensive parsing. Missing test: no test covers a SKIP entry followed by an APPLY entry to verify the state machine transitions correctly. No test covers IDs with backticks to exercise the `strip("`")` path.

### reasons_lib/api.py:apply_contradiction_plan
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Straightforward loop over plan entries calling `add_nogood`. Errors are caught per-entry so a single failure doesn't abort the whole plan. The return structure includes both applied nogoods and errors, enabling the CLI to report partial success. Test `test_apply_contradiction_plan` verifies the happy path with real DB operations.

### reasons_lib/contradictions.py:_flush_results
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Clean helper for incremental writes. Uses `header_written` as a boolean to decide append vs. create mode. The `from .api import write_contradiction_plan` is a circular-ish import (contradictions imports api, api imports contradictions), but since it's inside a function body it avoids circular import issues at module load time. Tested indirectly through `test_output_writes_plan` but no direct unit test for `_flush_results` itself.

### reasons_lib/contradictions.py:detect_contradictions (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `output_path` parameter is correctly threaded through to `_flush_results` after each batch. The `header_written` state variable is properly initialized before the loop and updated after each flush. Existing behavior is preserved when `output_path` is None.

### reasons_lib/contradictions.py:detect_contradictions_semantic
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The function skips clusters with fewer than 2 beliefs (`if len(cluster_ids) < 2: continue`). This means beliefs in singleton clusters are never checked for contradictions — they silently disappear from analysis. If the clustering algorithm produces many singletons, coverage could be significantly reduced. The test `test_semantic_single_belief_skips_llm` tests a single-belief *network*, not a singleton *cluster* within a multi-belief network. No test verifies that beliefs in singleton clusters are reported or that the caller is informed of reduced coverage. Also, there is no deduplication: if the same belief appears in multiple clusters (depending on the clustering algorithm), the same contradiction could be detected and reported twice.

### reasons_lib/cli.py:cmd_contradictions (modified)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Three issues: (1) **Breaking change**: `--dry-run` was removed with no deprecation. Existing scripts or automation using `--dry-run` will fail with an unrecognized argument error. (2) **Default output path always creates a file**: `--output` defaults to `"proposed-contradictions.md"`, and when `--auto-apply` is not set, the output_path is always passed to `detect_contradictions`. This means running `reasons contradictions` without `-o` will always create `proposed-contradictions.md` in the current directory — even if no contradictions are found (the file won't be created in that case because `_flush_results` returns early on empty results, but the "Wrote" message at cli.py:1385 will still print). Actually, looking more carefully: when no contradictions are found, the early return at line 1369-1370 prevents the "Wrote" message. But if contradictions *are* found, the plan file is written incrementally during detection (via `_flush_results`), and then the CLI also prints "Wrote {output}" — this is correct. (3) **`--accept` path check uses `sys.exit(1)` directly** instead of returning a non-zero code through the normal test harness flow — but the test `test_accept_missing_file` does verify `code == 1`, so this works.

### reasons_lib/cli.py argparse (contradictions subparser)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New flags `--semantic`, `--embedding-model`, `--accept`, and updated `-o` default are properly wired. The `--accept` uses `metavar="FILE"` for clear help text. The `--dry-run` removal is clean from the parser perspective.

### tests/test_contradictions.py:TestCmdContradictions (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Tests properly updated for the new workflow. `test_no_contradictions_message` no longer passes `--dry-run`. `test_displays_found_contradictions` now passes `-o` with a tmp_path file. `test_output_writes_plan` verifies the new plan format including the `[APPLY]` tag and `--accept` instruction. The old `test_dry_run_prevents_apply` test was correctly removed since `--dry-run` no longer exists.

### tests/test_contradictions.py:TestDetectContradictionsSemantic
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

`test_semantic_groups_by_cluster` checks that the auth beliefs appear in the prompts sent to the LLM, but the assertion `call[1].get("input", "")` accesses keyword arg `input` from `subprocess.run` — this is correct for the `claude` CLI invocation. `test_semantic_returns_found_contradictions` is a good functional test. However, the skip condition `@skip_no_cluster` relies on importing `HAS_CLUSTER_DEPS` which may import heavy dependencies — the `try/except ImportError` wrapper handles this gracefully. Missing: no test for `output_path` parameter in semantic mode, and no test for clusters with beliefs that exceed `CONTRADICTION_BATCH_SIZE` (the inner batch loop).

### tests/test_contradictions.py:TestContradictionPlan
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test class covering: write format, append mode, parse with APPLY/SKIP, parse with deleted entries, apply with real DB, full CLI accept flow, missing file error, and empty plan. The `test_accept_cli_flow` is particularly good — it does a full round-trip: init DB, add nodes, write plan file, run `--accept`, verify the nogood was applied.

### Self-Review
**Limitations:** Could not see the full `cluster.py` to verify whether beliefs can appear in multiple clusters or whether clustering produces many singletons. Could not run the test suite to confirm all existing tests pass after the `--dry-run` removal. Could not verify whether any CI scripts or documentation reference `--dry-run` for the contradictions command.

### Feature Requests
- Include the full source of newly-imported modules (e.g., `cluster.py`) when they are referenced in the diff but not included
- Show all callers of removed CLI flags to assess breaking change impact
- Include the existing test file in full when tests are modified, to verify no other test depends on removed functionality

## gemini [PASS]

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new functions `write_contradiction_plan`, `parse_contradiction_plan`, and `apply_contradiction_plan` correctly implement a robust human-in-the-loop workflow for managing contradictions. The update to `detect_contradictions` to support semantic clustering and incremental output is well-structured and handles optional parameters correctly. The regex-based parsing in `parse_contradiction_plan` is robust for the expected markdown format.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The CLI changes effectively expose the new contradiction management features. The `--accept` flow is clearly implemented at the start of `cmd_contradictions`, and the transition from the old `--dry-run` flag to a plan-based approach is handled smoothly with helpful user messages guiding the next steps.

### reasons_lib/contradictions.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The addition of `detect_contradictions_semantic` provides a valuable optimization for detecting conflicts in large networks by grouping topically related beliefs before LLM analysis. The incremental result flushing via `_flush_results` is a significant improvement for UX and reliability during long-running batch processes.

### tests/test_contradictions.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New tests provide comprehensive coverage for the semantic detection logic and the full lifecycle of contradiction plans (writing, parsing, and applying). The use of `pytest.mark.skipif` for optional clustering dependencies ensures the test suite remains portable.

### Self-Review
**Limitations:** I could not verify if `pathlib.Path` is imported at the top of `api.py` and `cli.py` as the module-level imports were not included in the diff. However, given its usage in the new code and the project's overall structure, I assume it is either already imported or will be caught by linting if missing.

### Feature Requests
- Include module-level imports in diffs to verify availability of classes/functions like `Path` or `re`.

PR #136, opened 2026-05-11
Add MCP server support for reasons ask

### LINKED ISSUE #48: Add data source adapters for `reasons ask`
## Summary

Add optional data source adapters that the LLM can invoke during `reasons ask` when beliefs are insufficient to answer the question.

## Background

Issue #43 shipped the core `reasons ask` command (PR #44, #47) with FTS5 belief search + LLM synthesis. The current implementation only searches beliefs. This issue adds the ability to query external data sources.

## Proposed behavior

```bash
# With a PostgreSQL adapter
reasons ask "How many active users last month?" --source postgres://...

# With a custom adapter
reasons ask "What's the pipeline status?" --source ./my_adapter.py
```

## Design

- Adapter interface: `query(question: str) -> str`
- The LLM decides whether to call an adapter based on belief sufficiency (same tool loop pattern)
- Durable facts from adapter results can be proposed as new beliefs
- Belief proposals from adapter results should respect the durability filter (permanent/quarterly only)

## Prior art

- `mini-analyze/src/mini_analyze/web.py` — tool loop where LLM decides whether to query external source
- #38 — RBAC for cached beliefs (relevant when adapter results become beliefs)

benthomasson:
## Summary
- New `--mcp` flag on `reasons ask` connects to MCP servers as data source adapters
- MCP tools are added to the prompt alongside `search_beliefs` — any MCP server becomes a tool source
- Server instructions (e.g., Snowflake data catalog) injected into the prompt automatically
- Tool loop iteration budget bumps from 3 to 5 when MCP servers are attached
- New `reasons_lib/mcp_client.py` module wraps the async MCP SDK with a sync `McpBridge` class using a background event loop thread

Closes #48

## Test plan
- [x] `TestBuildToolsSection` — builds tool definitions from MCP servers (3 tests)
- [x] `TestBuildMcpInstructions` — collects server instructions (2 tests)
- [x] `TestAskPromptWithMcp` — prompt injection of tools and instructions (4 tests)
- [x] `TestExtractToolCallMcp` — tool call parsing for MCP tools (2 tests)
- [x] Existing ask tests updated and passing (78 tests)
- [x] Full test suite: 900 passed, 1 skipped
- [x] Manual: `McpBridge("snowflake-mcp")` connects, lists 5 tools, gets 22K chars of instructions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ test = ["pytest", "pytest-cov"]
 pg = ["psycopg[binary]>=3.1"]
 test-pg = ["psycopg[binary]>=3.1", "pytest", "pytest-cov"]
 cluster = ["sentence-transformers>=2.2", "scikit-learn>=1.2"]
+mcp = ["mcp>=1.0"]
 
 [project.scripts]
 reasons = "reasons_lib.cli:main"
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -23,22 +23,20 @@ You are answering a question using a belief network (a Truth Maintenance System)
 Each belief has an ID, text, truth value (IN = held true, OUT = retracted), and
 may have justifications tracing why it is believed.
 
-You have one tool available:
-
-{{"tool": "search_beliefs", "query": "search terms"}}
+{tools_section}
 
 Rules:
 - If the belief matches below are sufficient to answer the question, write your
-  answer directly. Do NOT call the tool.
-- If you need to search for more beliefs, respond with ONLY a single JSON line
-  (no other text). The system will run the search and give you the results.
+  answer directly. Do NOT call a tool.
+- If you need more information, respond with ONLY a single JSON line
+  (no other text). The system will run the tool and give you the results.
 {cite_rule}
-- ONLY answer based on the beliefs provided. Do NOT use your training data or
-  general knowledge to fill gaps.
+- ONLY answer based on the beliefs and data provided. Do NOT use your training
+  data or general knowledge to fill gaps.
 - If the beliefs are insufficient to answer, respond EXACTLY with:
   "I don't have enough beliefs in the network to answer this question."
   Do NOT attempt a partial or speculative answer.
-
+{mcp_instructions}
 ## Question
 
 {question}
@@ -47,6 +45,12 @@ Rules:
 
 {beliefs_context}
 {tool_history}"""
+
+
+_DEFAULT_TOOLS_SECTION = """\
+You have one tool available:
+
+{{"tool": "search_beliefs", "query": "search terms"}}"""
 
 
 FINAL_ASK_PROMPT = """\
@@ -123,23 +127,29 @@ def extract_tool_call(text):
     return None
 
 
-def build_ask_prompt(question, beliefs_context, tool_history=None, natural=False):
+def build_ask_prompt(question, beliefs_context, tool_history=None, natural=False,
+                     tools_section=None, mcp_instructions=""):
     """Build the full prompt for LLM synthesis."""
     history_section = ""
     if tool_history:
         parts = []
         for entry in tool_history:
             parts.append(
-                f"### Tool call: search_beliefs(\"{entry['query']}\")\n\n"
+                f"### Tool call: {entry['tool_label']}\n\n"
                 f"{entry['result']}"
             )
         history_section = "\n\n## Additional search results\n\n" + "\n\n---\n\n".join(parts)
+
+    if mcp_instructions:
+        mcp_instructions = f"\n## Data Source Instructions\n\n{mcp_instructions}\n"
 
     return ASK_PROMPT.format(
         question=question,
         beliefs_context=beliefs_context,
         tool_history=history_section,
         cite_rule=_CITE_RULE_NATURAL if natural else _CITE_RULE,
+        tools_section=tools_section or _DEFAULT_TOOLS_SECTION,
+        mcp_instructions=mcp_instructions,
     )
 
 
@@ -150,7 +160,7 @@ def build_final_prompt(question, beliefs_context, tool_history=None, natural=Fal
         parts = []
         for entry in tool_history:
             parts.append(
-                f"### Tool call: search_beliefs(\"{entry['query']}\")\n\n"
+                f"### Tool call: {entry['tool_label']}\n\n"
                 f"{entry['result']}"
             )
         history_section = "\n\n## Additional search results\n\n" + "\n\n---\n\n".join(parts)
@@ -331,8 +341,39 @@ def _beliefs_or_no_match(beliefs_context):
     return beliefs_context
 
 
+def _build_tools_section(mcp_servers):
+    """Build the tools section for the prompt, including MCP server tools."""
+    lines = ["You have the following tools available:", "",
+             '{"tool": "search_beliefs", "query": "search terms"}']
+    if mcp_servers:
+        for bridge in mcp_servers:
+            for tool in bridge.list_tools():
+                schema = tool.get("input_schema", {})
+                props = schema.get("properties", {})
+                param_parts = []
+                for pname, pinfo in props.items():
+                    param_parts.append(f'"{pname}": "<{pinfo.get("description", pname)}>"')
+                example = ', '.join(param_parts)
+                lines.append(f'{{"tool": "{tool["name"]}", {example}}}')
+                if tool["description"]:
+                    desc = tool["description"].split("\n")[0]
+                    lines.append(f"  # {desc}")
+    return "\n".join(lines)
+
+
+def _build_mcp_instructions(mcp_servers):
+    """Collect server instructions from all MCP servers."""
+    parts = []
+    for bridge in mcp_servers:
+        instructions = bridge.get_instructions()
+        if instructions:
+            parts.append(instructions)
+    return "\n\n".join(parts)
+
+
 def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None,
-        model="claude", simple=False, sources_db=None, natural=False, dual=False):
+        model="claude", simple=False, sources_db=None, natural=False, dual=False,
+        mcp_servers=None):
     """Answer a question using FTS5 belief search and optional LLM synthesis.
 
     Args:
@@ -346,6 +387,7 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
               calls with simple=True (1 TMS + 1 FTS + 1 merge), or up to
               5 with simple=False (up to 3 TMS tool-loop rounds + 1 FTS
               + 1 merge). Timeout applies per call.
+        mcp_servers: list of connected McpBridge instances for external tools.
 
     Returns the answer text.
     """
@@ -423,16 +465,31 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
             return (beliefs_context or "") + sources_suffix
         return beliefs_context
 
+    tools_section = None
+    mcp_instructions = ""
+    mcp_tool_map = {}
+    max_iters = MAX_ITERATIONS
+
+    if mcp_servers:
+        tools_section = _build_tools_section(mcp_servers)
+        mcp_instructions = _build_mcp_instructions(mcp_servers)
+        for bridge in mcp_servers:
+            for tool in bridge.list_tools():
+                mcp_tool_map[tool["name"]] = bridge
+        max_iters = max(MAX_ITERATIONS, 5)
+
     tool_history = []
 
-    for iteration in range(MAX_ITERATIONS):
+    for iteration in range(max_iters):
         ctx = _full_context()
-        if iteration == MAX_ITERATIONS - 1:
+        if iteration == max_iters - 1:
             prompt = build_final_prompt(question, ctx, tool_history, natural=natural)
         else:
-            prompt = build_ask_prompt(question, ctx, tool_history, natural=natural)
+            prompt = build_ask_prompt(question, ctx, tool_history, natural=natural,
+                                      tools_section=tools_section,
+                                      mcp_instructions=mcp_instructions)
 
-        print(f"Synthesizing (round {iteration + 1}/{MAX_ITERATIONS})...",
+        print(f"Synthesizing (round {iteration + 1}/{max_iters})...",
               file=sys.stderr)
 
         try:
@@ -449,20 +506,37 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
         if tool_call is None:
             return response.strip()
 
-        if tool_call.get("tool") == "search_beliefs":
+        tool_name = tool_call.get("tool")
+
+        if tool_name == "search_beliefs":
             query = tool_call.get("query", "")
             print(f"  Searching: {query}", file=sys.stderr)
             result = api.search(query, db_path=db_path, format="markdown")
             history_result = _strip_belief_metadata(result) if natural and result else result
-            tool_history.append({"query": query, "result": history_result})
+            tool_history.append({
+                "tool_label": f'search_beliefs("{query}")',
+                "result": history_result,
+            })
             if result and result.strip() != "No results found.":
                 beliefs_context = result
                 if natural:
                     beliefs_context = _strip_belief_metadata(beliefs_context)
+        elif tool_name in mcp_tool_map:
+            bridge = mcp_tool_map[tool_name]
+            args = {k: v for k, v in tool_call.items() if k != "tool"}
+            print(f"  MCP tool: {tool_name}({json.dumps(args)[:80]})", file=sys.stderr)
+            try:
+                result = bridge.call_tool(tool_name, args)
+            except Exception as e:
+                result = f"Error calling {tool_name}: {e}"
+            tool_history.append({
+                "tool_label": f"{tool_name}(...)",
+                "result": result,
+            })
         else:
             return response.strip()
 
-        if iteration == MAX_ITERATIONS - 1:
+        if iteration == max_iters - 1:
             print(f"Synthesizing (final)...", file=sys.stderr)
             ctx = _full_context()
             prompt = build_final_prompt(question, ctx, tool_history, natural=natural)
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -400,7 +400,8 @@ def ask(question, db_path="reasons.db", timeout=300, no_synth=False, format=None
     if dual and sources_db:
         print("Dual path: running TMS...", file=sys.stderr)
         answer_tms = ask(question, db_path=db_path, timeout=timeout,
-                         model=model, simple=simple, natural=natural)
+                         model=model, simple=simple, natural=natural,
+                         mcp_servers=mcp_servers)
         print("Dual path: running FTS RAG...", file=sys.stderr)
         answer_fts = _fts_rag_answer(question, sources_db, model=model,
                                      timeout=timeout)
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -50,7 +50,7 @@ Rules:
 _DEFAULT_TOOLS_SECTION = """\
 You have one tool available:
 
-{{"tool": "search_beliefs", "query": "search terms"}}"""
+{"tool": "search_beliefs", "query": "search terms"}"""
 
 
 FINAL_ASK_PROMPT = """\
--- a/reasons_lib/ask.py
+++ b/reasons_lib/ask.py
@@ -354,7 +354,10 @@ def _build_tools_section(mcp_servers):
                 for pname, pinfo in props.items():
                     param_parts.append(f'"{pname}": "<{pinfo.get("description", pname)}>"')
                 example = ', '.join(param_parts)
-                lines.append(f'{{"tool": "{tool["name"]}", {example}}}')
+                if example:
+                    lines.append(f'{{"tool": "{tool["name"]}", {example}}}')
+                else:
+                    lines.append(f'{{"tool": "{tool["name"]}"}}')
                 if tool["description"]:
                     desc = tool["description"].split("\n")[0]
                     lines.append(f"  # {desc}")
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -708,19 +708,35 @@ def cmd_lookup(args):
 def cmd_ask(args):
     _require_sqlite(args, "ask")
     from .ask import ask
-    result = ask(
-        question=args.question,
-        db_path=args.db,
-        timeout=args.timeout,
-        no_synth=args.no_synth,
-        format=getattr(args, "format", None),
-        model=args.model or "claude",
-        simple=args.simple,
-        sources_db=args.full_sources,
-        natural=args.natural,
-        dual=args.dual,
-    )
-    print(result)
+
+    mcp_servers = None
+    if args.mcp:
+        from .mcp_client import McpBridge
+        mcp_servers = []
+        for cmd in args.mcp:
+            bridge = McpBridge(cmd)
+            bridge.connect()
+            mcp_servers.append(bridge)
+
+    try:
+        result = ask(
+            question=args.question,
+            db_path=args.db,
+            timeout=args.timeout,
+            no_synth=args.no_synth,
+            format=getattr(args, "format", None),
+            model=args.model or "claude",
+            simple=args.simple,
+            sources_db=args.full_sources,
+            natural=args.natural,
+            dual=args.dual,
+            mcp_servers=mcp_servers,
+        )
+        print(result)
+    finally:
+        if mcp_servers:
+            for bridge in mcp_servers:
+                bridge.close()
 
 
 def cmd_cluster_list(args):
@@ -1683,6 +1699,8 @@ def main():
                    help="Strip belief IDs, status, and justification metadata from context")
     p.add_argument("--dual", action="store_true",
                    help="Run TMS and FTS RAG separately, then merge (requires --full-sources)")
+    p.add_argument("--mcp", action="append", default=None,
+                   help="MCP server command as data source (repeatable, e.g. --mcp snowflake-mcp)")
 
     # deduplicate
     p = sub.add_parser("deduplicate", help="Find and optionally retract duplicate IN beliefs")
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -709,16 +709,15 @@ def cmd_ask(args):
     _require_sqlite(args, "ask")
     from .ask import ask
 
-    mcp_servers = None
-    if args.mcp:
-        from .mcp_client import McpBridge
-        mcp_servers = []
-        for cmd in args.mcp:
-            bridge = McpBridge(cmd)
-            bridge.connect()
-            mcp_servers.append(bridge)
-
+    mcp_servers = []
     try:
+        if args.mcp:
+            from .mcp_client import McpBridge
+            for cmd in args.mcp:
+                bridge = McpBridge(cmd)
+                bridge.connect()
+                mcp_servers.append(bridge)
+
         result = ask(
             question=args.question,
             db_path=args.db,
@@ -730,13 +729,12 @@ def cmd_ask(args):
             sources_db=args.full_sources,
             natural=args.natural,
             dual=args.dual,
-            mcp_servers=mcp_servers,
+            mcp_servers=mcp_servers or None,
         )
         print(result)
     finally:
-        if mcp_servers:
-            for bridge in mcp_servers:
-                bridge.close()
+        for bridge in mcp_servers:
+            bridge.close()
 
 
 def cmd_cluster_list(args):
--- a/reasons_lib/mcp_client.py
+++ b/reasons_lib/mcp_client.py
@@ -7,7 +7,7 @@ Requires: pip install 'mcp>=1.0'
 """
 
 import asyncio
-import concurrent.futures
+import shlex
 import threading
 
 try:
@@ -41,6 +41,7 @@ class McpBridge:
         self._loop = None
         self._thread = None
         self._session = None
+        self._shutdown = None
         self._ready = threading.Event()
         self._error = None
 
@@ -50,7 +51,8 @@ class McpBridge:
             target=self._run_loop, daemon=True
         )
         self._thread.start()
-        self._ready.wait(timeout=30)
+        if not self._ready.wait(timeout=30):
+            raise TimeoutError(f"MCP server '{self.command}' did not respond within 30s")
         if self._error:
             raise self._error
 
@@ -59,7 +61,7 @@ class McpBridge:
         self._loop.run_until_complete(self._session_lifecycle())
 
     async def _session_lifecycle(self):
-        parts = self.command.split()
+        parts = shlex.split(self.command)
         server_params = StdioServerParameters(
             command=parts[0], args=parts[1:],
         )
@@ -81,10 +83,9 @@ class McpBridge:
                         }
                         for t in tools_result.tools
                     ]
-                    self._ready.set()
 
-                    # Keep alive until close() is called
                     self._shutdown = asyncio.Event()
+                    self._ready.set()
                     await self._shutdown.wait()
                 except Exception as e:
                     self._error = e
--- a/reasons_lib/mcp_client.py
+++ b/reasons_lib/mcp_client.py
@@ -1,0 +1,130 @@
+"""MCP client bridge for connecting to MCP servers as data source adapters.
+
+Wraps the async MCP SDK with synchronous helpers using a background
+event loop thread, so ask.py can dispatch tool calls to any MCP server.
+
+Requires: pip install 'mcp>=1.0'
+"""
+
+import asyncio
+import concurrent.futures
+import threading
+
+try:
+    from mcp import ClientSession, StdioServerParameters
+    from mcp.client.stdio import stdio_client
+    _mcp_available = True
+except ImportError:
+    _mcp_available = False
+
+
+def _require_mcp():
+    if not _mcp_available:
+        raise ImportError(
+            "mcp is required for MCP server support. "
+            "Install it with: pip install 'mcp>=1.0'"
+        )
+
+
+class McpBridge:
+    """Synchronous bridge to an MCP server over stdio.
+
+    Runs a persistent asyncio event loop in a background thread
+    to keep the MCP session alive across multiple tool calls.
+    """
+
+    def __init__(self, command):
+        _require_mcp()
+        self.command = command
+        self._tools = []
+        self._instructions = ""
+        self._loop = None
+        self._thread = None
+        self._session = None
+        self._ready = threading.Event()
+        self._error = None
+
+    def connect(self):
+        self._loop = asyncio.new_event_loop()
+        self._thread = threading.Thread(
+            target=self._run_loop, daemon=True
+        )
+        self._thread.start()
+        self._ready.wait(timeout=30)
+        if self._error:
+            raise self._error
+
+    def _run_loop(self):
+        asyncio.set_event_loop(self._loop)
+        self._loop.run_until_complete(self._session_lifecycle())
+
+    async def _session_lifecycle(self):
+        parts = self.command.split()
+        server_params = StdioServerParameters(
+            command=parts[0], args=parts[1:],
+        )
+
+        async with stdio_client(server_params) as (read, write):
+            async with ClientSession(read, write) as session:
+                self._session = session
+                try:
+                    result = await session.initialize()
+                    if hasattr(result, "instructions") and result.instructions:
+                        self._instructions = result.instructions
+
+                    tools_result = await session.list_tools()
+                    self._tools = [
+                        {
+                            "name": t.name,
+                            "description": t.description or "",
+                            "input_schema": t.inputSchema,
+                        }
+                        for t in tools_result.tools
+                    ]
+                    self._ready.set()
+
+                    # Keep alive until close() is called
+                    self._shutdown = asyncio.Event()
+                    await self._shutdown.wait()
+                except Exception as e:
+                    self._error = e
+                    self._ready.set()
+
+    def list_tools(self):
+        return self._tools
+
+    def get_instructions(self):
+        return self._instructions
+
+    def call_tool(self, name, arguments):
+        future = asyncio.run_coroutine_threadsafe(
+            self._call_tool_async(name, arguments), self._loop
+        )
+        return future.result(timeout=60)
+
+    async def _call_tool_async(self, name, arguments):
+        result = await self._session.call_tool(name, arguments)
+        parts = []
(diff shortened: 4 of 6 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#136
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### reasons_lib/mcp_client.py (new file)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

1. **`_shutdown` attribute race in `close()`** (line 116): `close()` checks `self._shutdown` but `_shutdown` is only created inside `_session_lifecycle()` (line 87), not in `__init__`. If `close()` is called before `_session_lifecycle` reaches line 87 (e.g., connect times out or fails early), this raises `AttributeError`. Fix: initialize `self._shutdown = None` in `__init__`.

2. **`connect()` timeout is silent** (line 53): `self._ready.wait(timeout=30)` returns `False` on timeout but execution continues to `if self._error` which will be `None`. The caller gets an `McpBridge` with no tools and no session — subsequent `call_tool` calls will fail with a confusing error on `self._session` being `None`. Should raise on timeout.

3. **`command.split()` parsing** (line 62): Naive — doesn't handle quoted arguments or paths with spaces. Acceptable for a first pass but worth documenting the limitation.

4. **`concurrent.futures` imported but unused** (line 10).

5. **`call_tool` timeout** (line 103): Hardcoded 60s. The `ask()` function passes a `timeout` parameter for LLM calls but MCP tool calls get a fixed 60s. Not necessarily wrong but inconsistent.

6. **No test for `McpBridge` itself** — only prompt-building helpers and `FakeBridge` are tested. The real async lifecycle, connect/close, and error paths are untested. This is understandable given the external dependency but should be noted.

### reasons_lib/ask.py:_build_tools_section
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Line 358: `if tool["description"]:` — an empty string `""` is falsy, which is fine, but the tool description from the bridge is always set to `t.description or ""` in `mcp_client.py:79`. So a tool with `description=None` would get `""` here and skip the comment line. That's correct behavior. However, the generated JSON example on line 357 produces malformed JSON when there are zero properties — it would emit `{"tool": "name", }` with a trailing comma from the empty `example`. Actually no — `', '.join([])` produces `""` so it would be `{"tool": "name", }` which is... actually `{"tool": "name", }` — that trailing comma after `"name"` followed by `}` is invalid JSON. The LLM sees this as an example and won't parse it, but it's a misleading prompt.

Wait, looking again: line 357 is `f'{{"tool": "{tool["name"]}", {example}}}'`. If `example` is empty string, this becomes `{"tool": "name", }` — the LLM will see invalid JSON as the tool call format example. Minor but could confuse the model.

### reasons_lib/ask.py:build_ask_prompt
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean extension of the existing function with `tools_section` and `mcp_instructions` parameters, both optional with sensible defaults. The `mcp_instructions` formatting adds a section header only when non-empty. The `tool_label` change in tool_history is consistently applied across both `build_ask_prompt` and `build_final_prompt`.

### reasons_lib/ask.py:ask (MCP integration in tool loop)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

1. **MCP tool results don't update `beliefs_context`** (lines 525-534): When `search_beliefs` returns results, `beliefs_context` is updated so subsequent prompts include better context. MCP tool results are only appended to `tool_history`, meaning the LLM sees them in the "Additional search results" section but the main belief context never improves. This is a design choice but means MCP results get less prominent placement in the prompt than belief search results.

2. **`max_iters = max(MAX_ITERATIONS, 5)`** (line 481): Since `MAX_ITERATIONS = 3`, this always sets `max_iters = 5` when MCP servers are present. The magic number 5 is undocumented. This also means 5 LLM round-trips minimum capacity, which is a cost consideration.

3. **MCP servers not passed to recursive `ask()` call in dual mode** (line 399): When `dual=True`, the recursive call to `ask()` doesn't pass `mcp_servers`, so MCP tools are silently unavailable in dual mode. This could be intentional (dual mode is for sources_db) but is undocumented.

4. **No integration test with mocked `ask()` + `FakeBridge`**: The test file only tests prompt building and tool call extraction — no test verifies the actual tool loop dispatches to an MCP bridge and includes the result in subsequent prompts.

### reasons_lib/ask.py:ASK_PROMPT template
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Line 39 — `{mcp_instructions}` is placed between the rules section and `## Question`. When `mcp_instructions` is non-empty, `build_ask_prompt` wraps it in `\n## Data Source Instructions\n\n...\n`. This means the prompt structure becomes: Rules -> `## Data Source Instructions` -> `## Question`. The blank line before `## Question` (line 40) disappears because `{mcp_instructions}` already ends with `\n`. Minor formatting issue — there's no blank line separating the data source instructions section from the `## Question` heading.

### reasons_lib/cli.py:cmd_ask
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Properly wraps bridge lifecycle in try/finally for cleanup. The `--mcp` argument uses `action="append"` correctly for repeatability. One concern: if any bridge fails to `connect()`, previously connected bridges won't be closed (no cleanup of partial list). But this is a minor edge case. The CLI wiring is clean and follows existing patterns.

### reasons_lib/cli.py:main (argparse)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

`--mcp` argument added cleanly alongside existing `ask` subparser options. `action="append"` with `default=None` is correct — first use creates a list, subsequent uses append. Help text is clear.

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: `mcp = ["mcp>=1.0"]` as an optional dependency is the right pattern — matches how `pg` and `cluster` extras are handled. The `mcp_client.py` gracefully handles the import failure with `_mcp_available` guard.

---

### tests/test_ask.py (modifications)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Tool history dict migration from `{"query": ...}` to `{"tool_label": ...}` is correctly applied in the two affected test methods. All other tests are unchanged and should continue passing since `build_ask_prompt` defaults are backward compatible.

### tests/test_ask_mcp.py (new file)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Tests cover prompt building (`_build_tools_section`, `_build_mcp_instructions`, `build_ask_prompt` with MCP params) and tool call extraction. `FakeBridge` is a good test double. **Missing**: no test for the `ask()` function's MCP tool dispatch loop — the critical path where the LLM emits an MCP tool call, the bridge executes it, and the result feeds back into the next prompt. This is the most important integration path and it's untested.

### Self-Review
**Limitations:** Could not run the test suite to verify all existing tests still pass with the `tool_label` migration. Could not verify MCP SDK API compatibility (whether `ClientSession`, `StdioServerParameters`, `stdio_client` match the `mcp>=1.0` API surface). The observation's "related_test" entries from `.venv` (chromadb, jsonschema, kubernetes, sympy) are false positives from the code review tool and were correctly ignored.

### Feature Requests
- Filter out related_test matches from `.venv/` directories — these are third-party package tests and never relevant to reviewing project changes.
- Include a "run tests" observation that executes the project test suite and reports pass/fail counts, so reviewers don't have to request it manually.

## gemini [BLOCK]

### reasons_lib/mcp_client.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. The `close()` method (lines 115-123) will raise an `AttributeError` if `self._shutdown` has not yet been initialized. This attribute is created only after a successful session initialization in `_session_lifecycle` (line 88). If `connect()` times out or an error occurs during initialization, `cmd_ask` (in `cli.py`) will call `close()` in its `finally` block, causing the CLI to crash with an unhandled exception rather than reporting the original connection error.
2. `self.command.split()` (line 62) will fail for server commands containing spaces in paths or arguments (e.g., `"/path with spaces/mcp-server"`). `shlex.split()` should be used instead.
3. The implementation of `McpBridge` is entirely untested; `tests/test_ask_mcp.py` uses a `FakeBridge` which does not exercise the threading or asyncio bridging logic.

### reasons_lib/ask.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The integration of MCP tools into the `ask` loop is sound. It correctly builds the tools section, injects server-specific instructions, and handles multi-turn tool calling. The migration from `"query"` to `"tool_label"` in `tool_history` is handled consistently across `build_ask_prompt` and `build_final_prompt`. Increasing `max_iters` to 5 for MCP-enabled calls is a sensible heuristic for more complex tool interactions.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `--mcp` argument is correctly implemented as an "append" action, allowing multiple servers. The lifecycle management in `cmd_ask` (connect/close) is correctly placed in a `try...finally` block to ensure resources are cleaned up (though it is currently blocked by the bug in `McpBridge.close`).

### tests/test_ask_mcp.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

These tests provide good coverage for the prompt-building logic and the extraction of tool calls from LLM responses. While they don't test the actual `McpBridge` implementation, they sufficiently verify that the `ask` module correctly handles the abstract bridge interface.

### Self-Review
**Limitations:** I assumed the attribute names used for the MCP SDK objects (e.g., `t.inputSchema`, `result.instructions`) are correct for the `mcp>=1.0` specification. I did not have access to the external `mcp` library's source code to verify these. The lack of integration tests for the actual `McpBridge` means the threading/asyncio interaction is verified only by inspection.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#136
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **tests/test_ask_mcp.py** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Clean addition of `mcp` optional dependency. Keeps MCP as opt-in, consistent with how `pg` and `cluster` extras are structured. No impact on core dependencies.
---

### reasons_lib/mcp_client.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The sync-over-async bridge design is sound, but `_run_loop` (line 59-61) has no exception handling. If `_session_lifecycle` raises before `self._ready.set()` is called — e.g., `stdio_client` fails to connect, `shlex.split` returns an empty list, or the server binary is not found — the exception is silently swallowed by the daemon thread and `connect()` blocks for the full 30s timeout before raising a misleading `TimeoutError`. Fix: wrap `_run_loop` body in try/except, store the exception in `self._error`, and set `self._ready`. Additionally, `close()` calls `self._loop.close()` (line 123) even if the thread didn't join within 5s (line 120), which could close the loop under a still-running thread. No unit tests exist for McpBridge itself — the FakeBridge in tests is a hand-rolled substitute, so the actual threading, lifecycle, and error paths are untested.

### reasons_lib/ask.py:_build_tools_section
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly builds a tools section including both search_beliefs and MCP tools. Handles no-params case without trailing comma (tested). Description truncation to first line is reasonable for prompt economy. Called from `ask()` when `mcp_servers` is set.

### reasons_lib/ask.py:_build_mcp_instructions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple collector. Correctly filters empty instructions via truthiness check. Tests verify both empty and multi-server cases.

### reasons_lib/ask.py:build_ask_prompt
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New `tools_section` and `mcp_instructions` parameters are backward-compatible with defaults (`None` and `""`). The `tool_history` format change from `entry['query']` to `entry['tool_label']` is consistently applied in both `build_ask_prompt` and `build_final_prompt`. MCP instructions are injected under a "Data Source Instructions" heading only when non-empty. Tests in both `test_ask.py` and `test_ask_mcp.py` cover the new parameters.

### reasons_lib/ask.py:ask
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The MCP integration into the tool loop is logically correct — tool name dispatch, argument extraction (`{k: v for k, v in tool_call.items() if k != "tool"}`), error handling, and history tracking all work. Two concerns: (1) MCP tool results append to `tool_history` but don't appear in the main `beliefs_context`, so the LLM only sees them via the "Additional search results" section. This is probably intentional but means MCP results won't be in the final fallback `_beliefs_or_no_match(beliefs_context)` path. (2) The `mcp_servers` parameter is not passed through to the recursive `ask()` call in the `dual` path (line 403-404), so `--dual --mcp` silently drops MCP servers. **Most importantly, there are no integration tests** — no test exercises `ask()` with `mcp_servers=[FakeBridge(...)]` to verify the MCP tool dispatch loop end-to-end. The existing tests only cover prompt building and tool call parsing.

### reasons_lib/cli.py:cmd_ask
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correct lifecycle management with try/finally ensuring bridges are closed. The `--mcp` argument uses `action="append"` correctly for repeatable flags. The lazy import of `McpBridge` only when `--mcp` is used avoids requiring the `mcp` package for non-MCP usage. No CLI-level test exercises the `--mcp` flag, but this is hard to test without a real or mock MCP server.

### tests/test_ask.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Existing tests correctly updated from `{"query": ..., "result": ...}` to `{"tool_label": ..., "result": ...}` format. Two entries in `test_with_tool_history` (lines 90-91) use the new format consistently.

### tests/test_ask_mcp.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

Good coverage of the helper functions (`_build_tools_section`, `_build_mcp_instructions`) and prompt construction. `FakeBridge` is a clean test double. However, there are significant gaps: (1) No test exercises `ask()` with `mcp_servers` to verify the full tool dispatch loop — this is the most important integration path. (2) No test for MCP tool call error handling (the `except Exception as e` path in `ask()`). (3) No test for the `max_iters` bump to 5. (4) `FakeBridge.call_tool` returns JSON but no test uses it through the actual `ask()` function.

### Self-Review
**Limitations:** Could not run the test suite to confirm all tests pass. Could not verify the MCP SDK's actual interface (ClientSession, StdioServerParameters, etc.) matches what's used. Did not verify whether `inputSchema` vs `input_schema` is the correct attribute name on MCP tool objects.

### Feature Requests
- Include the MCP SDK's public API surface (or relevant type stubs) when reviewing MCP client code so attribute names can be verified
- Auto-run the test suite as part of review observations to catch runtime failures

## gemini [PASS]

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: CONSISTENT
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Correctly adds the `mcp` optional dependency and updates the project metadata to support the new feature.
---

### reasons_lib/ask.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The integration of MCP tools into the `ask` loop is well-implemented. It appropriately handles prompt construction with dynamically fetched tool definitions and instructions. The tool-execution loop correctly dispatches to either the internal belief search or external MCP tools. The breaking change in the `tool_history` schema (from `query` to `tool_label`) is consistently applied within the module and its tests.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly implements the `--mcp` CLI argument, allowing multiple servers to be specified. It properly manages the lifecycle of `McpBridge` instances using a `finally` block to ensure connections are closed even on failure.

### reasons_lib/mcp_client.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Provides a solid implementation of a synchronous bridge to the asynchronous MCP SDK. Using a background thread with its own event loop is an appropriate pattern for integrating async IO into a synchronous library. Error handling for connection timeouts and tool execution is included.

### tests/test_ask.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Properly updated to reflect the schema change in `tool_history`, ensuring existing tests remain valid.

### tests/test_ask_mcp.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New unit tests provide good coverage for the prompt building and tool section generation logic introduced for MCP support.

### Self-Review
**Limitations:** I verified that all internal callers of the modified `build_ask_prompt` and `build_final_prompt` functions were updated. However, if these functions are part of a public API used by external projects, the change to the `tool_history` schema would be a breaking change for them.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#136
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## Disagreements

- **tests/test_ask_mcp.py** [LOW]: claude: CONCERN, gemini: PASS

## claude [BLOCK]

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Clean addition of `mcp = ["mcp>=1.0"]` as an optional dependency, consistent with the `pg` and `cluster` extras pattern already in the project. Correctly kept out of core `dependencies`.
---

### reasons_lib/mcp_client.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The bridge architecture is sound — background event loop thread keeping the MCP session alive, `run_coroutine_threadsafe` for cross-thread dispatch, graceful import handling. Three concerns: (1) `close()` line 120 calls `self._thread.join(timeout=5)` but if the thread doesn't finish in 5s, `_thread` is set to None and then `self._loop.close()` is called on a **still-running** loop, which violates asyncio's contract. (2) `call_tool` hardcodes a 60s timeout (line 107) independent of the user's `--timeout` flag — a slow MCP tool can consume most of the budget with no way to control it. (3) If `inputSchema` is explicitly `None` on a tool, `schema.get("properties", {})` at `ask.py:353` would raise `AttributeError` since `None` has no `.get()`. Most importantly: **zero unit tests** for this module (confirmed in observations: `mcp_client_tests.test_count: 0`). The class is only exercised indirectly via `FakeBridge` in `test_ask_mcp.py`.

### reasons_lib/ask.py:_DEFAULT_TOOLS_SECTION
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

**Regression in the default (no-MCP) code path.** `_DEFAULT_TOOLS_SECTION` (line 50-53) uses `{{` and `}}` which are NOT escape sequences in a regular Python string — they produce literal double braces. In the old code, `{{"tool": ...}}` lived inside `ASK_PROMPT` which is a `.format()` template, where `{{` correctly escapes to `{`. Now `_DEFAULT_TOOLS_SECTION` is substituted as a VALUE via `tools_section=...`, so `.format()` does not interpret the `{{` — the LLM sees `{{"tool": "search_beliefs", "query": "search terms"}}` (double braces) instead of valid JSON. This could cause the LLM to emit malformed tool calls that `extract_tool_call` cannot parse. No test catches this because `test_tool_definition_in_prompt` only asserts `"search_beliefs" in prompt` and `'"tool"' in prompt`, which pass regardless of brace count. Fix: change line 53 to single braces `{"tool": "search_beliefs", "query": "search terms"}`.

### reasons_lib/ask.py:_build_tools_section
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly builds the tools section with single braces (line 347 uses `'{"tool": "search_beliefs", ...}'` as a regular string value, not a format template). Handles zero-parameter tools without trailing commas. Truncates descriptions to first line — reasonable prompt-length trade-off. One design note: `required` fields from input_schema aren't surfaced, so the LLM doesn't know which params are mandatory. Tests cover empty servers, multiple servers, params, and no-params edge cases.

### reasons_lib/ask.py:_build_mcp_instructions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple concatenation of non-empty server instructions. Correctly filters empty strings. Tested for both empty and multi-server cases.

### reasons_lib/ask.py:build_ask_prompt
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New parameters `tools_section` and `mcp_instructions` are cleanly integrated. The `tool_label` migration from `entry['query']` to `entry['tool_label']` is consistent across both `build_ask_prompt` and `build_final_prompt`. Conditional wrapping of `mcp_instructions` with `## Data Source Instructions` header only when non-empty avoids prompt clutter. Tests verify all new parameters.

### reasons_lib/ask.py:ask (MCP integration)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The MCP dispatch logic is correct: tool map is built, tool calls are routed, errors are caught and recorded in history. However: (1) `mcp_servers` is **ignored** in the `simple` path (line ~430-460) — `--mcp` with `--simple` silently wastes the connection with no warning. (2) Tool name collision between two MCP servers silently shadows (last bridge wins at line ~483) with no warning. (3) An MCP tool named `"search_beliefs"` would be shadowed by the built-in handler but would still appear in `_build_tools_section` output, confusing the LLM. (4) MCP tool results go into `tool_history` but never update `beliefs_context` — this is likely intentional but means MCP data is only available as "Additional search results", not primary context.

### reasons_lib/cli.py:cmd_ask
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Bridge lifecycle management has a resource leak: the `try/finally` block (line 721-739) only wraps the `ask()` call, not the bridge connection loop (lines 713-719). If the second `--mcp` server's `bridge.connect()` raises `TimeoutError`, the first bridge (already connected and appended to `mcp_servers`) is never closed — the `finally` block doesn't execute because the exception occurs before `try`. The subprocess and daemon thread leak until process exit. Fix: move the `try/finally` to wrap the entire connection + ask block, or use `contextlib.ExitStack`.

### reasons_lib/cli.py:main (--mcp argument)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

`action="append"` with `default=None` is correct for a repeatable flag. Help text accurately describes usage. Placed alongside other `ask` subparser arguments.

### tests/test_ask.py (changes)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly migrates `{"query": ...}` to `{"tool_label": ...}` in the tool history fixtures, matching the production code change. Existing test assertions remain valid.

### tests/test_ask_mcp.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Good coverage of the new MCP integration points: tools section building, instructions collection, prompt construction, tool dispatch, error handling, and iteration limit bumping. `FakeBridge` is a clean test double. Missing coverage: (1) No tests for the actual `McpBridge` class, only the `FakeBridge` substitute. (2) No test for `--mcp` with `--simple` (would reveal the silent ignore). (3) No test for `--mcp` with `--dual`. (4) No test asserting proper JSON formatting (single braces) in the default tools section — would have caught the `_DEFAULT_TOOLS_SECTION` regression. (5) No test for tool name collision between MCP servers.

### Self-Review
**Limitations:** Could not run the test suite to confirm the `_DEFAULT_TOOLS_SECTION` double-brace bug produces visible failures vs. being tolerated by LLMs. Did not have access to the MCP SDK source to verify `inputSchema` nullability guarantees. Could not verify whether `_loop.close()` on a running loop actually causes a crash or just a warning in practice.

### Feature Requests
- Include test run results (pass/fail status) in the observation data to confirm whether flagged issues are caught by existing tests
- Show the git diff of the original format-string template alongside the new extracted constant, to make format-escaping regressions more visible

## gemini [PASS]

### pyproject.toml
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly adds the `mcp` optional dependency to the project metadata.

### reasons_lib/ask.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The integration of MCP tools into the `ask` loop is well-implemented. It handles dynamic tool section generation, instruction injection, and correctly dispatches calls to MCP bridges. The increase in `max_iters` when MCP is active is a logical adjustment for potentially more complex reasoning chains. The safety fallback in the final iteration (returning context if the model still tries to call a tool) is a robust design choice.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Successfully adds the `--mcp` repeatable argument to the `ask` command. The lifecycle management of `McpBridge` (connect/close) in `cmd_ask` using a `try...finally` block is correct and prevents resource leaks.

### reasons_lib/mcp_client.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `McpBridge` provides a clean synchronous interface to the asynchronous MCP SDK. Using a background thread with its own event loop is an effective way to maintain a persistent MCP session across multiple tool calls without requiring the rest of the application to be async. Error handling during connection and initialization is handled properly.

### tests/test_ask_mcp.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new test suite provides good coverage for the MCP-specific logic, including prompt formatting and tool dispatching. The use of `FakeBridge` allows for testing the integration logic without dependencies on external MCP server binaries.

### Self-Review
**Limitations:** I assumed the `mcp` Python library's data models use `inputSchema` as used in `McpBridge`, which is common for MCP implementations following the TypeScript specification. I did not have the full content of `reasons_lib/llm.py`, but its usage in the diff is consistent with established patterns in the codebase.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#136
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Clean addition of `mcp = ["mcp>=1.0"]` as optional dependency, matching the pattern of other optional extras (`pg`, `cluster`). The `mcp_client.py` module guards the import with a try/except, so the dependency is truly optional.
---

### reasons_lib/ask.py:_DEFAULT_TOOLS_SECTION
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Extracting the single-tool prompt text into `_DEFAULT_TOOLS_SECTION` avoids the double-brace escaping problem that existed when it was inline in the `ASK_PROMPT` template. Since `.format()` operates on the template, not on substituted values, the literal JSON braces in the default section are safe. Test `test_default_tools_section` confirms `'{{' not in prompt`.

### reasons_lib/ask.py:build_ask_prompt
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New parameters `tools_section` and `mcp_instructions` are properly defaulted (`None` and `""`) for backward compatibility. The `mcp_instructions` block is only injected when non-empty. The `tool_label` key change in tool_history is consistent across all call sites — `build_ask_prompt`, `build_final_prompt`, and the `ask()` loop all use the same key. Tests in both `test_ask.py` and `test_ask_mcp.py` cover the new parameters.

### reasons_lib/ask.py:_build_tools_section
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Generates the tool section with correct f-string escaping for JSON braces (e.g., `f'{{"tool": "{tool["name"]}", {example}}}'`). The `if example` / `else` split at line 359-360 correctly avoids a trailing comma for parameterless tools — verified by test `test_no_params_no_trailing_comma`. The `tool["description"]` guard at line 362 is safe because `_session_lifecycle` normalizes None descriptions to empty strings.

### reasons_lib/ask.py:_build_mcp_instructions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple collector that filters empty instructions and joins with double newlines. Tested by `TestBuildMcpInstructions`.

### reasons_lib/ask.py:ask (MCP integration)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The MCP dispatch logic is sound for the tool-loop path, but two gaps exist: (1) When `simple=True`, `mcp_servers` is silently ignored — the `simple` branch (lines ~430-460) doesn't use MCP at all, with no warning. A user passing `--simple --mcp snowflake-mcp` would get no MCP integration. (2) When `dual=True`, MCP servers are passed to the recursive `ask()` call for TMS but not to `_fts_rag_answer()`, which is fine, but the docstring says "mcp_servers: list of connected McpBridge instances for external tools" without noting the simple-mode limitation. The MCP tool dispatch itself is correct — args are extracted by filtering `tool != "tool"`, errors are caught and recorded as strings, and `search_beliefs` always takes priority over a same-named MCP tool.

### reasons_lib/mcp_client.py:McpBridge
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The async-to-sync bridge pattern is correct — background event loop thread, `run_coroutine_threadsafe` for cross-thread dispatch, `threading.Event` for synchronization. Two issues: (1) **Startup failure reporting**: If `stdio_client.__aenter__` fails (e.g., command not found), the exception propagates up through `run_until_complete` in `_run_loop`, but `_ready` is never set, so `connect()` hangs for 30 seconds before raising a generic `TimeoutError`. The exception handler at line 90 only covers errors *inside* the `async with` blocks. Fix: wrap `_run_loop`'s body in a try/except that stores the error and sets `_ready`. (2) **`close()` race on timeout**: If `_thread.join(timeout=5)` times out, `_loop.close()` at line 123 is called on a still-running loop, which raises `RuntimeError`. For a CLI tool this is low-severity but could mask errors in the `finally` block of `cmd_ask`. No unit tests exist for this module (confirmed by observations), though the `FakeBridge` integration tests in `test_ask_mcp.py` cover the interface contract.

### reasons_lib/mcp_client.py:_session_lifecycle
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The try/except at line 72 only covers code *after* entering both context managers (`stdio_client` and `ClientSession`). If `session.initialize()` fails, it's properly caught. But failures in `stdio_client.__aenter__` (e.g., subprocess spawn failure) or `ClientSession.__aenter__` are uncaught here and propagate to `_run_loop`, where they're also uncaught, causing the thread to die silently while `connect()` waits 30s for a timeout.

### reasons_lib/cli.py:cmd_ask
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The try/finally pattern correctly ensures all successfully-connected bridges are closed. The lazy import of `McpBridge` is appropriate since it's behind an optional dependency. The `mcp_servers or None` idiom converts an empty list to None, matching the `ask()` API expectation. One minor note: if `bridge.connect()` raises for the Nth bridge, bridges 1..N-1 are still closed by the finally block since they were already appended to `mcp_servers`. No direct test for the CLI `--mcp` argument parsing, but this is covered by the `--mcp` argparse definition at line 1711.

### reasons_lib/cli.py:main (--mcp argument)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

`action="append"` is the correct argparse pattern for repeatable options. `default=None` means `args.mcp` is None when no `--mcp` flags are passed, which the `if args.mcp:` guard in `cmd_ask` handles correctly.

### tests/test_ask.py (tool_label migration)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `query` → `tool_label` key migration in test fixtures is consistent with the production code change. All existing test assertions still pass because they check for substring presence in the rendered prompt, and `tool_label` values like `'search_beliefs("propagation")'` contain the same search terms.

### tests/test_ask_mcp.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test coverage for the new MCP integration: prompt building (4 tests), instruction collection (2 tests), prompt injection (4 tests), tool call extraction (2 tests), and end-to-end dispatch (3 tests including error handling and iteration bumping). The `FakeBridge` class correctly implements the `McpBridge` interface contract. The `test_max_iterations_bumped_with_mcp` test correctly verifies 6 invoke_model calls (5 tool-loop rounds + 1 final synthesis).

### Self-Review
**Limitations:** Could not verify actual MCP SDK types (`ClientSession`, `StdioServerParameters`, tool result shapes) since the SDK isn't included. Relied on observations for full function bodies, which were thorough. Could not run the test suite to confirm all tests pass with the changes.

### Feature Requests
- Include the full contents of newly created files (not just diff) alongside the diff view, since `+` lines can be hard to read for large new files
- Show the schema/interface of external dependencies (like `mcp.ClientSession`) when reviewing code that wraps them
- Flag when a module has zero test coverage automatically in the observation results

## gemini [PASS]

### reasons_lib/mcp_client.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The implementation of the synchronous bridge for the MCP SDK is sound. It correctly uses a background thread and an asyncio event loop to manage the session across multiple tool calls. The use of `shlex.split` for command parsing and `asyncio.run_coroutine_threadsafe` for tool calls follows best practices for mixing sync/async code. While it lacks direct unit tests in this PR, it is effectively tested through integration in `test_ask_mcp.py`.

### reasons_lib/ask.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The modifications to support multi-tool loops and MCP integration are correct. The logic for building tool sections and merging MCP instructions into the prompt is well-implemented and preserves the existing belief search capability. The increase in `max_iters` to 5 when MCP is present is a reasonable adjustment for more complex reasoning tasks.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The CLI changes correctly add the `--mcp` option and manage the lifecycle of the `McpBridge` instances. The use of a `try...finally` block ensures that server connections are closed. Lazy importing of `McpBridge` prevents unnecessary dependency errors when the optional MCP feature is not used.

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Correctly adds `mcp` as an optional dependency, maintaining a lean core.
---

### tests/test_ask_mcp.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Provides comprehensive testing for the prompt generation, tool extraction, and dispatch logic associated with the new MCP feature using a robust mocking strategy.

### Self-Review
**Limitations:** I was unable to verify the exact structure of the `mcp` SDK's `InitializeResult` or `CallToolResult` objects beyond what is inferred from the code, but the usage appears consistent with the standard MCP specification.

### Feature Requests
- Include direct unit tests for `mcp_client.py` using a mock stdio server to verify the background loop and session management more rigorously.

PR #138, opened 2026-05-11
PgApi: Add maintenance operations (hash-sources, check-stale, lookup)

### LINKED ISSUE #67: PgApi: Add maintenance operations
## Summary

Add maintenance operations to PgApi for source tracking and search.

## Functions to add

- `hash_sources(force=False, repos=None)` — backfill source hashes for nodes with source paths but no stored hash
- `check_stale(repos=None)` — check all IN nodes for source file staleness (changed/deleted files)
- `lookup(query, visible_to=None)` — simple all-terms substring search over full belief block (searchable text built from node, source, dates, dependencies)

## Notes

`hash_sources` and `check_stale` require file I/O and repo path resolution. The SQLite API delegates to helper modules for the actual hashing/checking logic.

## Parent issue

Part of #60 (PgApi coverage gaps)

benthomasson:
## Summary

Closes #67. Adds `hash_sources`, `check_stale`, and `lookup` methods to PgApi, enabling these commands with `--pg`:

- **hash-sources**: Backfill source hashes for nodes with source paths but no stored hash. Reuses `resolve_source_path()` and `hash_file()` from `check_stale.py`.
- **check-stale**: Check all IN nodes for source file staleness (deleted, changed, truncated hashes). Supports `--upgrade-hashes`.
- **lookup**: Simple all-terms substring search over full belief blocks (id, text, source, hash, date, dependencies).

Also fixes test env var isolation so `test_pg_dispatch.py` passes when `REASONS_PG_CONNINFO` is set in the environment.

## Test plan

- [x] All 35 `test_pg_dispatch.py` tests pass (9 new)
- [x] Full suite: 921 passed, 166 skipped
- [x] Tests pass with `REASONS_PG_CONNINFO` set in environment
- [ ] Manual test with real PG database

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1381,6 +1381,7 @@ def check_stale(
     repos: dict[str, str] | None = None,
     upgrade_hashes: bool = False,
     db_path: str = DEFAULT_DB,
+    pg_conninfo=None, project_id=None,
 ) -> dict:
     """Check all IN nodes for source file staleness.
 
@@ -1390,6 +1391,9 @@ def check_stale(
     Returns: {"stale": list[dict], "checked": int, "stale_count": int,
               "upgraded": int}
     """
+    if pg_conninfo:
+        return _pg_dispatch(pg_conninfo, project_id, "check_stale",
+                            repos=repos, upgrade_hashes=upgrade_hashes)
     from pathlib import Path as P
     from .check_stale import check_stale as _check
 
@@ -1420,11 +1424,15 @@ def hash_sources(
     force: bool = False,
     repos: dict[str, str] | None = None,
     db_path: str = DEFAULT_DB,
+    pg_conninfo=None, project_id=None,
 ) -> dict:
     """Backfill source hashes for nodes with source paths but no stored hash.
 
     Returns: {"hashed": list[dict], "count": int}
     """
+    if pg_conninfo:
+        return _pg_dispatch(pg_conninfo, project_id, "hash_sources",
+                            force=force, repos=repos)
     from pathlib import Path as P
     from .check_stale import hash_sources as _hash
 
@@ -1464,7 +1472,8 @@ def compact(budget: int = 500, truncate: bool = True, visible_to: list[str] | No
         return _compact(net, budget=budget, truncate=truncate)
 
 
-def lookup(query: str, visible_to: list[str] | None = None, db_path: str = DEFAULT_DB) -> str:
+def lookup(query: str, visible_to: list[str] | None = None, db_path: str = DEFAULT_DB,
+           pg_conninfo=None, project_id=None) -> str:
     """Simple all-terms search over the full belief block — ID, text, source,
     dependencies, and metadata. Matches the same search corpus and output
     format as lookup_beliefs on a flat beliefs.md file.
@@ -1476,6 +1485,9 @@ def lookup(query: str, visible_to: list[str] | None = None, db_path: str = DEFAU
 
     Returns: formatted string with matching beliefs (full blocks)
     """
+    if pg_conninfo:
+        return _pg_dispatch(pg_conninfo, project_id, "lookup",
+                            query=query, visible_to=visible_to)
     with _with_network(db_path) as net:
         query_terms = query.lower().split()
         matches = []
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -597,8 +597,7 @@ def cmd_export_markdown(args):
 
 
 def cmd_hash_sources(args):
-    _require_sqlite(args, "hash-sources")
-    result = api.hash_sources(force=args.force, db_path=args.db)
+    result = api.hash_sources(force=args.force, **_backend_kwargs(args))
 
     if not result["hashed"]:
         print("No nodes to hash (all sources already have hashes, or source files not found).")
@@ -621,8 +620,7 @@ def cmd_hash_sources(args):
 
 
 def cmd_check_stale(args):
-    _require_sqlite(args, "check-stale")
-    result = api.check_stale(upgrade_hashes=args.upgrade_hashes, db_path=args.db)
+    result = api.check_stale(upgrade_hashes=args.upgrade_hashes, **_backend_kwargs(args))
 
     if result.get("upgraded"):
         print(f"Upgraded {result['upgraded']} truncated hash(es) to full length.")
@@ -696,8 +694,7 @@ def cmd_search(args):
 
 
 def cmd_lookup(args):
-    _require_sqlite(args, "lookup")
-    result = api.lookup(args.query, visible_to=_parse_visible_to(args), db_path=args.db)
+    result = api.lookup(args.query, visible_to=_parse_visible_to(args), **_backend_kwargs(args))
     print(result)
 
 
--- a/reasons_lib/pg.py
+++ b/reasons_lib/pg.py
@@ -780,19 +780,22 @@ class PgApi:
                 continue
             new_hash = hash_file(path)
             was_empty = not source_hash
-            with self.conn.cursor() as cur:
-                cur.execute(
-                    "UPDATE rms_nodes SET source_hash = %s "
-                    "WHERE id = %s AND project_id = %s",
-                    (new_hash, nid, pid),
-                )
-            self.conn.commit()
             results.append({
                 "node_id": nid,
                 "source": source,
                 "hash": new_hash,
                 "was_empty": was_empty,
             })
+
+        if results:
+            with self.conn.cursor() as cur:
+                for item in results:
+                    cur.execute(
+                        "UPDATE rms_nodes SET source_hash = %s "
+                        "WHERE id = %s AND project_id = %s",
+                        (item["hash"], item["node_id"], pid),
+                    )
+            self.conn.commit()
 
         return {"hashed": results, "count": len(results)}
 
@@ -818,6 +821,7 @@ class PgApi:
 
         results = []
         upgraded = 0
+        upgrades_to_commit = []
 
         for nid, source, source_hash, meta in sorted(rows, key=lambda r: r[0]):
             if isinstance(meta, str):
@@ -839,13 +843,7 @@ class PgApi:
             if current_hash != source_hash:
                 if len(source_hash) == 16 and current_hash.startswith(source_hash):
                     if upgrade_hashes:
-                        with self.conn.cursor() as cur:
-                            cur.execute(
-                                "UPDATE rms_nodes SET source_hash = %s "
-                                "WHERE id = %s AND project_id = %s",
-                                (current_hash, nid, pid),
-                            )
-                        self.conn.commit()
+                        upgrades_to_commit.append((current_hash, nid))
                         upgraded += 1
                         continue
                     results.append({
@@ -865,6 +863,16 @@ class PgApi:
                     "source_path": str(path),
                     "reason": "content_changed",
                 })
+
+        if upgrades_to_commit:
+            with self.conn.cursor() as cur:
+                for new_hash, nid in upgrades_to_commit:
+                    cur.execute(
+                        "UPDATE rms_nodes SET source_hash = %s "
+                        "WHERE id = %s AND project_id = %s",
+                        (new_hash, nid, pid),
+                    )
+            self.conn.commit()
 
         return {
             "stale": results,
@@ -888,20 +896,23 @@ class PgApi:
             node_rows = cur.fetchall()
 
             cur.execute(
-                "SELECT node_id, antecedents FROM rms_justifications "
+                "SELECT node_id, antecedents, outlist FROM rms_justifications "
                 "WHERE project_id = %s",
                 (pid,),
             )
             just_rows = cur.fetchall()
 
-        antecedents_by_node = {}
+        refs_by_node = {}
         dependents_by_node = {}
-        for node_id, ants in just_rows:
+        for node_id, ants, outs in just_rows:
             if isinstance(ants, str):
                 ants = json.loads(ants)
-            antecedents_by_node.setdefault(node_id, []).extend(ants)
-            for ant in ants:
-                dependents_by_node.setdefault(ant, []).append(node_id)
+            if isinstance(outs, str):
+                outs = json.loads(outs)
+            all_refs = list(ants) + list(outs or [])
+            refs_by_node.setdefault(node_id, []).extend(all_refs)
+            for ref in all_refs:
+                dependents_by_node.setdefault(ref, []).append(node_id)
 
         matches = []
         for nid, text, tv, source, source_hash, date, meta in node_rows:
@@ -917,8 +928,8 @@ class PgApi:
                 block_parts.append(source_hash)
             if date:
                 block_parts.append(date)
-            for ant in antecedents_by_node.get(nid, []):
-                block_parts.append(ant)
+            for ref in refs_by_node.get(nid, []):
+                block_parts.append(ref)
             for dep in dependents_by_node.get(nid, []):
                 block_parts.append(dep)
             block_lower = " ".join(block_parts).lower()
@@ -928,7 +939,7 @@ class PgApi:
                     "id": nid, "text": text, "truth_value": tv,
                     "source": source, "source_hash": source_hash,
                     "date": date,
-                    "antecedents": antecedents_by_node.get(nid, []),
+                    "depends_on": refs_by_node.get(nid, []),
                 })
 
         if not matches:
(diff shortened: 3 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#138
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/api.py:check_stale
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Follows the established dispatch pattern (see `export_markdown` at line 1319). Early return to `_pg_dispatch` with correct kwargs (`repos`, `upgrade_hashes`). SQLite path unchanged.

### reasons_lib/api.py:hash_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same pattern as check_stale. Passes `force` and `repos` correctly through dispatch.

### reasons_lib/api.py:lookup
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Dispatches `query` and `visible_to` correctly. The `db_path` parameter is correctly excluded from the PG dispatch path since PG doesn't use it.

### reasons_lib/cli.py:cmd_hash_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Removed `_require_sqlite` guard and switched from `db_path=args.db` to `**_backend_kwargs(args)`. This matches the pattern used by `cmd_export_markdown` (line 591) and `cmd_search` (line 692). The return dict format (`hashed`, `count`, `was_empty`) is identical between PG and SQLite backends, so CLI formatting code works unchanged.

### reasons_lib/cli.py:cmd_check_stale
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same pattern change as cmd_hash_sources. The `repos` argument isn't passed from CLI — relies on backend-internal repo resolution, which is consistent with how the SQLite path works (falls back to `net.repos`).

### reasons_lib/cli.py:cmd_lookup
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean switch to `**_backend_kwargs(args)`. Both backends return a string, so the CLI `print(result)` works unchanged.

### reasons_lib/pg.py:PgApi.hash_sources
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Two concerns:

1. **Per-row commits in loop** (lines ~790-794): Each node update does its own `self.conn.commit()`. If the process crashes mid-way, some nodes are updated and some aren't. The SQLite version writes atomically via `_with_network(write=True)`. Consider batching updates into a single transaction.

2. **Missing repo fallback**: The SQLite `api.hash_sources` falls back to `net.repos` when `repos` is None (line 1443-1445 in api.py). The PG version doesn't load stored repos. If PgApi doesn't support stored repos, this is fine, but it's a behavioral divergence worth documenting.

3. **No `db_dir` passed to `resolve_source_path`**: The SQLite path passes `db_dir=P(db_path).resolve().parent` which enables resolving sources relative to the database location. The PG path doesn't pass `db_dir`, so sources that rely on db-relative resolution won't be found. This is likely intentional but means the PG backend will silently skip some nodes that SQLite would find.

Only mock-based dispatch tests exist — no integration tests with actual PG queries.

### reasons_lib/pg.py:PgApi.check_stale
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Same concerns as `hash_sources`:

1. **Per-row commit on upgrade** (lines ~843-848): Each hash upgrade commits individually within the loop. Should be a single transaction.

2. **Missing repo fallback**: Same as hash_sources — no fallback to stored repos when `repos` is None.

3. **No `db_dir`**: Same as hash_sources — sources resolving via db-relative paths won't be found.

The staleness detection logic itself (truncated hash handling, source_deleted, content_changed) correctly mirrors `check_stale.py` and produces the same result dict format.

### reasons_lib/pg.py:PgApi.lookup
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Minor behavioral difference in the search corpus. The PG version queries only `antecedents` from `rms_justifications` (line ~898) and builds `dependents_by_node` from those. The SQLite version includes `node.dependents` which tracks nodes referenced via **both** antecedents and outlist. This means a search for a node ID that appears only in another node's outlist will match in SQLite but not in PG.

The PG lookup also doesn't include outlist entries in the per-node `antecedents` list in the search block, whereas the SQLite version iterates `node.justifications` which includes outlist members.

This is unlikely to cause visible issues in practice but is a subtle parity gap. The `metadata` field is also included in the PG output dict but not used in the search block — consistent with the SQLite version.

Only mock-based tests; no integration tests verify the actual SQL queries or result formatting match SQLite behavior.

### tests/test_pg_dispatch.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good additions:

- **Env var isolation fixes** (lines 35-37, 54-56, 84-87): Using `patch.dict("os.environ", env, clear=True)` to prevent `REASONS_PG_CONNINFO` / `REASONS_PROJECT_ID` from leaking from the test environment. This fixes real flake potential — the existing tests were brittle if these env vars happened to be set.

- **New test classes** cover dispatch routing, parameter forwarding, and CLI unblocking for all three operations. Tests verify mock calls with exact kwargs.

- `TestMaintenanceCliNoLongerBlocked` duplicates some assertions from the dispatch tests but serves as a smoke test for the end-to-end unblocking — acceptable redundancy.

The `__import__("os").environ.items()` pattern (lines 35, 55, 84) is unusual but functional. A top-level `import os` already exists implicitly through the test framework.

### Self-Review
**Limitations:** Could not read `tests/test_pg.py` to verify whether integration tests for the three new PgApi methods already exist. Could not verify whether PgApi stores/loads repo mappings anywhere. The `build/` directory has stale copies of source files that still contain `_require_sqlite` guards — unclear if that's intentional or just a stale build artifact (not part of this PR's scope).

### Feature Requests
- Show the full source of classes being modified (e.g., PgApi.__init__ and any repo-related methods) to verify behavioral parity between backends
- Flag stale build artifacts (`build/` copies) that diverge from source when reviewing PRs
- Include integration test files (test_pg.py) when PG implementation methods are added

## gemini [CONCERN]

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly extends `check_stale`, `hash_sources`, and `lookup` to support PostgreSQL dispatch. This follows the existing architectural pattern for backend-agnostic API calls and maintains signature consistency.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Enables the maintenance commands for PostgreSQL by removing the SQLite restriction and correctly passing backend parameters via `_backend_kwargs`.

### reasons_lib/pg.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

- `lookup` (line 876) queries `rms_justifications` but only selects `antecedents`, ignoring the `outlist` column. In the RMS architecture, `outlist` is critical for negative dependencies. Ignoring it results in an incomplete search corpus and incomplete reverse-dependency tracking compared to the SQLite implementation.
- `hash_sources` and `check_stale` (lines 782, 831) perform a database commit for every single node updated. For projects with many source-backed nodes, this will be significantly slower than using a single transaction or batch commits.
- The new `PgApi` methods lack direct integration tests (e.g., in `tests/test_pg.py`); only their dispatch from the API layer is verified in this PR.

### tests/test_pg_dispatch.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Effectively tests the dispatch logic in the API layer and the correct handling of backend-related CLI arguments/environment variables.

### Self-Review
**Limitations:** I assumed the existence of the `outlist` column in the PostgreSQL schema based on its usage in other parts of the project (e.g., `list_nodes`), which confirms the incompleteness of the new `lookup` query. I could not see the top-level imports in `pg.py` to verify if `json` was already imported globally, though it is used elsewhere in the file.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#138
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/api.py:check_stale
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean addition of `pg_conninfo`/`project_id` params with early-return dispatch via `_pg_dispatch`. Follows the exact same pattern used by `export_markdown`, `compact`, and other already-dispatched functions. The SQLite path is unchanged.

### reasons_lib/api.py:hash_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same dispatch pattern as `check_stale`. Kwargs passed through correctly: `force` and `repos`.

### reasons_lib/api.py:lookup
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Dispatch passes `query` and `visible_to` correctly. `db_path` is excluded from PG dispatch as expected since PgApi doesn't use it.

### reasons_lib/cli.py:cmd_hash_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Removed `_require_sqlite` guard and switched from `db_path=args.db` to `**_backend_kwargs(args)`. This matches the pattern used by `cmd_export_markdown` and `cmd_search`. The CLI handler itself doesn't change — it still processes the same `result` dict shape regardless of backend.

### reasons_lib/cli.py:cmd_check_stale
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same pattern change as `cmd_hash_sources`. Note that `repos` is not explicitly passed from the CLI — `check_stale` defaults `repos=None`, and the PgApi implementation handles repo resolution internally. This mirrors the SQLite path where `net.repos` is used as fallback.

### reasons_lib/cli.py:cmd_lookup
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean removal of `_require_sqlite` and switch to `**_backend_kwargs(args)`.

### reasons_lib/pg.py:PgApi.hash_sources
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The implementation does individual UPDATE statements in a loop (`pg.py:794-798`) rather than using batch execution (e.g., `executemany` or `VALUES` list). The recent commit `8da1b39` ("Fix review issues: batch transactions") suggests batching was already flagged as a concern in prior review. For small node counts this is fine, but it's inconsistent with the stated fix. Additionally, unlike the SQLite `hash_sources` in `check_stale.py:125-165`, the PG version does not use `db_dir` for path resolution — it only uses `repos`. This means nodes with sources relative to the DB directory (like `entries/topic.md`) won't resolve in PG mode. This may be intentional since PG deployments might not have a local `db_dir`, but it's a behavioral divergence from the SQLite path that isn't documented or tested.

### reasons_lib/pg.py:PgApi.check_stale
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Same concerns as `hash_sources`: (1) Individual UPDATEs in a loop for hash upgrades (`pg.py:866-871`) rather than batching. (2) No `db_dir` parameter — source path resolution only uses `repos` dict. The SQLite `check_stale` in `check_stale.py:59-122` uses `db_dir` as a fallback resolution path, so there's a feature gap. (3) Tests are all mock-based dispatch tests — no integration test with actual PG data verifying the SQL queries or the stale-detection logic. The logic itself is a faithful port of the SQLite version, so correctness is likely sound, but the missing `db_dir` path means some source paths that work in SQLite won't resolve in PG.

### reasons_lib/pg.py:PgApi.lookup
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The lookup method includes outlist references in the search corpus (line building `refs_by_node` from `outs`), matching the SQLite path which searches `node.dependents`. However, the SQLite `lookup` searches `j.antecedents` for dependency info and `node.dependents` for reverse deps, while the PG version searches both `antecedents` and `outlist` from justifications and also builds a `dependents_by_node` reverse map. This means the PG search corpus is slightly different — it includes outlist node IDs in the search, which the SQLite version doesn't (SQLite searches `j.antecedents` but not `j.outlist` in the block). The SQLite version at `api.py:1505` does `block_parts.extend(j.antecedents)` then `block_parts.append(dep_id)` for dependents — it does NOT include outlist. The PG version at `pg.py:905-908` includes both antecedents and outlist via `all_refs = list(ants) + list(outs or [])`. This is a behavioral difference that could cause divergent search results between backends.

### tests/test_pg_dispatch.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good additions: dispatch tests verify kwargs are passed correctly, CLI-no-longer-blocked tests confirm the `_require_sqlite` removal works. The env var isolation fixes (using `patch.dict` with `clear=True` equivalent) are correct and address real flaky-test issues where ambient `REASONS_PG_CONNINFO` could leak into tests. The `TestMaintenanceCliNoLongerBlocked` class has some overlap with `TestHashSourcesDispatch`/`TestCheckStaleDispatch`/`TestLookupDispatch` — the "accepts_pg" tests are near-duplicates of the "dispatches_to_pg" tests. Not harmful, just redundant.

### Self-Review
**Limitations:** Could not verify the PG SQL queries against an actual schema to confirm column names match. Could not see the full PgApi class to check if there are existing patterns for batch updates. The `_is_visible` helper was observed and confirmed correct. No spec or issue was provided, so compliance assessments are limited.

### Feature Requests
- Show the commit message for the HEAD commit to understand the PR's stated intent
- Include a "behavioral diff" summary when the same logical operation has both SQLite and PG implementations, highlighting divergences in search corpus, query semantics, or resolution paths

## gemini [PASS]

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The functions `check_stale`, `hash_sources`, and `lookup` were correctly updated to include `pg_conninfo` and `project_id` arguments and to dispatch to the PostgreSQL backend using `_pg_dispatch`. This ensures that the public API correctly supports both SQLite and PostgreSQL backends for these operations.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The CLI commands `cmd_hash_sources`, `cmd_check_stale`, and `cmd_lookup` were properly updated to use `**_backend_kwargs(args)`, enabling them to pick up PostgreSQL connection info from CLI flags or environment variables. The removal of `_require_sqlite` for these commands correctly unblocks their usage with the PG backend.

### reasons_lib/pg.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `PgApi` class now implements `hash_sources`, `check_stale`, and `lookup`. 
- `hash_sources` and `check_stale` correctly reuse the logic from `reasons_lib.check_stale` for path resolution and file hashing, ensuring behavioral parity with the SQLite implementation.
- `lookup` correctly implements a project-wide search over full belief blocks, including identifiers, text, sources, and dependencies (antecedents and dependents).
- The use of `json.loads` on metadata and justification fields provides robustness against different database driver behaviors regarding `jsonb` fields.

### tests/test_pg_dispatch.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New tests successfully verify that the API functions correctly dispatch to the `PgApi` methods when appropriate. The updates to `TestBackendKwargs` and `TestRequireSqlite` use improved environment isolation (`patch.dict` with `clear=True`), making the tests more robust.

### Self-Review
**Limitations:** I could not verify the complete `PgApi` class or the full PostgreSQL schema, so I cannot confirm if project-wide repository mappings (`repos`) are stored in the database and should have been loaded by `hash_sources` or `check_stale`. I also noted that `api.py:compact` appears to dispatch to a `PgApi.compact` method which is not present in the current diff for `pg.py`, suggesting potential incompleteness for that specific feature, although it was not the primary focus of this change.

PR #147, opened 2026-05-23
Add metadata section (schema version, repo name, timestamps) across all storage formats

### LINKED ISSUE #146: Add metadata section (schema version, repo name, timestamps) across all storage formats
## Summary

None of the storage formats currently carry metadata about themselves — no schema version, no repo/project name, no creation or last-modified timestamps. This makes it hard to identify, version, or trace the provenance of exported data.

## Current state

| Format | Metadata | Status |
|--------|----------|--------|
| `network.json` | Top-level keys: `nodes`, `nogoods`, `repos` | No metadata |
| `beliefs.md` | Markdown with belief listings | No metadata header |
| `reasons.db` (SQLite) | `network_meta` table exists (key/value) | Empty — unused |
| PostgreSQL (`rms_network_meta`) | Key/value per project | No standard metadata keys |

## Proposed metadata fields

All formats should carry a consistent set of metadata:

| Field | Description | Example |
|-------|-------------|---------|
| `schema_version` | Format version for forward compatibility | `"1.0"` |
| `project_name` | Name of the belief network / project | `"ftl-reasons-expert"` |
| `created_at` | ISO 8601 timestamp of initial creation | `"2026-04-15T10:30:00Z"` |
| `updated_at` | ISO 8601 timestamp of last modification | `"2026-05-23T14:00:00Z"` |
| `node_count` | Total nodes at time of export | `1101` |
| `generator` | Tool and version that produced the file | `"ftl-reasons/0.40.0"` |

### `network.json`
Add a top-level `"meta"` key:
```json
{
  "meta": {
    "schema_version": "1.0",
    "project_name": "ftl-reasons-expert",
    "created_at": "2026-04-15T10:30:00Z",
    "updated_at": "2026-05-23T14:00:00Z",
    "node_count": 1101,
    "generator": "ftl-reasons/0.40.0"
  },
  "nodes": { ... },
  "nogoods": [ ... ],
  "repos": { ... }
}
```

### `beliefs.md`
Add a YAML frontmatter block:
```markdown
---
schema_version: "1.0"
project_name: ftl-reasons-expert
updated_at: "2026-05-23T14:00:00Z"
node_count: 1101
generator: ftl-reasons/0.40.0
---

# Beliefs
...
```

### `reasons.db` (SQLite)
Populate the existing `network_meta` table with standard keys:
```
key              | value
schema_version   | 1.0
project_name     | ftl-reasons-expert
created_at       | 2026-04-15T10:30:00Z
updated_at       | 2026-05-23T14:00:00Z
```
Update `updated_at` on every save. Set `created_at` on `init`.

### PostgreSQL (`rms_network_meta`)
Same keys, scoped per `project_id` as all PgApi metadata already is.

## Design considerations

- `import-json` should read and preserve `meta` if present, ignore gracefully if absent (backward compat)
- `import-beliefs` should parse YAML frontmatter if present, ignore if absent
- `schema_version` enables future format migrations — reader can check version before parsing
- `updated_at` should be set automatically on every `export` / `save`, not manually
- `project_name` could default to the database filename or `--project-id` if not explicitly set

benthomasson:
Now I have a clear picture. Here's the PR description:

---

## Summary

Add a consistent metadata section (schema version, project name, timestamps, node count, generator) across all storage and export formats. Previously, no format carried provenance or versioning information, making it impossible to identify, version, or trace exported data. This centralizes metadata generation in a new `metadata.py` module used by all code paths.

Closes #146

## Changes

- Add `reasons_lib/metadata.py` with `build_meta()` helper and `SCHEMA_VERSION = "1.0"` constant
- **JSON export**: include top-level `"meta"` key with all metadata fields
- **Markdown export**: emit YAML frontmatter block with schema version, project name, timestamps, node count, and generator
- **SQLite storage**: populate `network_meta` table on `init` and update `updated_at` on every save; load metadata back into `network.meta` on read
- **PostgreSQL**: write standard metadata keys to `rms_network_meta` on `init_db`; include `meta` in export; preserve metadata on import
- **JSON import**: read and preserve `meta` section if present, ignore gracefully if absent
- **Markdown import**: parse YAML frontmatter via `strip_frontmatter()`, propagate `schema_version`/`project_name`/`created_at` into the network
- **CLI**: add `--project-name` flag to `init`; default export output to `network.json`/`beliefs.md` files (use `--output=-` for stdout)
- Add `Network.meta` dict attribute for in-memory metadata storage

## Test plan

- [x] `tests/test_metadata.py` — 223 lines covering `build_meta()`, frontmatter stripping, SQLite metadata round-trip, JSON export/import metadata preservation
- [x] `tests/test_metadata_extended.py` — 328 lines covering markdown frontmatter export, CLI `--project-name`, `updated_at` auto-refresh on save, backward compat with metadata-free files
- [x] Existing test suite passes (`tests/test_cli.py` updated for new default output behavior)
- [ ] Manual: `uv run reasons init --project-name test && uv run reasons export` produces JSON with `meta` block
- [ ] Manual: `uv run reasons export-markdown` produces markdown with YAML frontmatter

🤖 Generated with [ftl-sdlc-loop](https://github.com/benthomasson/ftl-sdlc-loop)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#147
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### reasons_lib/metadata.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean single-source-of-truth module. `build_meta` is deterministic given inputs and properly defaults `created_at` to now when not provided. `_get_generator` gracefully handles missing package metadata. `SCHEMA_VERSION` is a simple constant. Tests in `test_metadata.py::TestBuildMeta` and `test_metadata_extended.py::TestBuildMetaEdgeCases` cover the key scenarios.

### reasons_lib/network.py:__init__
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adding `self.meta: dict[str, str] = {}` to `Network.__init__` is minimal and backward-compatible. All consumers that reconstruct a `Network` (storage load, PG path in `export_markdown`, filtered network paths) now correctly propagate `.meta`.

### reasons_lib/storage.py:_init_schema
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `_is_new` flag is set before `sqlite3.connect()` which creates the file. This is correct for the current flow. However, there's a subtle issue: `_is_new` only seeds metadata on first creation. If an **existing** database was created before this change (no metadata rows exist), it won't get seeded. The `save()` method uses `setdefault` which would fill defaults on the next save, so this is self-healing — but the very first `load()` on an old DB will return an empty `network.meta` until a save occurs. Not a bug per se, but worth noting for migration awareness.

### reasons_lib/storage.py:save
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The save method correctly deletes all `network_meta` rows (line 110) and re-inserts everything. `meta.setdefault("created_at", now)` preserves the original `created_at` if it was loaded, and only defaults to now if it was never set. The `INSERT OR REPLACE` is belt-and-suspenders since the table was just cleared, but harmless. `test_metadata_extended.py::TestSqliteMetaExtended::test_created_at_preserved_across_saves` and `test_meta_survives_load_save_cycle` verify this works.

### reasons_lib/storage.py:load
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The refactored load correctly reads all `network_meta` rows in one query, routes `next_nogood_id` to its special field, and populates `network.meta` with everything else. Backward-compatible with old DBs via the `except Exception: pass` block.

### reasons_lib/api.py:init_db
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

When `project_name` is provided, it does a load-modify-save cycle after initial creation. This is slightly redundant since `Storage._init_schema` already seeds `project_name` from the filename stem — but the explicit set correctly overrides that default. Tested by `test_metadata.py::TestSqliteMeta::test_project_name_override`.

### reasons_lib/api.py:export_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly builds `meta` via `build_meta()` using stored values from `net.meta`, and `node_count` is computed from the already-filtered `nodes` dict. The return shape change (adding `"meta"` key) is backward-compatible — all internal callers access `data["nodes"]` or `data["nogoods"]` by key, not by position. Tested extensively.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly reads `meta` from imported JSON and writes only the meaningful fields (schema_version, project_name, created_at) into `net.meta`. Safely handles missing `meta` key via `.get()`. The empty-string check (`if imported_meta[key]`) prevents overwriting existing values with empty strings.

### reasons_lib/api.py:export_markdown
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Both code paths (PG and SQLite) now propagate `net.meta`. The PG path reconstructs `meta` from `data.get("meta")`, and the filtered (visible_to) SQLite path copies `filtered.meta = net.meta`. Both are correct.

### reasons_lib/export_markdown.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Generates YAML frontmatter before the `# Belief Registry` heading. The empty project_name case (`project_name:` with no value) is handled. Schema version is quoted to prevent YAML interpreting `1.0` as a float. Tests verify the frontmatter structure and field values.

### reasons_lib/import_beliefs.py:strip_frontmatter
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple but correct YAML frontmatter parser. Handles missing closing `---`, quoted values (both single and double), and empty values. No YAML library dependency is appropriate here — the frontmatter is known to be simple key-value pairs emitted by our own exporter. Tested by `test_metadata.py::TestStripFrontmatter` and `test_metadata_extended.py::TestStripFrontmatterEdgeCases`.

### reasons_lib/import_beliefs.py:import_into_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Strips frontmatter before parsing beliefs, writes metadata to `network.meta`. Only writes non-empty values, preserving any existing metadata. Round-trip tested in `test_metadata.py::TestMarkdownRoundTrip`.

### reasons_lib/cli.py:cmd_export and cmd_export_markdown
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

**Breaking change to CLI default behavior.** Previously, `reasons export` with no `-o` flag printed to stdout. Now it writes to `network.json` by default. Similarly, `reasons export-markdown` defaults to writing `beliefs.md`. This is a user-facing behavioral change that could surprise existing users or break scripts that pipe stdout (e.g., `reasons export | jq .nodes`). The `--output=-` escape hatch exists but the default shift is significant. The existing tests were updated to use `--output=-`, which masks that the old piping behavior is gone. Additionally, with `nargs="?"` and `const`, there's ambiguity: `reasons export -o` (no value) gives `network.json`, but `reasons export` (no `-o` at all) also gives `network.json` via `default`. This is technically correct argparse, but the UX regression from "stdout by default" to "file by default" should be intentional and documented.

### reasons_lib/pg.py:init_db
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Correctly seeds metadata into `rms_network_meta` using `ON CONFLICT DO NOTHING` (won't overwrite on re-init). Uses `project_name or self.project_id` as a sensible default. The `from .metadata import SCHEMA_VERSION` inside the method body avoids circular imports. PG tests are behind `DATABASE_URL` env, so coverage depends on CI config.

### reasons_lib/pg.py:export_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Reads stored metadata from `rms_network_meta`, filters out `next_nogood_id`, and passes it to `build_meta()`. The `build_meta` call correctly overrides `updated_at` and `node_count` with current values while preserving `created_at` and `project_name`.

### reasons_lib/pg.py:import_json
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Mirrors the SQLite `import_json` logic. Uses `ON CONFLICT DO UPDATE` which will overwrite existing metadata — appropriate for import semantics.

### reasons_lib/pg.py:import_beliefs
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Correctly strips frontmatter before passing to `parse_beliefs`/`parse_nogoods`. Metadata upsert is consistent with `import_json` path.

### tests/test_pg_dispatch.py (NOT in diff but affected)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

`EXPORT_DATA` fixture at line 104 does not include a `"meta"` key. The `export_markdown` PG path in `api.py` now reads `data.get("meta")` and sets `net.meta = dict(meta)` when truthy — so missing `meta` is handled safely (just gets `None`, skips). However, `test_empty_network` at line 202 uses `{"nodes": {}, "nogoods": [], "repos": {}}` without `meta`. This is fine because the code uses `.get("meta")` with no default, but it means the PG dispatch tests don't verify that metadata actually flows through the markdown export path. Not a blocker, but a coverage gap.

### tests/test_metadata.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good coverage of the core metadata flows: build_meta, JSON export/import, markdown export/import, SQLite storage, and round-trips. Backward compatibility tested with JSON-without-meta and markdown-without-frontmatter.

### tests/test_metadata_extended.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Thorough edge-case coverage. One minor concern: `test_save_updates_updated_at` (line 166) asserts `created_at` is unchanged but doesn't actually assert `updated_at` changed — the test name suggests it should verify `updated_at` changes between saves, but it only checks `created_at` stability. The test is technically valid but misleadingly named. Also, duplicate test coverage with `test_metadata.py` on some scenarios (e.g., both files test `build_meta` basics and JSON round-trips), which is harmless but adds maintenance burden.

### tests/test_cli.py (modified tests)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Tests correctly updated to use `--output=-` to get stdout behavior. The modifications are consistent with the CLI default change.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Could not verify the `rms_network_meta` table schema in `pg.py` (the `SCHEMA` constant observation failed). The PG integration tests are behind `DATABASE_URL` so their pass/fail status is unknown. Did not see the full `test_pg.py` to check if PG export tests verify the new `meta` key in the return value.

### Feature Requests
- Include the schema/DDL constants (e.g., `SCHEMA` in `pg.py`) in observations when database schema changes are part of the diff
- Flag test fixtures that mock return values of modified functions but haven't been updated to match the new return shape
- Show whether CI is passing for the branch being reviewed

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#147
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **reasons_lib/api.py:init_db** [LOW]: claude: CONCERN, gemini: PASS
- **reasons_lib/api.py:export_network** [LOW]: claude: PASS, gemini: CONCERN

## claude [CONCERN]

### reasons_lib/metadata.py (new file)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean single-source-of-truth module for metadata fields. `build_meta` is deterministic (given wall clock), uses `setdefault` pattern correctly for `created_at` preservation. `_get_generator` handles missing package metadata gracefully. `SCHEMA_VERSION` as a module-level constant is clean. Called from `api.py`, `storage.py`, `export_markdown.py`, and `pg.py` — well-integrated.

### reasons_lib/network.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Single-line addition of `self.meta: dict[str, str] = {}` to `Network.__init__`. Properly initialized as empty dict. All producers (storage.load, import_into_network) and consumers (storage.save, export_network, export_markdown) reference this field correctly.

### reasons_lib/storage.py:_init_schema
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`_is_new` flag set in `__init__` before `sqlite3.connect` — since `connect` creates the file, `_is_new` correctly captures whether this is first-time creation. Uses `INSERT OR IGNORE` so re-running on an existing DB won't clobber metadata. The `project_name` defaults to `self.db_path.stem`, which is a reasonable convention.

### reasons_lib/storage.py:save
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `DELETE FROM network_meta` + re-insert pattern is consistent with the existing delete-and-rewrite strategy for all other tables (nodes, nogoods, repos, etc.). `meta.setdefault("created_at", now)` correctly preserves existing `created_at` while always updating `updated_at`. `str(val)` coercion handles the `node_count` int that could leak in from `build_meta` if someone stores it — though `network.meta` shouldn't contain `node_count` (that's only in export output). Minor: `build_meta` is imported but not called here — `save` builds its own meta dict from `network.meta` + defaults. This is correct since `build_meta` includes transient fields like `node_count` and `generator` that shouldn't be persisted.

### reasons_lib/storage.py:load
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Changed from single-key lookup to loading all `network_meta` rows into `network.meta`, correctly filtering out `next_nogood_id` (which is handled separately as an internal counter). The `except Exception: pass` guard for old DBs without the `network_meta` table is preserved.

### reasons_lib/api.py:init_db
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

When `project_name` is provided, the code does an extra load/save cycle to overwrite the name. This works but overwrites the `created_at` timestamp written moments earlier by `_init_schema` with a new `now` value from `save()`. In practice the difference is milliseconds, but it's conceptually sloppy — the creation timestamp shifts. More importantly, when `project_name` is empty (the common case), no load/save happens, so the `updated_at` from `_init_schema` and the `created_at` are identical — good. The concern is minor.

### reasons_lib/api.py:export_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`build_meta` is called with `net.meta` values, `node_count` computed from the filtered `nodes` dict (correctly reflects visibility filtering), and the return dict now includes `"meta"`. The return type change from `{"nodes", "nogoods"}` to `{"meta", "nodes", "nogoods", "repos"}` is additive — all existing callers that access `result["nodes"]` or `result["nogoods"]` continue to work. Verified by checking the observation data for all callers.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly imports only the durable metadata keys (`schema_version`, `project_name`, `created_at`) from the JSON `meta` dict, ignoring transient fields like `updated_at`, `node_count`, `generator`. The guard `if imported_meta and isinstance(imported_meta, dict)` handles backward-compat with old exports that lack `meta`. The key whitelist prevents arbitrary metadata injection.

### reasons_lib/api.py:export_markdown
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Both the pg_conninfo path (reconstruct from export_network data) and the local DB path correctly propagate `net.meta` through to the `_export` function. The `filtered.meta = net.meta` line ensures visibility-filtered exports still include metadata. The `meta = data.get("meta"); if meta: net.meta = dict(meta)` pattern in the pg path handles missing meta gracefully.

### reasons_lib/export_markdown.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The YAML frontmatter is hand-constructed rather than using a YAML library. This works for simple values but has an edge case: if `project_name` contains YAML-special characters (colons, quotes, `#`, etc.), the output will be malformed YAML. The `strip_frontmatter` parser on the import side does a simple `partition(":")` which would handle colons in values, but a YAML-aware consumer could choke. For the expected use case (project names like "my-project") this is fine, but it's a latent bug. Also, `updated_at` is quoted but `project_name` is unquoted — inconsistent quoting strategy.

### reasons_lib/import_beliefs.py:strip_frontmatter
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple and correct frontmatter parser. Handles: no frontmatter, missing closing `---`, empty values, quoted values (both single and double). The `partition(":")` approach correctly handles values containing colons (only splits on the first). Used in both `import_into_network` and `pg.import_beliefs`. Edge case: a line without a colon inside the frontmatter block is silently skipped — acceptable behavior.

### reasons_lib/import_beliefs.py:import_into_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`strip_frontmatter` is called first, stripping frontmatter before the rest of the function processes the body text. Metadata keys are selectively imported (same whitelist as `import_json`). The empty-string guard (`if frontmatter[key]`) prevents overwriting existing meta with empty values from the frontmatter.

### reasons_lib/pg.py:init_db
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Uses `ON CONFLICT DO NOTHING` so re-initializing won't overwrite existing metadata — correct for idempotent init. `project_name` defaults to `self.project_id` (UUID) when not provided — less friendly than the SQLite default of `db_path.stem`, but acceptable for multi-tenant PG where project_id is the primary identifier. PG tests require a running database and are likely skipped in CI without `DATABASE_URL`.

### reasons_lib/pg.py:export_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Fetches stored metadata from `rms_network_meta`, filters out `next_nogood_id`, and passes through `build_meta` to add transient fields. Consistent with the SQLite path. The separate cursor `cur2` avoids conflict with the main query cursor.

### reasons_lib/pg.py:import_json / import_beliefs
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Both methods mirror the SQLite import logic: whitelist-based metadata import with `ON CONFLICT DO UPDATE`. `import_beliefs` correctly calls `strip_frontmatter` and commits metadata before processing beliefs. The `commit()` after metadata insert is explicit rather than relying on the later commit — safe, though slightly redundant.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

`--project-name` argument added to `init` subparser. `getattr(args, "project_name", "") or ""` is defensive — handles both missing attribute and `None`. The CLI is the primary user-facing entry point for setting project_name at init time.

### tests/test_metadata.py (new file)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test coverage across all storage formats: `build_meta` unit tests, JSON export/import round-trips, markdown export frontmatter, `strip_frontmatter` parsing, markdown round-trips, backward compatibility (no-meta JSON/markdown), and SQLite metadata verification. Tests are well-structured by feature area.

### tests/test_metadata_extended.py (new file)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good edge case coverage. However, there's significant overlap with `test_metadata.py` — e.g., `TestBuildMetaEdgeCases.test_created_at_not_overwritten_when_provided` duplicates `TestBuildMeta.test_preserves_created_at` in spirit. `import time` on line 14 is unused. `test_save_updates_updated_at` (line 192) asserts `created_at` is preserved but doesn't actually assert `updated_at` changed — the test name is misleading (though proving `updated_at` changed is hard without sleeping, so the test is still useful for what it does check).

### tests/test_pg_dispatch.py (not modified but affected)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The `EXPORT_DATA` fixture at line 105 lacks the `"meta"` key. The `export_markdown` function handles this via `data.get("meta")` returning `None`, so tests won't crash. However, the mock data no longer matches the real `export_network` return shape. The `test_empty_network` test (line 202) also lacks `"meta"`. These tests should be updated to include mock metadata for better coverage of the pg→markdown path with metadata.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Could not see the full PG schema (`SCHEMA` constant in `pg.py`) to verify `rms_network_meta` table definition. Did not check whether `conftest.py`'s `pg_api` fixture calls `init_db` with `project_name` param. Could not verify backward compatibility with real old database files.

### Feature Requests
- Include constants/schema definitions referenced by modified code (e.g., the `SCHEMA` SQL string in `pg.py`)
- Flag test fixtures in other test files that mock the modified function's return value (like `EXPORT_DATA` in `test_pg_dispatch.py`) — these are high-risk for silent breakage
- Include unused import detection in the diff analysis

## gemini [CONCERN]

### reasons_lib/metadata.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Provides a centralized, consistent way to generate metadata. Correctly utilizes `importlib.metadata` for version tracking and handles timestamps appropriately.

### reasons_lib/network.py:Network.__init__
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Minimal and correct addition of the `meta` attribute to the core `Network` class.

### reasons_lib/api.py:init_db
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly integrates `project_name` into the database initialization flow. For SQLite, it performs an initial save with the custom name, which is functional and safe for new databases.

### reasons_lib/api.py:export_network
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

While the docstring is updated to include `repos` in the return dictionary, the implementation for the SQLite path (lines 885-919) fails to actually include the `repos` key. This creates an inconsistency with the PostgreSQL implementation in `pg.py` and causes repository mapping data to be lost during JSON round-trips on SQLite-backed networks.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly restores metadata and repositories from the imported JSON structure.

### reasons_lib/api.py:export_markdown
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Ensures metadata is correctly propagated to the markdown generator for both local and PostgreSQL backends.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly exposes the new `--project-name` option to the user via the `init` command.

### reasons_lib/export_markdown.py:export_markdown
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly implements YAML frontmatter generation. Handles the case of empty project names gracefully to ensure valid YAML structure.

### reasons_lib/import_beliefs.py:strip_frontmatter
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The simple YAML parser is appropriate for the limited scope of metadata fields. It correctly handles quoting and separates frontmatter from the markdown body.

### reasons_lib/import_beliefs.py:import_into_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Successfully wires the frontmatter extraction into the belief import pipeline.

### reasons_lib/pg.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

PostgreSQL implementation is thorough. `export_network` (lines 1295-1314) correctly includes `repos`, unlike the SQLite counterpart in `api.py`. `import_json` and `import_beliefs` handle metadata updates using `ON CONFLICT` correctly.

### reasons_lib/storage.py:Storage
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly handles persistence of metadata in the `network_meta` table. The `load` method (lines 238-251) maintains backward compatibility for older databases and correctly restores the `meta` dictionary.

### Self-Review
**Limitations:** The inconsistency in `api.export_network` regarding the `repos` key was identified by comparing implementations across backends. A specific test for repository round-tripping via JSON would have caught this automatically.

### Feature Requests
- Implement the missing `repos` key in the SQLite path of `api.export_network` to match the PostgreSQL implementation and the updated docstring.
- Add a test case specifically for repository preservation during JSON export/import cycles.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#147
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/metadata.py (new file)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean single-source-of-truth module. `build_meta` produces consistent metadata dicts. `_get_generator` gracefully handles missing package metadata. `SCHEMA_VERSION` is a clear constant imported by both storage backends. The `created_at or now` pattern correctly preserves existing timestamps. Tests in `TestBuildMeta` and `TestBuildMetaEdgeCases` cover the key behaviors.

### reasons_lib/network.py:Network.__init__
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adding `self.meta: dict[str, str] = {}` is the right approach — an empty dict that gets populated by storage on load and by import on ingest. All consumers (storage, api, export_markdown) read/write through this field consistently. Type annotation `dict[str, str]` is accurate since all meta values are stored as strings.

### reasons_lib/storage.py:Storage.__init__
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `_is_new` flag (checked before `sqlite3.connect` creates the file) correctly gates initial metadata seeding in `_init_schema`. The `project_name` parameter flows through `api.init_db` → `Storage(db_path, project_name=project_name)`. Only the `init_db` path in `api.py` passes `project_name`; the `_with_network` context manager does not, which is correct — existing DBs already have their metadata.

### reasons_lib/storage.py:_init_schema
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `_is_new` check at line 85 (`if self._is_new`) seeds metadata only for new databases. This is correct for fresh DBs. However, there's a subtle race: `_is_new` is set at line 72 (`not self.db_path.exists()`), but `sqlite3.connect` at line 74 creates the file. If `Storage.__init__` is called concurrently for the same new path, both instances could see `_is_new=True`. This is unlikely in practice (SQLite handles concurrent writes), but worth noting. The `INSERT OR IGNORE` mitigates data corruption from this case.

### reasons_lib/storage.py:save
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Line 111 does `DELETE FROM network_meta`, then line 146-148 inserts `next_nogood_id`, then lines 151-161 insert all meta keys via `INSERT OR REPLACE`. The `DELETE` + re-insert is fine within the transaction. However, `meta.setdefault("created_at", now)` at line 155 means that if `network.meta` doesn't have `created_at` (e.g., a Network created in memory without loading), it gets the current time — not the original creation time. This is technically correct for a fresh Network, but if someone calls `save()` on a network that was constructed in memory (not loaded from DB), the `created_at` will be "now" rather than preserving any earlier value. The `load()` method does populate `network.meta` from DB, so the normal load→modify→save cycle preserves `created_at` correctly. Tests confirm this (`test_created_at_preserved_across_saves`, `test_save_preserves_created_at`).

### reasons_lib/storage.py:load
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The refactored `network_meta` loading at lines 239-248 now reads all keys into `network.meta` (except `next_nogood_id` which goes to `_next_nogood_id`). This is clean and backward-compatible — the `try/except` still handles old DBs without the `network_meta` table. The filter `if key == "next_nogood_id"` vs `else` is clear and correct.

### reasons_lib/api.py:init_db
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `project_name` parameter is threaded through to `Storage(db_path, project_name=project_name)` and to `_pg_dispatch`. The CLI wires it via `--project-name` argument. Backward compatible — defaults to `""`.

### reasons_lib/api.py:export_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `meta` dict is built via `build_meta` using values from `net.meta`, then included in the return dict. The `node_count` is computed from the filtered `nodes` dict (correct — it reflects what's actually exported). The return type change from `{"nodes", "nogoods"}` to `{"meta", "nodes", "nogoods", "repos"}` is a breaking change to callers, but all callers in the codebase access `data["nodes"]` or `data.get("nodes")`, so they're unaffected by the addition of `meta`. The `repos` key was already present in the PG path but not in the SQLite path — wait, looking more carefully, the SQLite `export_network` doesn't include `repos` in its return. Let me check.

### reasons_lib/api.py:export_network (repos key missing in SQLite path)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The docstring says the return includes `"repos": dict`, and the PG path (`pg.py:export_network`) does return `repos`. But the SQLite path (lines 889-918 in the diff) returns `{"meta": meta, "nodes": nodes, "nogoods": [...]}` — no `"repos"` key. The `export_markdown` function at line 1398 does `repos=net.repos` directly, so it works. But for JSON export symmetry, the SQLite path is missing `"repos"` in its return dict. This predates this PR (the old code also didn't return `repos`), but the updated docstring now claims it does. The PG path returns it; the SQLite path doesn't.

### reasons_lib/api.py:import_json (meta import)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Line 1164-1166: the meta import uses `if key in imported_meta` without checking if the value is truthy, unlike the `import_into_network` function which checks `if key in frontmatter and frontmatter[key]`. This means an empty string value like `"project_name": ""` will overwrite the destination's project_name. This inconsistency between JSON import and markdown import paths could lead to surprising behavior, though it's arguably correct for JSON (explicit empty string means "clear it").

### reasons_lib/api.py:export_markdown
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Both the PG path and the SQLite path now propagate `net.meta` to the `_export` (export_markdown) function via `network.meta`. The filtered-network path at line 1407 copies `filtered.meta = net.meta`, ensuring metadata survives visibility filtering. Clean integration.

### reasons_lib/export_markdown.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The frontmatter generation is clean. The conditional at line 26 (`if meta["project_name"]` → quoted value, else `project_name:`) correctly handles both cases. The `created_at` field from `build_meta` is not emitted in the frontmatter — this is intentional (only `updated_at` is shown), though it means frontmatter doesn't round-trip `created_at` unless it was in `network.meta`. Wait — actually looking at `strip_frontmatter`, it only imports `schema_version`, `project_name`, and `created_at`. Since `created_at` isn't in the frontmatter output, it won't round-trip. This is fine — `created_at` is a DB-level concern, and `updated_at` is the appropriate field for the export artifact.

### reasons_lib/import_beliefs.py:strip_frontmatter
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple, correct frontmatter parser. Handles missing closing `---`, empty values, quoted values. The `val.strip().strip('"').strip("'")` chain at line 42 correctly strips outer quotes. Edge case: a value like `'"quoted"'` would strip to `quoted` — acceptable for this use case. Tests cover no-frontmatter, with-frontmatter, quoted values, incomplete frontmatter.

### reasons_lib/import_beliefs.py:import_into_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The frontmatter-to-meta propagation is clean — only imports `schema_version`, `project_name`, `created_at`, and only if they have non-empty values. This prevents overwriting with empty strings. Called before the main parse loop, so the body text (without frontmatter) is what gets parsed for beliefs.

### reasons_lib/pg.py:init_db
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Seeds `rms_network_meta` with the same four keys as SQLite. Uses `ON CONFLICT DO NOTHING` so re-init doesn't overwrite. The `project_name` defaults to `self.project_id` (UUID) which is a reasonable fallback for PG projects. Test coverage is via the PG integration tests (requires `DATABASE_URL`).

### reasons_lib/pg.py:export_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Reads stored metadata from `rms_network_meta`, filters out `next_nogood_id`, then builds the `meta` dict via `build_meta`. The import of `build_meta` inside the function is fine for avoiding circular imports.

### reasons_lib/pg.py:import_json / import_beliefs
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Both methods mirror their SQLite counterparts. `import_json` inserts/updates meta keys with `ON CONFLICT DO UPDATE`. `import_beliefs` uses `strip_frontmatter` and stores frontmatter values. Consistent with the SQLite path.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

`--project-name` arg added to `init` subcommand. The `getattr(args, "project_name", "") or ""` at line 22 is defensive but correct — ensures empty string rather than None. Clean CLI integration.

### tests/test_metadata.py (new file)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive tests covering: `build_meta` basics, JSON export/import meta, markdown export frontmatter, `strip_frontmatter`, markdown round-trip, SQLite metadata seeding and preservation. Good backward-compat tests (JSON without meta, old beliefs.md without frontmatter).

### tests/test_metadata_extended.py (new file)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Significant overlap with `test_metadata.py` — many tests are near-duplicates (e.g., `test_full_roundtrip_preserves_meta` vs `test_import_preserves_meta`, `test_created_at_preserved_across_saves` vs `test_save_preserves_created_at`, `test_default_project_name_from_filename` vs `test_export_default_project_name`). This duplication increases maintenance burden. Consider consolidating into a single test file.

### tests/test_pg_dispatch.py (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`EXPORT_DATA` fixture now includes `"meta"` key — necessary because `export_markdown` in the PG path now expects it. The `test_empty_network` fixture also updated. Correct and minimal changes.

### tests/test_export_markdown.py (existing, not modified)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The existing `test_export_markdown.py` was NOT updated, but it should still pass — its assertions use `in md` checks (e.g., `assert "# Belief Registry" in md`) which will match even with the new frontmatter prepended. However, there are no tests in this file for frontmatter behavior — that coverage lives in `test_metadata.py` instead, which is a reasonable separation.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Could not see the full `api.py` export_network function to confirm whether `repos` is in the SQLite return dict (only saw the diff hunks + observation excerpts). Did not see the PG schema definition (`SCHEMA` constant in `pg.py`) to verify `rms_network_meta` table structure exists. Could not verify whether existing tests in `test_import_json.py` or `test_api.py` break due to the changed return shape of `export_network`.

### Feature Requests
- Include the full source of constants (like `SCHEMA` SQL) referenced in modified functions, not just function bodies
- Flag when a function's return type/shape changes and show all callers that destructure or assert on the return value
- When duplicate test coverage is detected across test files, quantify the overlap (percentage of equivalent assertions)

## gemini [PASS]

### .gitignore
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: CONSISTENT
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Added `.sdlc-loop/` to ignored files. Simple cleanup.

---

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Updated API functions to handle project metadata. `init_db` now accepts `project_name`. `export_network` includes a `meta` field in the return dictionary. `import_json` and `export_markdown` correctly handle metadata and frontmatter. Logic is sound and integration with other components (Storage, PG, Network) is correct.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Added the `--project-name` argument to the `init` command and updated `cmd_init` to pass it to the API. Correctly wired up.

### reasons_lib/export_markdown.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Added YAML frontmatter to markdown export containing metadata (schema version, project name, timestamps, node count, generator). Uses `build_meta` for consistency across formats.

### reasons_lib/import_beliefs.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Added `strip_frontmatter` helper to parse YAML frontmatter from markdown. Updated `import_into_network` to extract and apply metadata to the network object. The parser is simple but sufficient for the supported fields.

### reasons_lib/metadata.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New utility module providing a consistent metadata structure and generator strings across all formats. Good separation of concerns and facilitates reuse between SQLite and PostgreSQL paths.

### reasons_lib/network.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Added `meta` dictionary to the `Network` class to store project-level metadata in memory.

### reasons_lib/pg.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Updated PostgreSQL backend to store and retrieve metadata from the `rms_network_meta` table. Implementation mirrors the SQLite logic and correctly handles `ON CONFLICT` and `UPDATE` scenarios for metadata preservation during imports.

### reasons_lib/storage.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Updated SQLite backend to handle project-level metadata. Correctly initializes metadata for new databases and preserves `created_at` during subsequent saves. `load()` includes backward compatibility for databases created before the metadata table was introduced.

### tests/test_metadata.py & tests/test_metadata_extended.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test suite covering all new metadata functionality across storage backends and export formats. Includes round-trip tests, edge cases (e.g., numeric values in frontmatter), and backward compatibility checks for legacy files without metadata.

### Self-Review
**Limitations:** I did not see the content of the `SCHEMA` constant in `reasons_lib/storage.py` or `reasons_lib/pg.py` to verify the exact table definition of `network_meta` and `rms_network_meta`, but the code's usage (insert/select/update) implies they are correctly defined in the schema script. The diff did not include updates to `test_pg.py`, but `test_pg_dispatch.py` was updated and serves as the primary integration test for the PG backend's new data structure.

### Feature Requests
- Include the content of constants like `SCHEMA` when they are referenced in modified code, as they are critical for verifying database migrations and schema consistency.

PR #153, opened 2026-05-25
Add import-api and export-api for agentic-mind-service sync

### LINKED ISSUE #152: Add import-api and export-api commands for agentic-mind-service sync
## Summary

Add `reasons import-api` and `reasons export-api` commands to sync beliefs between local `reasons.db` and the remote agentic-mind-service at `api.agentic-mind.ai`.

## Motivation

Agents provisioned via ftl2-agentcraft now have access to a remote memory service (agentic-mind-service) that persists beliefs, entries, and checkpoints across VM reprovisioning. Currently agents can push individual beliefs to the API via curl, but there's no way to:

1. Pull remote beliefs back into the local TMS (reasons.db) to get full explain/retract/nogood support
2. Bulk-push all local beliefs to the remote service

This creates a gap: beliefs saved remotely can't be used locally with the full TMS features (justification chains, retraction cascades, contradiction detection).

### `reasons import-api`

Pull beliefs from the remote service into local reasons.db.

```bash
reasons import-api --url $MIND_SERVICE_URL --agent-id $MIND_AGENT_ID --api-key $MIND_API_KEY
```

- Fetches all beliefs from `GET /api/agents/{agent_id}/beliefs`
- Imports them into local reasons.db (skip duplicates by node_id)
- Preserves truth_value, justifications where available

### `reasons export-api`

Push all local beliefs to the remote service.

```bash
reasons export-api --url $MIND_SERVICE_URL --agent-id $MIND_AGENT_ID --api-key $MIND_API_KEY
```

- Reads all nodes from local reasons.db
- POSTs each to `POST /api/agents/{agent_id}/beliefs`
- Idempotent — the service already handles duplicate node_ids

## Environment Variables

Both commands should read from environment variables by default (set automatically on agent VMs):
- `MIND_SERVICE_URL` — e.g. `https://api.agentic-mind.ai`
- `MIND_AGENT_ID` — UUID, e.g. `c45e0d40-8d12-41ed-afba-ba22d0a5a577`
- `MIND_API_KEY` — Bearer token string

## API Reference

Source code: https://github.com/benthomasson/agentic-mind-service

All endpoints require `Authorization: Bearer <MIND_API_KEY>` header.

### `POST /api/agents/{agent_id}/beliefs` — Add a belief

**Request body:**
```json
{
  "node_id": "string (required — unique belief ID, e.g. 'sky-blue')",
  "text": "string (required — belief content)",
  "sl": "string (optional — comma-separated SL justification antecedents, e.g. 'a,b')",
  "cp": "string (optional — comma-separated CP justification antecedents)",
  "unless": "string (optional — comma-separated outlist for non-monotonic justification)",
  "label": "string (optional)",
  "source": "string (optional — source attribution, e.g. file path)"
}
```

**Response (200):**
```json
{
  "node_id": "sky-blue",
  "truth_value": "IN",
  "type": "premise",
  "premise_count": 0
}
```

### `GET /api/agents/{agent_id}/beliefs` — List beliefs

**Query params:** `?status=IN` (optional, filter by truth value)

**Response (200):**
```json
{
  "nodes": [
    {
      "id": "sky-blue",
      "text": "The sky is blue",
      "truth_value": "IN",
      "justification_count": 0,
      "dependent_count": 0,
      "challenges": [],
      "last_reviewed": null,
      "review_result": null
    }
  ],
  "count": 1
}
```

### `GET /api/agents/{agent_id}/beliefs/{node_id}` — Get belief details

**Response (200):**
```json
{
  "id": "sky-blue",
  "text": "The sky is blue",
  "truth_value": "IN",
  "justifications": [],
  "dependents": [],
  "challenges": [],
  "source": "observation",
  "label": "",
  "created_at": "2026-05-25T11:50:04",
  "last_reviewed": null,
  "review_result": null
}
```

### `DELETE /api/agents/{agent_id}/beliefs/{node_id}` — Retract belief

**Query params:** `?reason=...` (optional retraction reason)

**Response (200):**
```json
{
  "node_id": "sky-blue",
  "truth_value": "OUT",
  "retracted": true
}
```

### `GET /api/agents/{agent_id}/export` — Export full network

Returns the complete belief network (nodes + justifications) as JSON. Useful for bulk import.

### `POST /api/agents/{agent_id}/nogoods` — Record contradiction

**Request body:**
```json
{
  "node_ids": ["node-a", "node-b"]
}
```

## Implementation Notes

- The service wraps `reasons_lib.api` (the same library used by the `reasons` CLI) — so the data model is identical
- Each agent gets its own SQLite reasons.db on the server at `data/{agent_id}/reasons.db`
- For `import-api`: use `GET /beliefs` to list all nodes, then call `reasons_lib.api.add_node()` locally for each. Nodes with SL justifications should be imported after their antecedents
- For `export-api`: use `reasons_lib.api.list_nodes()` locally, then `POST /beliefs` for each. The endpoint is idempotent on node_id
- Auth: all requests need header `Authorization: Bearer <key>`

benthomasson:
## Summary

- Adds `reasons import-api` — downloads full belief network from agentic-mind-service via `GET /api/agents/{agent_id}/export` and imports into local `reasons.db`
- Adds `reasons export-api` — pushes each local belief to agentic-mind-service via `POST /api/agents/{agent_id}/beliefs`
- New `reasons_lib/mind_service.py` HTTP client module (stdlib `urllib`, no new deps) with config resolution from CLI flags or `MIND_SERVICE_URL` / `MIND_AGENT_ID` / `MIND_API_KEY` env vars

## Test plan

- [x] 22 new tests in `tests/test_mind_service.py` (all mocked HTTP, no real network calls)
- [x] Full test suite passes (1066 passed, 166 skipped)
- [ ] Manual smoke test with live agentic-mind-service credentials

Closes #152

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -8,6 +8,7 @@ All functions return dicts suitable for JSON serialization.
 """
 
 import json
+import logging
 import re
 import sqlite3
 import sys
@@ -20,6 +21,8 @@ from .metadata import build_meta
 from .network import Network
 from .storage import Storage
 
+
+logger = logging.getLogger(__name__)
 
 DEFAULT_DB = "reasons.db"
 
@@ -1399,7 +1402,8 @@ def export_api(url: str | None = None, agent_id: str | None = None,
                 source=ndata.get("source", ""),
             )
             exported += 1
-        except Exception:
+        except Exception as exc:
+            logger.warning("Failed to push node %s: %s", nid, exc)
             errors += 1
 
     return {"nodes_exported": exported, "errors": errors}
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1338,7 +1338,6 @@ def import_api(url: str | None = None, agent_id: str | None = None,
 
     Returns: {"nodes_imported": int, "nogoods_imported": int}
     """
-    import json as json_mod
     import tempfile
 
     from .mind_service import _resolve_config, fetch_export
@@ -1349,7 +1348,7 @@ def import_api(url: str | None = None, agent_id: str | None = None,
 
     db_exists = Path(db_path).exists()
     if not db_exists and (init or not pg_conninfo):
-        data = json_mod.loads(json_str)
+        data = json.loads(json_str)
         project_name = data.get("meta", {}).get("project_name", "")
         init_db(db_path=db_path, force=False,
                 project_name=project_name,
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1325,6 +1325,86 @@ def import_hf(repo_id: str, init: bool = False, token: str | None = None,
     return result
 
 
+def import_api(url: str | None = None, agent_id: str | None = None,
+               api_key: str | None = None, init: bool = False,
+               db_path: str = DEFAULT_DB,
+               pg_conninfo=None, project_id=None) -> dict:
+    """Import beliefs from agentic-mind-service.
+
+    Downloads the full network via GET /export and imports it locally.
+
+    Returns: {"nodes_imported": int, "nogoods_imported": int}
+    """
+    import json as json_mod
+    import tempfile
+
+    from .mind_service import _resolve_config, fetch_export
+
+    resolved_url, resolved_agent_id, resolved_api_key = _resolve_config(
+        url, agent_id, api_key)
+    json_str = fetch_export(resolved_url, resolved_agent_id, resolved_api_key)
+
+    db_exists = Path(db_path).exists()
+    if not db_exists and (init or not pg_conninfo):
+        data = json_mod.loads(json_str)
+        project_name = data.get("meta", {}).get("project_name", "")
+        init_db(db_path=db_path, force=False,
+                project_name=project_name,
+                pg_conninfo=pg_conninfo, project_id=project_id)
+
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".json", delete=False) as f:
+        f.write(json_str)
+        temp_path = f.name
+
+    try:
+        return import_json(temp_path, db_path=db_path,
+                           pg_conninfo=pg_conninfo, project_id=project_id)
+    finally:
+        Path(temp_path).unlink()
+
+
+def export_api(url: str | None = None, agent_id: str | None = None,
+               api_key: str | None = None,
+               db_path: str = DEFAULT_DB,
+               pg_conninfo=None, project_id=None) -> dict:
+    """Export local beliefs to agentic-mind-service.
+
+    Pushes each node via POST /beliefs. Idempotent on node_id.
+
+    Returns: {"nodes_exported": int, "errors": int}
+    """
+    from .mind_service import _resolve_config, push_belief
+
+    resolved_url, resolved_agent_id, resolved_api_key = _resolve_config(
+        url, agent_id, api_key)
+
+    data = export_network(db_path=db_path, pg_conninfo=pg_conninfo,
+                          project_id=project_id)
+
+    exported = 0
+    errors = 0
+    for nid, ndata in data.get("nodes", {}).items():
+        sl_parts = []
+        for j in ndata.get("justifications", []):
+            if j.get("type") == "SL" and j.get("antecedents"):
+                sl_parts.extend(j["antecedents"])
+        sl = ",".join(sl_parts) if sl_parts else ""
+
+        try:
+            push_belief(
+                resolved_url, resolved_agent_id, resolved_api_key,
+                node_id=nid,
+                text=ndata.get("text", ""),
+                sl=sl,
+                source=ndata.get("source", ""),
+            )
+            exported += 1
+        except Exception:
+            errors += 1
+
+    return {"nodes_exported": exported, "errors": errors}
+
+
 def derive_prompt(domain: str | None = None, db_path: str = DEFAULT_DB) -> dict:
     """Build a derive prompt from the current network.
 
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -596,6 +596,41 @@ def cmd_import_hf(args):
         print(f"Imported {result['nogoods_imported']} nogoods")
 
 
+def cmd_import_api(args):
+    try:
+        result = api.import_api(
+            url=args.url,
+            agent_id=args.agent_id,
+            api_key=args.api_key,
+            init=args.init,
+            **_backend_kwargs(args),
+        )
+    except (RuntimeError, FileExistsError) as e:
+        print(f"Error: {e}", file=sys.stderr)
+        sys.exit(1)
+
+    print(f"Imported {result['nodes_imported']} nodes")
+    if result['nogoods_imported']:
+        print(f"Imported {result['nogoods_imported']} nogoods")
+
+
+def cmd_export_api(args):
+    try:
+        result = api.export_api(
+            url=args.url,
+            agent_id=args.agent_id,
+            api_key=args.api_key,
+            **_backend_kwargs(args),
+        )
+    except RuntimeError as e:
+        print(f"Error: {e}", file=sys.stderr)
+        sys.exit(1)
+
+    print(f"Exported {result['nodes_exported']} nodes")
+    if result['errors']:
+        print(f"Errors: {result['errors']}")
+
+
 def cmd_export(args):
     data = api.export_network(visible_to=_parse_visible_to(args), **_backend_kwargs(args))
     output = json.dumps(data, indent=2)
@@ -1675,6 +1710,20 @@ def main():
                    help="Initialize reasons.db if it doesn't exist")
     p.add_argument("--token", help="HuggingFace token (default: from HF_TOKEN or ~/.cache/huggingface/token)")
 
+    # import-api
+    p = sub.add_parser("import-api", help="Import beliefs from agentic-mind-service")
+    p.add_argument("--url", help="Service URL (default: MIND_SERVICE_URL env)")
+    p.add_argument("--agent-id", help="Agent UUID (default: MIND_AGENT_ID env)")
+    p.add_argument("--api-key", help="API key (default: MIND_API_KEY env)")
+    p.add_argument("--init", action="store_true",
+                   help="Initialize reasons.db if it doesn't exist")
+
+    # export-api
+    p = sub.add_parser("export-api", help="Export beliefs to agentic-mind-service")
+    p.add_argument("--url", help="Service URL (default: MIND_SERVICE_URL env)")
+    p.add_argument("--agent-id", help="Agent UUID (default: MIND_AGENT_ID env)")
+    p.add_argument("--api-key", help="API key (default: MIND_API_KEY env)")
+
     # export
     p = sub.add_parser("export", help="Export network as JSON")
     p.add_argument("-o", "--output", default="network.json", nargs="?", const="network.json",
@@ -1881,6 +1930,8 @@ def main():
         "import-beliefs": cmd_import_beliefs,
         "import-json": cmd_import_json,
         "import-hf": cmd_import_hf,
+        "import-api": cmd_import_api,
+        "export-api": cmd_export_api,
         "export": cmd_export,
         "export-markdown": cmd_export_markdown,
         "export-card": cmd_export_card,
--- a/reasons_lib/mind_service.py
+++ b/reasons_lib/mind_service.py
@@ -1,0 +1,81 @@
+"""HTTP client for agentic-mind-service belief sync."""
+
+import json
+import os
+from urllib.error import HTTPError
+from urllib.request import Request, urlopen
+
+
+def _resolve_config(
+    url: str | None = None,
+    agent_id: str | None = None,
+    api_key: str | None = None,
+) -> tuple[str, str, str]:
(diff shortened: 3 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#153
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### reasons_lib/mind_service.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean HTTP client using stdlib `urllib`. Config resolution is solid — explicit params override env vars, missing values produce a clear error listing all missing fields. `fetch_export` and `push_belief` handle 401/404 with readable `RuntimeError` messages and re-raise other `HTTPError`s. Trailing slash stripping on URL prevents double-slash in endpoint construction. No external dependencies added (stdlib only), consistent with `dependencies = []` in pyproject.toml. One minor note: the 60s and 30s timeouts on `urlopen` are hardcoded but reasonable defaults.

### reasons_lib/api.py:import_api
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Follows the exact same pattern as `import_hf` — fetch remote JSON string, auto-init DB if missing, write to temp file, delegate to `import_json`, clean up in `finally`. The temp-file-then-`import_json` approach avoids duplicating the topological import logic. One difference from `import_hf`: the `init` guard uses `init or not pg_conninfo` which auto-inits SQLite DBs even without `--init` when no pg backend is configured — this matches `import_hf` behavior exactly. The `finally` block ensures the temp file is cleaned up even on import failure.

### reasons_lib/api.py:export_api
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Two concerns:

1. **Lossy justification export**: The SL flattening logic collects antecedents from all SL justifications into a single comma-joined string. If a node has multiple SL justifications (e.g., `SL(a,b)` and `SL(c,d)`), they get merged into `"a,b,c,d"` — the server receives one justification instead of two. Whether this matters depends on the server's API contract, but the local model distinguishes them.

2. **Outlist dropped**: `export_api` only sends `antecedents` from SL justifications but ignores `outlist` entirely. Non-monotonic beliefs (`--unless`) will lose their outlist during export, making round-trips lossy. The `push_belief` API only accepts `sl` (antecedents string) with no outlist field, so this is a design limitation rather than a bug, but it should be documented or flagged.

3. **Silent exception swallowing**: The `except Exception` catch-all increments an error counter but discards the exception entirely. If 50 out of 100 nodes fail, the user sees `Errors: 50` with no diagnostic info. At minimum, logging the first error or the node ID that failed would help debugging.

No test covers the error-counting path (e.g., mock `urlopen` to raise on specific nodes).

### reasons_lib/cli.py (cmd_import_api, cmd_export_api, argparse wiring)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

CLI handlers follow the established pattern from `cmd_import_hf`. Both subcommands are registered in `main()` and wired into the dispatch dict. The `--pg` and `--project-id` args are inherited via `_backend_kwargs(args)`. One note: `cmd_import_api` catches `FileExistsError` (matching `import_hf`), which is correct since `init_db` raises it. `cmd_export_api` only catches `RuntimeError`, which covers the `_resolve_config` failure path. No CLI-level integration tests, but this is consistent with the rest of the CLI — CLI handlers are thin wrappers over `api.*`.

### tests/test_mind_service.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Good coverage of the happy paths and config resolution edge cases. The mock pattern for `urlopen` is clean. Gaps:

1. **No test for `export_api` error path**: The `except Exception` branch in `export_api` that increments `errors` is untested. A test where `urlopen` raises for one node but succeeds for another would verify the partial-failure behavior.
2. **No test for non-monotonic/outlist export**: No test verifies what happens when exporting nodes with outlist justifications — the lossy behavior should be documented via a test even if intentional.
3. **`_mock_response` returns the same response object for all calls**: In `test_export_pushes_all_nodes`, the same mock response is reused across multiple `urlopen` calls. This works because `read()` is called once per context manager entry, but the mock's `read` always returns the same bytes. This is fine for counting calls but wouldn't catch response-parsing bugs if the test evolved.
4. **Unused import**: `call` is imported from `unittest.mock` but never used.

### Self-Review
**Limitations:** Could not verify the agentic-mind-service API contract (what fields POST /beliefs expects, whether `sl` as a comma-string is the correct format, whether outlist support exists). The review assumes the client matches the server's expectations based on field naming. Could not see if `export_network` includes `source_url` in its output dict — the observation shows it does, and `push_belief` doesn't send it, which may be another field lost on export.

### Feature Requests
- Show the server API spec or OpenAPI schema when reviewing client code that calls external services
- Flag unused imports automatically in test files included in the diff

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#153
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## Disagreements

- **tests/test_mind_service.py** [HIGH]: claude: PASS, gemini: BLOCK
- **reasons_lib/api.py:export_api** [LOW]: claude: PASS, gemini: CONCERN

## claude [CONCERN]

### reasons_lib/mind_service.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean HTTP client using only stdlib (`urllib`), consistent with the project having zero required dependencies. Config resolution logic is sound — explicit params override env vars, missing values produce a clear error. Error handling maps 401/404 to user-friendly `RuntimeError` messages and re-raises other `HTTPError`s. The `_resolve_config` function is exported with a leading underscore — it's internal but imported by `api.py` and tests. This is fine given the module is internal to the package. The 60s/30s timeouts for fetch/push are reasonable defaults.

### reasons_lib/api.py:import_api
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Two concerns:

1. **Temp file not cleaned on `import_json` exception**: The `finally` block does call `unlink()`, so this is actually handled. Scratch that — this is correct.

2. **Auto-init logic is subtly wrong for fresh installs without `--init`**: When `db_path` doesn't exist and `init=False` and `pg_conninfo` is None, the condition `not db_exists and (init or not pg_conninfo)` evaluates to `True`, so the DB gets auto-created even without `--init`. This differs from `import_hf` which passes `init` through and lets `init_db` raise `FileExistsError` if the file already exists. The `--init` flag becomes meaningless for SQLite — the DB is always created. If this is intentional (convenience for import-api), the `--init` CLI flag is misleading since it suggests the DB won't be created without it. Compare with `cmd_import_hf` which requires `--init` to auto-create.

3. **Redundant `import json as json_mod`**: The module-level `import json` already exists (line 10). The local alias `json_mod` is copied from `import_json` where it made sense to avoid shadowing the `json_file` parameter, but here there's no name conflict.

### reasons_lib/api.py:export_api
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Iterates all exported nodes, extracts SL justification antecedents, and pushes each via `push_belief`. The `logger.warning` for failed pushes is appropriate — the previous commit message mentions this was a review fix. The SL extraction logic correctly handles multiple justifications by extending `sl_parts` across all SL-type justifications, then joining with commas. One minor note: outlists are silently dropped during export (the server API apparently doesn't support them), which means a round-trip through the service loses non-monotonic justification information. This is acceptable given the service schema but worth noting.

### reasons_lib/cli.py:cmd_import_api
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Follows the same pattern as `cmd_import_hf`. Exception handling catches `RuntimeError` and `FileExistsError`, matching the exceptions that `import_api` and its callees can raise. Properly wired into the argument parser and dispatch table. No CLI-level tests, but the API-level tests cover the core logic adequately.

### reasons_lib/cli.py:cmd_export_api
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Follows the established CLI pattern. Only catches `RuntimeError` (not `FileExistsError`), which is correct since export doesn't create databases. Properly wired into parser and dispatch. The `--pg` and `--project-id` args are inherited via `_backend_kwargs`.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Both subparsers are registered with appropriate `--url`, `--agent-id`, `--api-key`, and `--init` (import only) arguments. Help text references env var fallbacks. The dispatch dict maps both commands correctly.

### tests/test_mind_service.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test suite covering: config resolution (6 tests), fetch_export (5 tests including error paths), push_belief (5 tests), import_api (3 tests), export_api (4 tests including error counting and SL justification verification). The mock pattern using `_mock_response` with context manager protocol is clean. Tests verify URL construction, auth headers, HTTP method, JSON body contents, and error mapping. The `test_export_counts_errors` test uses `side_effect` list to simulate one success + one failure, which correctly validates the partial-failure accounting. One observation: `TestExportApi.test_export_sends_sl_justifications` verifies the derived node's body contains the antecedent IDs in `sl`, which is a good integration check.

### Overall Assessment
**Verdict:** CONCERN

The implementation is well-structured, follows existing patterns, uses only stdlib (no new dependencies), and has thorough test coverage. The one substantive concern is the auto-init behavior in `import_api` — the DB is always auto-created for SQLite even without `--init`, making the flag a no-op. This is inconsistent with `import_hf` and could surprise users who expect `--init` to be required. If intentional, the `--init` flag should be removed from `import-api`; if not, the condition should be `not db_exists and init`.

### Self-Review
**Limitations:** Could not verify what the `agentic-mind-service` API actually expects — reviewing the client against an unseen server API. Could not verify whether `export_network` output format is stable or if other callers exist that might be affected. Did not have access to the full `api.py` file to check for other patterns around auto-init.

### Feature Requests
- Show the server API spec or OpenAPI schema (if available) alongside client code to verify request/response contracts
- Include the full diff of `cli.py`'s argument parser section to verify `--pg`/`--project-id`/`--db` args are inherited correctly by the new subparsers (they may be missing if they're not global args)

## gemini [BLOCK]

### reasons_lib/mind_service.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The HTTP client is cleanly implemented using only standard library dependencies (`urllib`). It handles service configuration resolution correctly and includes appropriate error handling for authentication (401) and missing resources (404).

### reasons_lib/api.py:import_api
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The function checks for the existence of `db_path` (defaulting to `reasons.db`) to decide whether to call `init_db`. If a user is targeting a PostgreSQL backend (`pg_conninfo` is set) but a local `reasons.db` file happens to exist, `init_db` will be skipped for the PostgreSQL project even if `--init` was requested. This is a leak of the SQLite abstraction into the general logic. Additionally, the internal `import json as json_mod` is redundant as `json` is already imported at the module level.

### reasons_lib/api.py:export_api
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The function flattens all `SL` justifications for a node into a single comma-separated `sl` string. In a TMS, multiple justifications represent an "OR" relationship, but combining their antecedents into one list effectively treats them as one large "AND" relationship (or at least loses the disjunctive structure), which is lossy. Furthermore, it completely ignores the `outlist` portion of Doyle-style SL justifications, meaning non-monotonic dependencies will be lost during export.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The CLI wiring correctly integrates the new commands and follows the project's established patterns for backend dispatch and error handling.

### tests/test_mind_service.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

This file contains a critical syntax error: every test case uses the decorator `@tests/test_pg_dispatch.py(...)`. This is not valid Python and appears to be a botched search-replace that should have used `@patch`. The tests will fail to load or run in any environment.

### Self-Review
**Limitations:** I could not verify if the `agentic-mind-service` API actually supports multiple justifications or if the `sl` field is intended to be a simple dependency list. However, based on the Doyle TMS model used elsewhere in the project, the current export logic is clearly lossy.

PR #163, opened 2026-05-29
Add reasons repair-smuggled command

### LINKED ISSUE #155: Automate search-and-link repair for smuggled premises
## Problem

When review-beliefs flags a derived belief as invalid due to smuggled premises, the current workflow is retraction. But data from bare-metal-expert shows that 59% of smuggled premises are recoverable -- the facts exist in the belief network or source entries but were not properly cited as antecedents.

Retract-only workflows destroy value. The search-and-link repair pattern recovers beliefs by finding the proper antecedent and wiring it up.

## Evidence

From bare-metal-expert (611 beliefs, 39 invalid):
- 17 smuggled premises identified
- 10 recovered via search-and-link (59%)
- 7 had existing premises that just needed to be wired up
- 2 required creating new premises from source entries
- 1 was already justified transitively
- Only 5 required softening (no source support)
- 2 were already OUT from cascades

The same four failure categories replicate across domains (redhat-expert, bare-metal-expert), confirming these are properties of LLM derivation, not domain-specific.

## Proposed Implementation

After review-beliefs identifies a smuggled premise (introduces X which appears in no antecedent):

1. Extract the smuggled claim from the review output
2. Search the belief network via reasons search
3. If a matching premise exists: add-justification with the found premise as antecedent
4. If no premise but source entry exists: add new premise from source, then add justification
5. If no source support: soften the derived belief text to remove the unsupported claim

Steps 1-3 are fully automatable. Step 4 requires source entry search. Step 5 requires LLM judgment.

## Possible CLI

reasons repair-smuggled (automatic review then search-and-link)
reasons review-beliefs --repair (as a flag on existing command)

## Automation Profile

| Strategy | Automatable | Notes |
|---|---|---|
| Search and link | High | Pattern match: identify smuggled text, search beliefs, add justification |
| Soften and ground | Medium | Requires judgment about how much to weaken |
| Retract (abandon) | Trivial | Already automated via reasons retract |

## References

- beliefs-pi/entries/2026/05/28/review-categories-replicate-in-bare-metal-expert.md
- beliefs-pi/entries/2026/05/28/search-and-link-repair-executed-on-bare-metal-expert.md
- beliefs-pi/entries/2026/05/25/derive-review-research-pattern-recurs-in-citation-fidelity.md

benthomasson:
## Summary

- Adds `reasons repair-smuggled` command to automate search-and-link repair for smuggled premises (issue #155)
- Uses two LLM calls per invalid belief: extract the smuggled claim, then match against existing premises via search
- Wires matched premises as new antecedents via `add_justification` (appends alternative support path)
- Accepts either a `--review-file` JSON report or belief IDs to re-review inline

## Test plan

- [x] 29 new tests in `tests/test_repair.py` covering parse functions, mocked LLM calls, orchestration flows, and CLI integration
- [x] Full test suite passes (1105 passed, 166 skipped)
- [ ] Manual test with real beliefs: `reasons review-beliefs --sample 5 -o review.json && reasons repair-smuggled --review-file review.json --dry-run`

Closes #155

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2399,6 +2399,70 @@ def review_beliefs(
     }
 
 
+def repair_smuggled(
+    review_file: str | None = None,
+    belief_ids: list[str] | None = None,
+    model: str = "claude",
+    timeout: int = 300,
+    dry_run: bool = False,
+    db_path: str = DEFAULT_DB,
+) -> dict:
+    """Repair smuggled premises by searching for and linking existing premises.
+
+    Two input modes:
+        review_file: path to a review-beliefs JSON report — extract invalid results
+        belief_ids: re-review these beliefs inline, then repair invalids
+
+    Returns dict with repairs list and summary counts.
+    """
+    from .repair import repair_smuggled_beliefs
+
+    if review_file:
+        import json as _json
+        with open(review_file) as f:
+            report = _json.load(f)
+        review_results = report.get("results", [])
+    elif belief_ids:
+        result = review_beliefs(
+            belief_ids=belief_ids, model=model, timeout=timeout,
+            dry_run=True, db_path=db_path,
+        )
+        review_results = result["results"]
+    else:
+        raise ValueError("Provide either review_file or belief_ids")
+
+    invalid = [r for r in review_results if not r.get("valid", True)]
+
+    if not invalid:
+        return {
+            "repairs": [],
+            "total_invalid": 0,
+            "repaired": 0,
+            "no_candidates": 0,
+            "no_match": 0,
+            "extraction_failed": 0,
+            "errors": 0,
+        }
+
+    net = export_network(db_path=db_path)
+    nodes = net.get("nodes", {})
+
+    repairs = repair_smuggled_beliefs(
+        invalid, nodes, model=model, timeout=timeout,
+        db_path=db_path, dry_run=dry_run,
+    )
+
+    return {
+        "repairs": repairs,
+        "total_invalid": len(invalid),
+        "repaired": sum(1 for r in repairs if r["status"] == "repaired"),
+        "no_candidates": sum(1 for r in repairs if r["status"] == "no_candidates"),
+        "no_match": sum(1 for r in repairs if r["status"] == "no_match"),
+        "extraction_failed": sum(1 for r in repairs if r["status"] == "extraction_failed"),
+        "errors": sum(1 for r in repairs if r["status"] == "error"),
+    }
+
+
 def detect_contradictions(
     belief_ids: list[str] | None = None,
     model: str = "claude",
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -1424,6 +1424,48 @@ def cmd_review_beliefs(args):
                 print(f"  ERROR retracting {r['id']}: {e}", file=sys.stderr)
 
 
+def cmd_repair_smuggled(args):
+    _require_sqlite(args, "repair-smuggled")
+
+    model = getattr(args, "model", None) or "claude"
+    review_file = getattr(args, "review_file", None)
+    belief_ids = args.ids if args.ids else None
+
+    if not review_file and not belief_ids:
+        print("Error: provide either --review-file or belief IDs", file=sys.stderr)
+        sys.exit(1)
+
+    result = api.repair_smuggled(
+        review_file=review_file,
+        belief_ids=belief_ids,
+        model=model,
+        timeout=args.timeout,
+        dry_run=args.dry_run,
+        db_path=args.db,
+    )
+
+    repairs = result["repairs"]
+    for r in repairs:
+        status = r["status"].upper()
+        print(f"  [{status}] {r['id']}")
+        if r.get("smuggled_claim"):
+            print(f"    Smuggled: {r['smuggled_claim']}")
+        if r.get("matched_premises"):
+            print(f"    Linked: {', '.join(r['matched_premises'])}")
+        if r.get("rationale"):
+            print(f"    Rationale: {r['rationale']}")
+        if r.get("error"):
+            print(f"    Error: {r['error']}")
+
+    print(f"\nTotal invalid: {result['total_invalid']}")
+    print(f"  Repaired: {result['repaired']}  No candidates: {result['no_candidates']}"
+          f"  No match: {result['no_match']}  Extract failed: {result['extraction_failed']}"
+          f"  Errors: {result['errors']}")
+
+    if args.dry_run:
+        print("\n  (dry run -- no changes applied)")
+
+
 def cmd_contradictions(args):
     _require_sqlite(args, "detect-contradictions")
 
@@ -1880,6 +1922,19 @@ def main():
     p.add_argument("--no-report", action="store_true",
                    help="Skip JSON report generation")
 
+    # repair-smuggled
+    p = sub.add_parser("repair-smuggled",
+        help="Repair smuggled premises by finding and linking existing premises")
+    p.add_argument("ids", nargs="*", help="Belief IDs to review and repair")
+    p.add_argument("--review-file", default=None,
+                   help="Path to review-beliefs JSON report")
+    p.add_argument("-m", "--model", default=None,
+                   help="Model to use (default: claude)")
+    p.add_argument("--timeout", type=int, default=300,
+                   help="LLM timeout in seconds (default: 300)")
+    p.add_argument("--dry-run", action="store_true",
+                   help="Report findings without applying repairs")
+
     # contradictions
     p = sub.add_parser("contradictions", help="Detect contradictions between IN beliefs")
     p.add_argument("ids", nargs="*", help="Specific belief IDs to check (default: all IN)")
@@ -1973,6 +2028,7 @@ def main():
         "list-gated": cmd_list_gated,
         "list-negative": cmd_list_negative,
         "review-beliefs": cmd_review_beliefs,
+        "repair-smuggled": cmd_repair_smuggled,
         "contradictions": cmd_contradictions,
         "namespaces": cmd_namespaces,
     }
--- a/reasons_lib/repair.py
+++ b/reasons_lib/repair.py
@@ -1,0 +1,260 @@
+"""Repair smuggled premises by search-and-link.
+
+Given beliefs flagged invalid by review-beliefs (smuggled premise —
+conclusion introduces facts not present in any antecedent), this module:
+1. Extracts the specific smuggled claim via LLM
+2. Searches the belief network for matching premises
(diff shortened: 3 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#163
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### reasons_lib/repair.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The core repair logic is sound — extract smuggled claim via LLM, search for matching premises, match via LLM, wire via `add_justification`. Two issues:

1. **Outlist dropped on repair** (lines 232-238): The new justification copies antecedents from `first_just` but does NOT copy the outlist. If the original justification had `outlist=[C]` ("believe X unless C"), the repaired justification has `outlist=[]`, removing the non-monotonic constraint. This could cause a repaired belief to remain IN even when the original outlist condition says it shouldn't. Fix: pass `unless=",".join(first_just.get("outlist", []))` to `add_justification`.

2. **Only first justification considered** (line 232): `first_just = node.get("justifications", [{}])[0]` takes only the first justification's antecedents for the repair. If the review flagged the belief as invalid and a different justification was the problematic one, the wrong antecedents are used. This is acceptable as a pragmatic choice but worth documenting.

### reasons_lib/repair.py:parse_extract_response
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly strips whitespace, removes "Smuggled claim:" prefix (case-insensitive check), and strips enclosing quotes. All branches are tested.

### reasons_lib/repair.py:parse_match_response
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Uses `json.JSONDecoder().raw_decode()` to find a JSON object anywhere in the response — robust against LLMs that add prose around JSON. Correctly filters `matched_ids` against `valid_ids` set, preventing hallucinated IDs from being used. Returns safe defaults on malformed input.

### reasons_lib/repair.py:repair_smuggled_beliefs
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The candidate filtering logic (lines 196-212) is well-designed — excludes self, existing antecedents, OUT nodes, and derived beliefs. The `search_fn` injection makes testing clean. However: (1) outlist is dropped as noted above, (2) no test covers a node with multiple justifications, (3) no test covers a node with an outlist, (4) the `except Exception` catch-all at line 240 swallows all errors silently into the result dict — while this prevents one bad belief from stopping the whole batch, it also hides bugs during development.

### reasons_lib/api.py:repair_smuggled
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Clean API wrapper with two input modes (review_file and belief_ids). The `review_file` mode loads JSON, the `belief_ids` mode re-reviews inline with `dry_run=True` (correctly avoiding side effects from the re-review). Summary counts are computed correctly. No direct unit test for this function, but it's exercised through CLI tests.

### reasons_lib/cli.py:cmd_repair_smuggled
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

CLI handler follows the same pattern as other commands. Correctly validates input (requires `review_file` or `belief_ids`). Output formatting is clear. Parser is properly wired in the dispatch dict. The `--timeout` default of 300 matches the API default.

### tests/test_repair.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good coverage across the main scenarios (full repair, dry run, no candidates, no match, extraction failure, filtering existing antecedents, filtering derived beliefs, missing nodes, valid beliefs skipped). Two issues:

1. **Fragile assertion** (line 222): `assert "premise-a" in call_kwargs[1]["sl"] or "premise-a" in call_kwargs[0][1] if call_kwargs[0] else call_kwargs[1]["sl"]` — Python precedence makes this a ternary. The truthy branch evaluates `"premise-a" in call_kwargs[1]["sl"] or "premise-a" in call_kwargs[0][1]`. Since there's only one positional arg, `call_kwargs[0][1]` would be an `IndexError` if the first `or` branch were False. This works only because the first branch happens to succeed. Should be: `assert "premise-a" in call_kwargs[1]["sl"]`.

2. **Missing test scenarios**: No test for nodes with outlists (would expose the outlist-dropping bug), no test for multiple invalid beliefs in one batch, no test for exception during LLM call.

### Self-Review
**Limitations:** I did not run the test suite to confirm tests pass. I relied on the observation data for `add_justification`'s signature — verified it accepts `unless` kwarg, confirming the outlist-dropping fix is viable. I did not see the full `api.search` return format for `format="json"` to verify the search result schema matches what repair.py expects (fields like `id`, `text`, `truth_value`).

### Feature Requests
- Include the return schema/format of functions called by the changed code (e.g., what does `api.search(query, format="json")` actually return?) to verify field-level compatibility
- Flag test assertions with complex operator precedence as potential issues automatically

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#163
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### reasons_lib/repair.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

**Potential IndexError (line 237):** `first_just = node.get("justifications", [{}])[0]` — if the key exists but the list is empty (`[]`), this raises `IndexError`. The default `[{}]` only applies when the key is absent. Derived beliefs *should* always have justifications, but an empty-list edge case from malformed data isn't defended against. Wrap in a guard or use `(node.get("justifications") or [{}])[0]`.

**Search format coupling (line 187-196):** The code calls `api.search(claim, format="json", db_path=db_path)` and expects the result to be a JSON list of dicts with `id`, `text`, `truth_value` keys. All tests mock `search_fn` entirely, so no test exercises the real `api.search` → `_format_json` → parse path. If `_format_json` returns a different structure (e.g. a dict with `"matched"` and `"neighbors"` keys, or nested objects), this will silently produce empty candidates and report `no_candidates` for everything — a silent failure mode.

**Minor:** `parse_extract_response` prefix stripping (line 82) does `text.lower().startswith(prefix.lower())` then slices by `len(prefix)` — this is correct since both strings have the same length regardless of case, but it's a pattern that could confuse a reader.

Overall the module is well-structured: injectable `search_fn`, two-phase LLM pipeline, good candidate filtering (skips self, existing antecedents, OUT nodes, derived beliefs).

### reasons_lib/repair.py:repair_smuggled_beliefs
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

**Justification wiring uses only the first justification (line 237-245):** When a node has multiple justifications, the repair always copies the outlist from index 0. This is a reasonable default but could produce surprising results if the invalid review was about a different justification. The design choice is defensible but should be documented.

**The new justification is additive, not a replacement:** `api.add_justification` adds a *new* justification alongside the original. After repair, the node has both the original (invalid) justification and the new (enhanced) one. Since a node is IN if ANY justification is valid, this is correct — the node stays IN via the repaired justification even if the original is later removed. Good TMS-aware design.

### reasons_lib/api.py:repair_smuggled
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Clean API wrapper with two input modes. The `belief_ids` path correctly passes `dry_run=True` to `review_beliefs` so it only reads without writing review metadata. Status counting logic is straightforward.

The `belief_ids` path (re-review inline → repair) lacks an end-to-end test — only the `review_file` path is tested through the CLI. The unit tests for `repair_smuggled_beliefs` cover the repair engine itself, but the `review_beliefs` → extract-invalids → repair chain is untested.

`export_network()` is called to get all nodes. For very large networks this could be expensive but is acceptable since the operation is already LLM-bound.

### reasons_lib/cli.py:cmd_repair_smuggled
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Straightforward CLI handler. Output formatting is clear. The error case (no review_file and no belief_ids) exits with code 1 and prints to stderr. Parser registration at line 2028 is correct. The `--dry-run` flag is properly threaded through and announced in output.

Minor: `getattr(args, "model", None) or "claude"` — argparse already sets `default=None`, so `getattr` is unnecessary (just `args.model or "claude"` would suffice), but this is harmless.

### tests/test_repair.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test suite covering:
- Both parse functions with edge cases (malformed JSON, prefix stripping, ID filtering)
- LLM integration points (extract, match) with proper subprocess mocking
- Full repair flow: happy path, dry run, no candidates, no match, extraction failure
- Important edge cases: existing antecedents skipped, derived beliefs excluded from candidates, missing nodes, outlist preservation, multiple beliefs, valid beliefs skipped
- CLI integration: help, error, review-file path, dry run

`test_from_review_file` is a real integration test that creates actual DB nodes and runs through the CLI — only the LLM subprocess is mocked. This exercises the real search and add_justification paths.

One gap: no CLI test for the `belief_ids` input mode (only `--review-file` is tested). This leaves the inline re-review → repair path exercised only at the API level (indirectly through `TestRepairSmuggledBeliefs`).

### Self-Review
**Limitations:** Could not see `_format_json` implementation to verify the JSON structure that `api.search(format="json")` actually returns — this is the key coupling point between repair.py and the search subsystem. Could not verify whether FTS5 tables are auto-created by `add_node` (relevant for the CLI integration test). Did not see the `review.py:format_belief_for_review` callers to check if the import in repair.py could create a circular dependency (api → repair → review, api → review — but since repair.py imports api lazily inside the function, this should be fine).

### Feature Requests
- Include the implementation of functions imported from sibling modules (e.g., `_format_json` from api.py) when they are called by the diff under review — format coupling is a common source of silent failures.
- Flag when test mocks completely bypass a production code path (e.g., all tests mock `search_fn`, so the real `api.search` → parse → filter chain is never tested in unit tests).

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

PR #175, opened 2026-06-04
Add git-aware source pinning for staleness detection

### LINKED ISSUE #169: Pin source links to commit SHA for staleness detection
## Problem

Source links in beliefs currently point to branch URLs (e.g., blob/main/path/to/file.py), which are moving targets. A belief verified against todays main may be stale tomorrow if the source file changes, but the link does not capture when it was verified or against which version.

## Proposal

Store the full commit SHA in source links so that:

1. **Verification is pinned to immutable code** -- a belief verified against a SHA is verified against bytes that can never change
2. **Mechanical staleness detection** -- compare the stored SHA against the current HEAD for that file path. If they differ, the belief is potentially stale. This is a string comparison, no LLM needed.
3. **Line-range pinning (future)** -- store not just the file SHA but the line span the belief depends on. Staleness check becomes did the diff touch these lines -- still mechanical, far fewer false positives.

## Detection Pipeline

SHA changed? (string comparison, cheap) -> No: belief still verified, skip. Yes: did relevant lines change? (git diff, cheap) -> No: update SHA, no content change. Yes: LLM re-verification (expensive, fallible) -> Human reviews result.

Each layer does what it is good at. Cheap reliable parts carry the routine load. Expensive fallible part (LLM) fires only when needed.

## Additional metadata

- Store verification date alongside SHA for triage context
- Use full 40-char SHA, not short SHA, to avoid ambiguity
- GitHub raw endpoint at full SHA is content-addressed and stable

## Origin

Feedback from Opus 4.8 reviewing https://expert.ftl2.com -- independently traced belief to entry to source code and identified that branch URLs are a moving target that undermines verification. Extends existing reasons check-stale with more precise, mechanical detection.

benthomasson:
## Summary

- Adds `pin-sources` command to store git commit SHA (`pinned_sha`) and verification date in node metadata
- Adds `pin-update` command to bump `pinned_sha` after re-verifying stale beliefs
- Adds `--git` flag to `check-stale` for tiered staleness detection: skip content hashing when no commits touched the file since pinned SHA, auto-bump SHA when content unchanged
- Adds `--pin-urls` flag to rewrite GitHub branch URLs to immutable SHA-pinned URLs

Closes #169

## Test plan

- [x] 28 new tests pass covering all pin functions and git-aware staleness
- [x] 45 existing check-stale tests still pass
- [ ] Manual integration test with real git repo

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1653,14 +1653,15 @@ def check_stale(
             1 for n in net.nodes.values()
             if n.truth_value == "IN" and n.source and n.source_hash
         )
-        results, upgraded = _check(net, repo_paths, db_dir=db_dir,
-                                   upgrade_hashes=upgrade_hashes,
-                                   git_aware=git_aware)
+        results, upgraded, sha_bumped = _check(net, repo_paths, db_dir=db_dir,
+                                               upgrade_hashes=upgrade_hashes,
+                                               git_aware=git_aware)
         return {
             "stale": results,
             "checked": in_with_source,
             "stale_count": len(results),
             "upgraded": upgraded,
+            "sha_bumped": sha_bumped,
         }
 
 
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1618,6 +1618,7 @@ def export_card(visible_to: list[str] | None = None, db_path: str = DEFAULT_DB,
 def check_stale(
     repos: dict[str, str] | None = None,
     upgrade_hashes: bool = False,
+    git_aware: bool = False,
     db_path: str = DEFAULT_DB,
     pg_conninfo=None, project_id=None,
 ) -> dict:
@@ -1625,6 +1626,9 @@ def check_stale(
 
     If upgrade_hashes=True, truncated hashes that prefix-match the current
     full hash are upgraded in place and saved to the database.
+
+    If git_aware=True, nodes with pinned_sha skip content hashing when
+    no commits touched the file since the pinned SHA.
 
     Returns: {"stale": list[dict], "checked": int, "stale_count": int,
               "upgraded": int}
@@ -1637,7 +1641,8 @@ def check_stale(
 
     db_dir = P(db_path).resolve().parent
 
-    with _with_network(db_path, write=upgrade_hashes) as net:
+    needs_write = upgrade_hashes or git_aware
+    with _with_network(db_path, write=needs_write) as net:
         repo_paths = repos
         if repo_paths is None and net.repos:
             repo_paths = net.repos
@@ -1649,7 +1654,8 @@ def check_stale(
             if n.truth_value == "IN" and n.source and n.source_hash
         )
         results, upgraded = _check(net, repo_paths, db_dir=db_dir,
-                                   upgrade_hashes=upgrade_hashes)
+                                   upgrade_hashes=upgrade_hashes,
+                                   git_aware=git_aware)
         return {
             "stale": results,
             "checked": in_with_source,
@@ -1685,6 +1691,62 @@ def hash_sources(
 
         results = _hash(net, repo_paths, force=force, db_dir=db_dir)
         return {"hashed": results, "count": len(results)}
+
+
+def pin_sources(
+    force: bool = False,
+    pin_urls: bool = False,
+    repos: dict[str, str] | None = None,
+    db_path: str = DEFAULT_DB,
+) -> dict:
+    """Pin IN nodes to their current git commit SHA.
+
+    Stores pinned_sha and verified_at in node metadata.
+
+    Returns: {"pinned": list[dict], "count": int}
+    """
+    from pathlib import Path as P
+    from .check_stale import pin_sources as _pin
+
+    db_dir = P(db_path).resolve().parent
+
+    with _with_network(db_path, write=True) as net:
+        repo_paths = repos
+        if repo_paths is None and net.repos:
+            repo_paths = net.repos
+        if repo_paths:
+            repo_paths = {k: P(v) for k, v in repo_paths.items()}
+
+        results = _pin(net, repo_paths, db_dir=db_dir,
+                        force=force, pin_urls=pin_urls)
+        return {"pinned": results, "count": len(results)}
+
+
+def pin_update(
+    node_ids: list[str],
+    repos: dict[str, str] | None = None,
+    db_path: str = DEFAULT_DB,
+) -> dict:
+    """Bump pinned_sha to current HEAD for specified nodes.
+
+    Returns: {"updated": list[dict], "count": int, "errors": int}
+    """
+    from pathlib import Path as P
+    from .check_stale import pin_update as _update
+
+    db_dir = P(db_path).resolve().parent
+
+    with _with_network(db_path, write=True) as net:
+        repo_paths = repos
+        if repo_paths is None and net.repos:
+            repo_paths = net.repos
+        if repo_paths:
+            repo_paths = {k: P(v) for k, v in repo_paths.items()}
+
+        results = _update(net, node_ids, repo_paths, db_dir=db_dir)
+        errors = sum(1 for r in results if "error" in r)
+        return {"updated": results, "count": len(results) - errors,
+                "errors": errors}
 
 
 def compact(budget: int = 500, truncate: bool = True, visible_to: list[str] | None = None, db_path: str = DEFAULT_DB,
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1631,11 +1631,12 @@ def check_stale(
     no commits touched the file since the pinned SHA.
 
     Returns: {"stale": list[dict], "checked": int, "stale_count": int,
-              "upgraded": int}
+              "upgraded": int, "sha_bumped": int}
     """
     if pg_conninfo:
         return _pg_dispatch(pg_conninfo, project_id, "check_stale",
-                            repos=repos, upgrade_hashes=upgrade_hashes)
+                            repos=repos, upgrade_hashes=upgrade_hashes,
+                            git_aware=git_aware)
     from pathlib import Path as P
     from .check_stale import check_stale as _check
 
--- a/reasons_lib/check_stale.py
+++ b/reasons_lib/check_stale.py
@@ -6,6 +6,9 @@ source_hash against the current SHA-256 hash of the source file.
 """
 
 import hashlib
+import re
+import subprocess
+from datetime import date
 from pathlib import Path
 
 from .network import Network
@@ -61,11 +64,15 @@ def check_stale(
     repos: dict[str, Path] | None = None,
     db_dir: Path | None = None,
     upgrade_hashes: bool = False,
+    git_aware: bool = False,
 ) -> tuple[list[dict], int]:
     """Check all IN nodes for source staleness.
 
     If upgrade_hashes=True, truncated hashes that are a prefix of the
     current full hash are upgraded in place (caller must save the network).
+
+    If git_aware=True, nodes with pinned_sha in metadata skip content
+    hashing when no commits touched the file since the pinned SHA.
 
     Returns (stale_results, upgraded_count).
     """
@@ -94,6 +101,11 @@ def check_stale(
             })
             continue
 
+        pinned_sha = node.metadata.get("pinned_sha") if node.metadata else None
+        if git_aware and pinned_sha:
+            if not file_changed_since(path, pinned_sha):
+                continue
+
         current_hash = hash_file(path)
         if current_hash != node.source_hash:
             if len(node.source_hash) == 16 and current_hash.startswith(node.source_hash):
@@ -118,6 +130,13 @@ def check_stale(
                 "source_path": str(path),
                 "reason": "content_changed",
             })
+        elif git_aware and pinned_sha:
+            # Content hash unchanged despite commits — auto-bump pinned_sha
+            new_sha = get_file_commit_sha(path)
+            if new_sha and new_sha != pinned_sha:
+                node.metadata["pinned_sha"] = new_sha
+                node.metadata["verified_at"] = date.today().isoformat()
+                upgraded += 1
 
     return results, upgraded
 
@@ -160,6 +179,157 @@ def hash_sources(
             "source": node.source,
             "hash": new_hash,
             "was_empty": was_empty,
+        })
+
+    return results
+
+
+# --- Git-aware pinning ---
+
+
+def get_file_commit_sha(filepath: Path) -> str | None:
+    """Get the last commit SHA that touched this file."""
+    result = subprocess.run(
+        ["git", "log", "-1", "--format=%H", "--", str(filepath.name)],
+        capture_output=True, text=True,
+        cwd=str(filepath.parent),
+    )
+    if result.returncode == 0 and result.stdout.strip():
+        return result.stdout.strip()
+    return None
+
+
+def file_changed_since(filepath: Path, since_sha: str) -> bool:
+    """Check if file was modified in any commit after since_sha."""
+    result = subprocess.run(
+        ["git", "log", "--format=%H", f"{since_sha}..HEAD",
+         "--", str(filepath.name)],
+        capture_output=True, text=True,
+        cwd=str(filepath.parent),
+    )
+    return bool(result.returncode == 0 and result.stdout.strip())
+
+
+def pin_source_url(url: str, sha: str) -> str:
+    """Rewrite GitHub blob URL from branch ref to commit SHA."""
+    m = re.match(r'(https://github\.com/[^/]+/[^/]+/blob/)[^/]+/(.*)', url)
+    if m:
+        return f"{m.group(1)}{sha}/{m.group(2)}"
+    return url
+
+
+def pin_sources(
+    network: Network,
+    repos: dict[str, Path] | None = None,
+    db_dir: Path | None = None,
+    force: bool = False,
+    pin_urls: bool = False,
+) -> list[dict]:
+    """Pin IN nodes to their current git commit SHA.
+
+    Stores pinned_sha and verified_at in node metadata. Also backfills
+    source_hash if missing.
+
+    Returns list of dicts: {"node_id", "source", "pinned_sha", "was_empty"}
+    """
+    if repos is None and network.repos:
+        repos = {k: Path(v) for k, v in network.repos.items()}
+
+    today = date.today().isoformat()
+    results = []
+
+    for nid, node in sorted(network.nodes.items()):
+        if node.truth_value != "IN":
+            continue
+        if not node.source:
+            continue
+        if not force and node.metadata.get("pinned_sha"):
+            continue
+
+        agent = node.metadata.get("agent") if node.metadata else None
+        path = resolve_source_path(node.source, repos, db_dir, agent=agent)
+        if path is None:
+            continue
+
+        sha = get_file_commit_sha(path)
+        if sha is None:
+            continue
+
+        was_empty = not node.metadata.get("pinned_sha")
+        node.metadata["pinned_sha"] = sha
+        node.metadata["verified_at"] = today
+
+        if not node.source_hash:
+            node.source_hash = hash_file(path)
+
+        if pin_urls and node.source_url:
+            node.source_url = pin_source_url(node.source_url, sha)
+
+        results.append({
+            "node_id": nid,
+            "source": node.source,
+            "pinned_sha": sha,
+            "was_empty": was_empty,
+        })
+
+    return results
+
+
+def pin_update(
+    network: Network,
+    node_ids: list[str],
+    repos: dict[str, Path] | None = None,
+    db_dir: Path | None = None,
+) -> list[dict]:
+    """Bump pinned_sha to current HEAD for specified nodes.
+
+    Returns list of dicts: {"node_id", "old_sha", "new_sha", "source"}
+    """
+    if repos is None and network.repos:
+        repos = {k: Path(v) for k, v in network.repos.items()}
+
+    today = date.today().isoformat()
+    results = []
+
+    for nid in node_ids:
+        node = network.nodes.get(nid)
+        if node is None:
+            results.append({
+                "node_id": nid, "error": "not found",
+            })
+            continue
+
+        if not node.source:
+            results.append({
+                "node_id": nid, "error": "no source",
+            })
+            continue
+
+        agent = node.metadata.get("agent") if node.metadata else None
+        path = resolve_source_path(node.source, repos, db_dir, agent=agent)
+        if path is None:
+            results.append({
+                "node_id": nid, "error": "source not found",
+            })
+            continue
+
+        sha = get_file_commit_sha(path)
+        if sha is None:
+            results.append({
+                "node_id": nid, "error": "not in git repo",
+            })
+            continue
+
+        old_sha = node.metadata.get("pinned_sha", "")
+        node.metadata["pinned_sha"] = sha
+        node.metadata["verified_at"] = today
+        node.source_hash = hash_file(path)
+
+        results.append({
+            "node_id": nid,
+            "old_sha": old_sha,
+            "new_sha": sha,
+            "source": node.source,
         })
 
     return results
--- a/reasons_lib/check_stale.py
+++ b/reasons_lib/check_stale.py
@@ -65,7 +65,7 @@ def check_stale(
     db_dir: Path | None = None,
     upgrade_hashes: bool = False,
     git_aware: bool = False,
-) -> tuple[list[dict], int]:
+) -> tuple[list[dict], int, int]:
     """Check all IN nodes for source staleness.
 
     If upgrade_hashes=True, truncated hashes that are a prefix of the
@@ -74,13 +74,14 @@ def check_stale(
     If git_aware=True, nodes with pinned_sha in metadata skip content
     hashing when no commits touched the file since the pinned SHA.
 
-    Returns (stale_results, upgraded_count).
+    Returns (stale_results, upgraded_count, sha_bumped_count).
     """
     if repos is None and network.repos:
         repos = {k: Path(v) for k, v in network.repos.items()}
 
     results = []
     upgraded = 0
+    sha_bumped = 0
 
     for nid, node in sorted(network.nodes.items()):
         if node.truth_value != "IN":
@@ -131,14 +132,13 @@ def check_stale(
                 "reason": "content_changed",
             })
         elif git_aware and pinned_sha:
-            # Content hash unchanged despite commits — auto-bump pinned_sha
             new_sha = get_file_commit_sha(path)
             if new_sha and new_sha != pinned_sha:
                 node.metadata["pinned_sha"] = new_sha
                 node.metadata["verified_at"] = date.today().isoformat()
-                upgraded += 1
+                sha_bumped += 1
 
-    return results, upgraded
+    return results, upgraded, sha_bumped
 
 
 def hash_sources(
@@ -200,14 +200,30 @@ def get_file_commit_sha(filepath: Path) -> str | None:
 
 
 def file_changed_since(filepath: Path, since_sha: str) -> bool:
-    """Check if file was modified in any commit after since_sha."""
+    """Check if file was modified since since_sha (committed or uncommitted).
+
+    Returns True on git errors to force fallback to content hashing.
+    """
+    # Check committed changes
     result = subprocess.run(
         ["git", "log", "--format=%H", f"{since_sha}..HEAD",
          "--", str(filepath.name)],
         capture_output=True, text=True,
         cwd=str(filepath.parent),
     )
-    return bool(result.returncode == 0 and result.stdout.strip())
+    if result.returncode != 0:
+        return True
+    if result.stdout.strip():
+        return True
+    # Check uncommitted changes (staged + unstaged)
+    result = subprocess.run(
+        ["git", "diff", "--quiet", "HEAD", "--", str(filepath.name)],
+        capture_output=True, text=True,
+        cwd=str(filepath.parent),
+    )
+    if result.returncode != 0:
+        return True
+    return False
 
 
 def pin_source_url(url: str, sha: str) -> str:
--- a/reasons_lib/check_stale.py
+++ b/reasons_lib/check_stale.py
@@ -193,6 +193,7 @@ def get_file_commit_sha(filepath: Path) -> str | None:
         ["git", "log", "-1", "--format=%H", "--", str(filepath.name)],
         capture_output=True, text=True,
         cwd=str(filepath.parent),
+        timeout=30,
     )
     if result.returncode == 0 and result.stdout.strip():
         return result.stdout.strip()
@@ -204,33 +205,49 @@ def file_changed_since(filepath: Path, since_sha: str) -> bool:
 
     Returns True on git errors to force fallback to content hashing.
     """
-    # Check committed changes
-    result = subprocess.run(
-        ["git", "log", "--format=%H", f"{since_sha}..HEAD",
-         "--", str(filepath.name)],
-        capture_output=True, text=True,
-        cwd=str(filepath.parent),
-    )
-    if result.returncode != 0:
+    try:
+        # Check committed changes
+        result = subprocess.run(
+            ["git", "log", "--format=%H", f"{since_sha}..HEAD",
+             "--", str(filepath.name)],
+            capture_output=True, text=True,
+            cwd=str(filepath.parent),
+            timeout=30,
+        )
+        if result.returncode != 0:
+            return True
+        if result.stdout.strip():
+            return True
+        # Check uncommitted changes (staged + unstaged)
+        result = subprocess.run(
+            ["git", "diff", "--quiet", "HEAD", "--", str(filepath.name)],
+            capture_output=True, text=True,
+            cwd=str(filepath.parent),
+            timeout=30,
+        )
+        if result.returncode != 0:
+            return True
+        return False
+    except subprocess.TimeoutExpired:
         return True
-    if result.stdout.strip():
-        return True
-    # Check uncommitted changes (staged + unstaged)
-    result = subprocess.run(
-        ["git", "diff", "--quiet", "HEAD", "--", str(filepath.name)],
-        capture_output=True, text=True,
-        cwd=str(filepath.parent),
-    )
-    if result.returncode != 0:
-        return True
-    return False
 
 
-def pin_source_url(url: str, sha: str) -> str:
-    """Rewrite GitHub blob URL from branch ref to commit SHA."""
-    m = re.match(r'(https://github\.com/[^/]+/[^/]+/blob/)[^/]+/(.*)', url)
-    if m:
-        return f"{m.group(1)}{sha}/{m.group(2)}"
+def pin_source_url(url: str, sha: str, source_path: str = "") -> str:
+    """Rewrite GitHub blob URL from branch ref to commit SHA.
+
+    If source_path is provided, uses it to locate the file path portion
+    of the URL (handles branch names containing slashes).
+    """
+    m = re.match(r'(https://github\.com/[^/]+/[^/]+/blob/).+', url)
+    if not m:
+        return url
+    prefix = m.group(1)
+    if source_path and url.endswith(source_path):
+        return f"{prefix}{sha}/{source_path}"
+    rest = url[len(prefix):]
+    parts = rest.split("/", 1)
+    if len(parts) == 2:
+        return f"{prefix}{sha}/{parts[1]}"
     return url
 
 
@@ -279,7 +296,7 @@ def pin_sources(
             node.source_hash = hash_file(path)
 
         if pin_urls and node.source_url:
-            node.source_url = pin_source_url(node.source_url, sha)
+            node.source_url = pin_source_url(node.source_url, sha, node.source)
 
         results.append({
             "node_id": nid,
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -696,7 +696,9 @@ def cmd_hash_sources(args):
 
 
 def cmd_check_stale(args):
-    result = api.check_stale(upgrade_hashes=args.upgrade_hashes, **_backend_kwargs(args))
+    result = api.check_stale(upgrade_hashes=args.upgrade_hashes,
+                             git_aware=getattr(args, "git", False),
+                             **_backend_kwargs(args))
 
     if result.get("upgraded"):
         print(f"Upgraded {result['upgraded']} truncated hash(es) to full length.")
@@ -726,6 +728,51 @@ def cmd_check_stale(args):
     print(f"{fresh} fresh, {len(stale)} stale (of {result['checked']} checked)")
     if stale:
         sys.exit(1)
+
+
+def cmd_pin_sources(args):
+    _require_sqlite(args, "pin-sources")
+    result = api.pin_sources(
+        force=args.force,
+        pin_urls=getattr(args, "pin_urls", False),
+        db_path=args.db,
+    )
+
+    if not result["pinned"]:
+        print("No nodes to pin (all sources already pinned, or source files not in git).")
+        if not args.force:
+            print("Use --force to re-pin nodes that already have a pinned_sha.")
+        return
+
+    for item in result["pinned"]:
+        action = "pinned" if item["was_empty"] else "re-pinned"
+        print(f"  {action}  {item['node_id']}  {item['pinned_sha'][:12]}  ({item['source']})")
+
+    new = sum(1 for p in result["pinned"] if p["was_empty"])
+    re = result["count"] - new
+    parts = []
+    if new:
+        parts.append(f"{new} pinned")
+    if re:
+        parts.append(f"{re} re-pinned")
+    print(f"\n{', '.join(parts)}")
+
+
+def cmd_pin_update(args):
+    _require_sqlite(args, "pin-update")
+    result = api.pin_update(node_ids=args.node_ids, db_path=args.db)
+
+    for item in result["updated"]:
+        if "error" in item:
+            print(f"  ERROR  {item['node_id']}: {item['error']}")
+        else:
+            old = item['old_sha'][:12] if item['old_sha'] else "(none)"
+            print(f"  UPDATED  {item['node_id']}  {old} -> {item['new_sha'][:12]}")
+
+    if result["errors"]:
+        print(f"\n{result['count']} updated, {result['errors']} errors")
+    else:
+        print(f"\n{result['count']} updated")
 
 
 def cmd_compact(args):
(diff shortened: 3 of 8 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#175
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### reasons_lib/check_stale.py:get_file_commit_sha
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Uses `filepath.name` with `cwd=filepath.parent`, which correctly scopes the git log to the file's location in the repo. Returns `None` on failure (non-git directory, untracked file). 4 tests cover the happy and error paths.

### reasons_lib/check_stale.py:file_changed_since
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

When `git log` fails (e.g., file not in a git repo, or `since_sha` is invalid/orphaned), `returncode != 0` causes it to return `False` ("no change"). In `check_stale`, this means a node with a `pinned_sha` referencing an invalid/force-pushed SHA silently skips staleness checking, treating the file as fresh. The practical risk is low since `pinned_sha` is only set by `pin_sources` which validates git access, but a rebased repo could hit this.

### reasons_lib/check_stale.py:check_stale (git_aware additions)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `upgraded` counter is overloaded — it counts both truncated hash upgrades (line ~108) and pinned_sha auto-bumps (line ~137). The CLI handler at `cli.py:703` prints `"Upgraded {n} truncated hash(es) to full length."` which is misleading when auto-bumps contribute to the count. The auto-bump logic itself (elif branch at line ~131) is correct: it only fires when `file_changed_since` was True but content hash matches, which means a benign commit touched the file. No test covers the interaction where both truncated-hash upgrades and pinned_sha bumps happen in the same run.

### reasons_lib/check_stale.py:pin_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly skips OUT nodes, sourceless nodes, already-pinned nodes (unless `force`). Backfills `source_hash` when empty. URL pinning with `pin_urls` is correctly gated. 9 tests cover all branches including force re-pin, URL rewriting, and skip conditions.

### reasons_lib/check_stale.py:pin_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Error handling for missing nodes, sourceless nodes, unresolvable paths, and non-git files is thorough — each returns a dict with an `"error"` key. Successfully updates `pinned_sha`, `verified_at`, and refreshes `source_hash`. 5 tests cover success and all 4 error paths.

### reasons_lib/check_stale.py:pin_source_url
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Regex correctly matches GitHub blob URLs and rewrites the branch/ref segment to a commit SHA. Non-matching URLs pass through unchanged. 3 tests cover branch URL, non-GitHub URL, and arbitrary branch names.

### reasons_lib/api.py:check_stale (git_aware parameter)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

The `git_aware` parameter is accepted but **not forwarded** to the PG dispatch path (line ~1637). If a user runs `reasons check-stale --git --pg ...`, the `--git` flag is silently ignored. Either `git_aware` should be forwarded (if PgApi supports it) or the CLI should reject the combination. The `needs_write = upgrade_hashes or git_aware` change is correct for the SQLite path.

### reasons_lib/api.py:pin_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Follows the same repo-resolution pattern as `hash_sources`. Opens network with `write=True`. No PG support, which is consistent with `cmd_pin_sources` calling `_require_sqlite`. No direct test of this API wrapper (only the underlying `check_stale.pin_sources` is tested), but the wrapper is thin enough that the risk is low.

### reasons_lib/api.py:pin_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Follows same pattern as `pin_sources`. Correctly computes `errors` count. No direct test of this API wrapper but the underlying function is well-tested.

### reasons_lib/cli.py:cmd_pin_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Calls `_require_sqlite` to block PG usage. Output formatting mirrors `cmd_hash_sources` style. Correctly distinguishes "pinned" vs "re-pinned" in output. Registered in both the parser and dispatch dict. No CLI-level test exists but the function is straightforward.

### reasons_lib/cli.py:cmd_pin_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Calls `_require_sqlite`. Handles error vs success results in display. Registered in parser (with `nargs="+"` for multiple node IDs) and dispatch dict. No CLI-level test.

### reasons_lib/cli.py:cmd_check_stale (modified)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Uses `getattr(args, "git", False)` which is unnecessary since `--git` is defined on the parser (it will always have the attribute), but this is harmless. The real issue: the "Upgraded N truncated hash(es)" message (line 703-704) will now also count pinned_sha auto-bumps, producing a misleading message. The `--git` flag should probably not be combinable with `--pg` but there's no guard here (the guard is only on `pin-sources`/`pin-update`).

### tests/test_pin_sources.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

29 tests across 6 test classes covering all new functions and the git_aware integration with `check_stale`. The `git_repo` fixture properly initializes a temporary git repository with config. Tests cover: tracked/untracked files, change detection, URL rewriting, pin/re-pin/skip/force flows, all error paths in `pin_update`, and the git_aware check_stale scenarios including auto-bump behavior. Missing: no test for the `upgraded` counter confusion when both hash upgrades and SHA bumps occur in the same run.

### Self-Review
**Limitations:** Could not verify PgApi.check_stale signature to confirm whether it could accept a git_aware parameter. Could not run the test suite to confirm all 29 tests pass. Did not have access to the Node dataclass definition to verify metadata is always a dict (not None) when accessed in pin_sources/pin_update — the code does check `node.metadata.get(...)` but `add_node` defaults metadata to `{}`, so this should be safe.

### Feature Requests
- Include the PgApi class definition (or at least the relevant method signatures) when changes affect the PG dispatch path, to verify parameter forwarding
- Show the full Node/dataclass definition when field access patterns are modified

## gemini [BLOCK]

### reasons_lib/api.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The new `pin_sources` and `pin_update` functions are correctly added but are SQLite-only for now; they properly use `_require_sqlite` in the CLI to prevent use with PostgreSQL. However, the `check_stale` function drops the `git_aware` argument when dispatching to the PostgreSQL backend (line 1637), meaning the `--git` flag will be silently ignored when using `--pg`.

### reasons_lib/check_stale.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `git_aware` staleness check has two significant issues:
1. **Misses Uncommitted Changes**: `file_changed_since` uses `git log` to check for changes since a SHA (lines 189-196). `git log` only inspects committed history. If a user modifies a file locally but has not committed it, `file_changed_since` will return `False` (assuming `pinned_sha` is HEAD), and `check_stale` will incorrectly report the node as fresh (line 98). A more correct approach would be using `git diff --quiet <sha> -- <file>`.
2. **Fragile Error Handling**: If `git log` fails (e.g., invalid SHA, not in a git repo), `file_changed_since` returns `False` (line 197). This causes `check_stale` to skip the content hash fallback and report the node as fresh even if it's potentially stale or the git check failed. Errors in the git check should force a fallback to content hashing.

### reasons_lib/cli.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The CLI integration is mostly correct and follows existing patterns. However, the summary message in `cmd_check_stale` (lines 704-705) is misleading when `git_aware` is used: it reports all "upgraded" nodes as "truncated hash(es) upgraded", even if the upgrade was actually an auto-bump of a `pinned_sha` in metadata (line 127 in `check_stale.py`). The `upgraded` count should likely be split or the message generalized.

### tests/test_pin_sources.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Excellent test suite. It uses a real git repository to verify the subprocess interactions and covers a wide range of edge cases including untracked files, URL rewriting, and forced re-pinning.

### Self-Review
**Limitations:** I was able to see full function bodies for all modified logic. I verified the git command behavior via standard git documentation and common usage. I identified that the PostgreSQL backend (`pg.py`) was not updated, which matches the explicit exclusion in `cli.py`.

### Feature Requests
- It would be helpful to have `pg.py` included in the diff whenever `api.py` methods are modified to ensure backend parity.
- A "run tests" action in the review tool would allow verifying the false negative behavior for uncommitted changes directly.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#175
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [BLOCK]

### reasons_lib/check_stale.py:get_file_commit_sha
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Uses `filepath.name` (basename only) in the git command at line ~190: `"--", str(filepath.name)`. This means if two files in different subdirectories share the same name, git could return the wrong commit SHA. Should use the full path relative to the repo root, or the absolute path. The `cwd=filepath.parent` partially mitigates this for files directly in the repo root, but for files in subdirectories (e.g. `entries/topic.md`), `filepath.parent` is the `entries/` dir and `filepath.name` is `topic.md` — this actually works correctly since git scopes the log to files in the cwd subtree. However, if the cwd isn't inside a git repo (or is a nested repo), behavior is undefined. The fail-open design (return None) is appropriate.

### reasons_lib/check_stale.py:file_changed_since
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same `filepath.name` pattern as `get_file_commit_sha`. Returns True (fail-open) on git errors, which is the right default. However, the `git diff --quiet HEAD` check at line ~215 will fail with returncode 128 (not just 1) if there is no HEAD (empty repo) — this is still safe because returncode != 0 returns True (forces content hash fallback). One edge case: uncommitted *new* files that aren't tracked — `git diff HEAD` won't see them, but these files wouldn't have a `pinned_sha` anyway so the code path wouldn't be reached.

### reasons_lib/check_stale.py:check_stale (modified)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The SHA auto-bump logic at lines 134-139 runs in the `elif` branch — meaning it only fires when `current_hash == node.source_hash` AND `git_aware` AND `pinned_sha` is set. This is correct: content matches but a new commit touched the file, so bump the pin. However, there's a subtle issue: the early-continue at line 106-108 skips content hashing entirely when `file_changed_since` returns False. This means if git says "no change" but the file WAS actually modified outside git (e.g. direct filesystem edit not yet staged), the staleness goes undetected. This is by design (the docstring says "skip content hashing") but worth noting. The return type change from 2-tuple to 3-tuple is a breaking change for `PgApi.check_stale` (see below).

### reasons_lib/check_stale.py:pin_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean implementation. Correctly skips OUT nodes, nodes without source, already-pinned nodes (unless force=True). Backfills `source_hash` when missing. The `pin_urls` flag correctly rewrites GitHub blob URLs. Tests cover all branches.

### reasons_lib/check_stale.py:pin_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Error handling covers all cases (missing node, no source, unresolvable source, not in git). Correctly refreshes both `pinned_sha` and `source_hash`. Good that it unconditionally refreshes `source_hash` since the user is explicitly saying "I've verified this."

### reasons_lib/check_stale.py:pin_source_url
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple regex rewrite. Correctly preserves non-GitHub URLs by returning them unchanged. Covers the common `github.com/.../blob/<ref>/...` pattern.

### reasons_lib/api.py:check_stale (modified)
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The `_pg_dispatch` call at line 1637 passes `git_aware` through to `PgApi.check_stale`, but `PgApi.check_stale` (line 821 in pg.py) does NOT accept a `git_aware` parameter — its signature is `check_stale(self, repos=None, upgrade_hashes=False)`. When `pg_conninfo` is set, `_pg_dispatch` will forward `git_aware=True` as a keyword argument, which will raise `TypeError: check_stale() got an unexpected keyword argument 'git_aware'`. Additionally, the docstring at line 1633-1634 lists the return dict keys but omits `sha_bumped` which is now included.

### reasons_lib/api.py:pin_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Correctly delegates to the check_stale module function. Opens the network in write mode. No direct tests of the API layer (tests call `check_stale.pin_sources` directly), but the CLI integration wires through this, so it's exercised end-to-end if CLI tests exist.

### reasons_lib/api.py:pin_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Same pattern as pin_sources. Error counting logic is correct. Same caveat about no direct API-layer tests.

### reasons_lib/cli.py:cmd_pin_sources
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Uses `_require_sqlite` which is good. However, variable `re` shadows the stdlib `re` import at module level (line 754: `re = result["count"] - new`). While `re` isn't imported in cli.py, this is a confusing variable name. Minor: no CLI-level tests for the new commands in the diff.

### reasons_lib/cli.py:cmd_pin_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Clean implementation. Handles error and success cases in output. Parser correctly requires `node_ids` as positional args with `nargs="+"`.

### reasons_lib/cli.py:cmd_check_stale (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Uses `getattr(args, "git", False)` which is safe. New `--git` flag wired correctly to `api.check_stale(git_aware=...)`. Output message for `sha_bumped` is clear.

### tests/test_pin_sources.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good test coverage. Creates real git repos for integration testing. Covers: SHA retrieval, change detection, URL rewriting, pinning (with force, skip, backfill, URL rewrite), updating (with all error cases), and git-aware staleness checks (fresh, stale, auto-bump, fallback, git_aware=False). The `test_auto_bumps_sha_when_content_unchanged` test is particularly good — it verifies the change-then-revert scenario.

### tests/test_check_stale.py + tests/test_check_stale_issue25.py (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All existing tests updated from `results, _` to `results, *_` to handle the new 3-tuple return. This is backward-compatible and clean. The tests that explicitly unpack `upgraded` use `results, upgraded, *_`. No test logic changes — purely adapting to the new return type.

### Self-Review
**Limitations:** Could not see the full `_pg_dispatch` implementation to verify exactly how kwargs are forwarded to PgApi methods. Could not verify whether there are CLI-level test patterns elsewhere that would suggest CLI tests are expected. Did not run the test suite to confirm all tests pass.

### Feature Requests
- Include the full body of `_pg_dispatch` when a function delegates to it — it's a critical integration point
- Show the full PgApi method when the corresponding check_stale function is modified, to catch signature mismatches like the one found here

## gemini [CONCERN]

### reasons_lib/api.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

The implementation of `check_stale` fails to pass the new `git_aware` parameter to the PostgreSQL dispatch path (lines 1633-1634), meaning the feature is silently ignored when using a PG backend. Additionally, the docstring (lines 1626-1631) was not updated to include `sha_bumped` in the return dictionary description. The new `pin_sources` and `pin_update` functions are correctly implemented for SQLite but lack PostgreSQL support, which is correctly enforced at the CLI level but leaves an architectural gap.

### reasons_lib/check_stale.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `pin_source_url` function (lines 228-232) uses a regex `[^/]+` to match the branch name in GitHub URLs. This logic is broken for branch names containing slashes (e.g., `feature/fix`), as it will only match the first part of the branch name and treat the remainder as part of the file path. Additionally, the `subprocess.run` calls in `get_file_commit_sha` and `file_changed_since` lack timeouts, which could lead to CLI hangs in environments with slow I/O or large Git repositories.

### reasons_lib/pg.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** MISSING

This file was not updated to support the new `git_aware` functionality or return the `sha_bumped` count. While the existing code remains functional, it creates a feature disparity between SQLite and PostgreSQL backends that isn't clearly documented in the API or CLI.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The CLI changes correctly expose the new functionality and provide appropriate feedback to the user. It correctly restricts the new `pin-*` commands to SQLite backends via `_require_sqlite`, preventing undefined behavior on PostgreSQL.

### tests/test_pin_sources.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

This is a high-quality test suite that provides comprehensive coverage for the new git-aware logic, including edge cases like uncommitted changes, auto-bumping of SHAs when content is unchanged, and URL rewriting.

### Self-Review
**Limitations:** I was unable to verify if any external tools or plugins rely on the specific return structure of the `api.check_stale` function, though the internal callers were all updated. The regex concern for GitHub URLs was verified via mental trace and is a known edge case for the implementation provided.

### Feature Requests
- Include a way to verify PostgreSQL integration even if the implementation is primarily focused on SQLite.
- Automate detection of signature mismatches between API dispatchers and their backend implementations (e.g., `api.py` vs `pg.py`).

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#175
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [BLOCK]

### reasons_lib/check_stale.py:get_file_commit_sha
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Uses `filepath.name` with `cwd=filepath.parent`, which correctly scopes the git query to the file's directory. Returns None on failure (non-zero return or empty output). `timeout=30` is reasonable. Tests cover tracked, untracked, non-git, and post-commit scenarios.

### reasons_lib/check_stale.py:file_changed_since
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Checks both committed changes (`git log SHA..HEAD`) and uncommitted changes (`git diff --quiet HEAD`). Returns True (fail-open) on any git error or timeout, which is the safe default — it forces fallback to content hashing. Tests cover no-change, committed-change, and changes-to-other-files scenarios. Missing test for uncommitted (staged/unstaged) changes and `TimeoutExpired`, but the fail-open design makes these low-risk.

### reasons_lib/check_stale.py:pin_source_url
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Handles slashed branch names by accepting an optional `source_path` parameter. Falls back to splitting on first `/` when no `source_path` is provided. Returns the URL unchanged for non-GitHub URLs. Tests cover branch rewriting, non-GitHub URLs, slashed branches, and arbitrary branch names.

### reasons_lib/check_stale.py:pin_sources
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Line 279 calls `node.metadata.get("pinned_sha")` without a None guard, while line 282 does `node.metadata.get("agent") if node.metadata else None`. In practice `metadata` is always `{}` (dataclass default + storage default), so this won't crash, but the inconsistency is suspicious — if you trust `metadata` is never None, both guards are unnecessary; if you don't, line 279 is a latent bug. Minor issue, not blocking.

### reasons_lib/check_stale.py:pin_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good error handling with distinct error strings for each failure mode. Refreshes both `pinned_sha` and `source_hash` on update. Tests cover all error paths (not found, no source, source not found) and the happy path including hash refresh.

### reasons_lib/check_stale.py:check_stale (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The git-aware fast path at lines 105-108 correctly skips content hashing when `file_changed_since` returns False. The SHA auto-bump at lines 134-139 correctly triggers only when content hash matches but `pinned_sha` is stale (file was touched by commits but content reverted). Return type cleanly extended from 2-tuple to 3-tuple. All existing callers updated to `*_` unpacking.

### reasons_lib/api.py:check_stale (modified)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

`needs_write = upgrade_hashes or git_aware` is correct — git_aware can bump `pinned_sha` metadata. The new `git_aware` kwarg is forwarded to `_pg_dispatch` which passes it to `PgApi.check_stale`, and that method accepts it but ignores it (returns `sha_bumped: 0`). However, the `test_pg_dispatch.py` tests at lines 401 and 414 use `assert_called_once_with` without the new `git_aware` kwarg, so those tests will now **fail**. This is a test breakage introduced by this PR.

### reasons_lib/api.py:pin_sources (new)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Clean API wrapper following the established pattern (resolve db_dir, load network, delegate to check_stale module, return results). No PG dispatch — `_require_sqlite` in CLI blocks PG usage. No direct API-level tests (only CLI/unit tests), but the function is thin enough that the unit tests on `check_stale.pin_sources` provide effective coverage.

### reasons_lib/api.py:pin_update (new)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Same clean wrapper pattern. Correctly counts errors vs successes for the return dict. No API-level tests but underlying `check_stale.pin_update` is well tested.

### reasons_lib/pg.py:check_stale (modified)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

Accepts `git_aware=False` parameter but silently ignores it — always returns `sha_bumped: 0`. This is a reasonable stub for now (git operations on a PG backend are harder), but there's no warning or documentation that the flag is a no-op on PG. If a user passes `--git` with `--pg`, they'd silently get non-git-aware behavior. Should at minimum log a warning, or the CLI should block `--git` with `--pg` the way it blocks `pin-sources` with `--pg`.

### reasons_lib/cli.py (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

New subparsers `pin-sources` and `pin-update` are correctly wired with argparse and the dispatch dict. `pin-sources` correctly uses `_require_sqlite`. `pin-update` correctly uses `_require_sqlite`. The `sha_bumped` output in `cmd_check_stale` is clean. `cmd_pin_sources` display logic is correct — `re` variable name shadows the `re` module import in `check_stale.py` but not in `cli.py`, so no conflict. No CLI-level integration tests, but the functions are thin display wrappers.

### tests/test_check_stale.py (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All existing unpacking changed from `results, _` to `results, *_` to accommodate the new 3-tuple return. This is a robust pattern that won't break if additional return values are added later. All assertions remain valid — they only check `results` and optionally `upgraded`.

### tests/test_check_stale_issue25.py (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same mechanical `_, *_` unpacking update. All assertions remain valid.

### tests/test_pin_sources.py (new)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Thorough test suite covering all new functions. The `git_repo` fixture creates a real temporary git repo, which is the right approach for testing git operations. Tests cover: SHA retrieval, file change detection, URL rewriting, pinning (new/skip/force/out/no-source/nonexistent/URL rewrite), updating (happy/errors/hash refresh), and git-aware staleness (fresh/stale/auto-bump/fallback/flag-off). Missing: test for `file_changed_since` with uncommitted changes, and test for `TimeoutExpired` in `file_changed_since`.

### tests/test_pg_dispatch.py (NOT modified — but should be)
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

Lines 401 and 414 assert `mock_pg.check_stale.assert_called_once_with(repos=None, upgrade_hashes=False)` and `assert_called_once_with(repos=None, upgrade_hashes=True)` respectively. Since `api.check_stale` now passes `git_aware=False` through `_pg_dispatch` to the PG backend, these assertions are missing the `git_aware=False` kwarg and will fail. This test file was not updated in the diff.

### Self-Review
**Limitations:** Could not execute the test suite to confirm the `test_pg_dispatch.py` breakage — inferred from static analysis of `_pg_dispatch` forwarding `**kwargs` and the assertion expecting an exact match without `git_aware`. Could not verify whether the `build/` directory copies are used by anything or just stale build artifacts.

### Feature Requests
- Auto-run the test suite and include pass/fail results alongside the diff review
- Flag test files that import modified functions but are not included in the diff — these are prime candidates for breakage

## gemini [CONCERN]

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

API signatures for `check_stale` were updated correctly to include the `git_aware` flag. New API methods `pin_sources` and `pin_update` are well-structured and use the standard context manager pattern for database access. Integration with Postgres via `_pg_dispatch` is correctly handled.

### reasons_lib/check_stale.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The core logic for git-aware staleness checking is sound. `file_changed_since` correctly checks both committed and uncommitted (staged/unstaged) changes using `git log` and `git diff`. `pin_source_url` handles GitHub blob URLs correctly, including the edge case of branch names containing slashes. The `sha_bumped` logic is an elegant way to keep metadata fresh when content hasn't changed but the HEAD has moved.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New CLI commands `pin-sources` and `pin-update` are correctly implemented and registered. The `check-stale --git` flag is appropriately wired through to the API. Use of `_require_sqlite` for the new pinning commands is consistent with existing project patterns for features not yet ported to Postgres.

### reasons_lib/pg.py:check_stale
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

While the function signature was updated to accept `git_aware=False`, the implementation ignores the flag and performs standard content hashing. It returns `sha_bumped: 0` regardless of the input. This doesn't break existing functionality, but it means the "git-aware" optimization is currently unavailable for Postgres backends.

### tests/test_pin_sources.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

A comprehensive new test suite that effectively uses a temporary git repository fixture to verify all aspects of the pinning logic, including edge cases like uncommitted changes and source URL rewriting.

### tests/test_check_stale.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Existing tests were correctly updated to handle the change in the return type of `check_stale` (from a 2-tuple to a 3-tuple). The use of `*_` for unpacking is a robust way to handle the additional return value without breaking test readability.

### Self-Review
**Limitations:** I identified that the Postgres implementation of `check_stale` is lagging behind the SQLite/Network implementation, which is a known architectural pattern in this project but results in a "PARTIAL" integration verdict. Spawning a `git` subprocess per node in `check_stale` may present performance issues for networks with thousands of nodes, but it is correct and sufficient for the current scale.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#175
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/check_stale.py:get_file_commit_sha
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Uses `filepath.name` (bare filename) instead of `filepath` (full path) in the git command. This means `git log -1 --format=%H -- source.md` runs from `filepath.parent`. This works correctly *only if* the file is directly in the repo root or the `cwd` is set to the file's parent directory, which it is — so this is functionally correct in the current usage. However, if the file is in a subdirectory of the repo (e.g. `repo/entries/topic.md`), `cwd` would be `repo/entries/` and `git log -- topic.md` would work. But if the git repo root is `repo/` and not `repo/entries/`, git still searches from the repo root with relative path resolution, so this is fine. The `subprocess.TimeoutExpired` exception is caught in `file_changed_since` but **not** in `get_file_commit_sha` — a timeout there would propagate to the caller, though the 30s timeout makes this unlikely in practice.

### reasons_lib/check_stale.py:file_changed_since
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Fail-open on errors (returns `True` to force content hashing fallback) is the correct design. Catches `TimeoutExpired`. Checks both committed and uncommitted changes. Uses `filepath.name` consistently with `get_file_commit_sha`.

### reasons_lib/check_stale.py:check_stale (modified)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The git-aware skip path (lines 105-106) and the sha-bump path (lines 134-139) have an interaction worth noting. When `git_aware=True` and `pinned_sha` is set: if `file_changed_since` returns `False`, the node is skipped entirely (no hash check, no stale report). If it returns `True`, the code falls through to content hashing. If content hash matches, the `elif` block bumps `pinned_sha` to current HEAD. This is logically sound. One edge case: if the node has `pinned_sha` but `git_aware=False`, the sha-bump `elif` block at line 134 checks `git_aware and pinned_sha` — so it correctly does nothing. The `sha_bumped` branch calls `get_file_commit_sha` which can return `None` (if the file was somehow removed from git tracking), guarded by the `if new_sha` check. Looks correct.

### reasons_lib/check_stale.py:pin_source_url
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The regex `r'(https://github\.com/[^/]+/[^/]+/blob/).+'` matches any GitHub blob URL. When `source_path` is not provided and the branch name contains slashes (e.g. `feature/fix`), the `rest.split("/", 1)` will split on the first slash of the branch name, treating only the first segment as the branch — this produces an incorrect path. The `source_path` parameter handles this case correctly by matching from the end. The fallback without `source_path` is documented to be best-effort. Tests cover both cases. Acceptable tradeoff but the no-`source_path` path is silently wrong for slashed branches.

### reasons_lib/check_stale.py:pin_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly skips OUT nodes, nodes without source, and already-pinned nodes (unless `force=True`). Backfills `source_hash` when missing. URL pinning is gated behind `pin_urls` flag. The `node.metadata.get("pinned_sha")` guard on line 266 is redundant when metadata is always a dict, but harmless.

### reasons_lib/check_stale.py:pin_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Error handling for missing nodes, nodes without source, unresolvable paths, and non-git files is thorough. Each error case produces a result dict with an `"error"` key. Refreshes both `pinned_sha` and `source_hash`, which is the right behavior for a "bump to current" operation.

### reasons_lib/api.py:check_stale (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `git_aware` parameter is threaded through to `_pg_dispatch` and to the inner `_check` call. `needs_write = upgrade_hashes or git_aware` correctly opens the network for writing when `git_aware` is set (since sha-bumping mutates node metadata). Return dict now includes `sha_bumped`.

### reasons_lib/api.py:pin_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Follows the same pattern as `hash_sources` — resolves repos, opens network with `write=True`, delegates to inner function. No PG dispatch path, which is consistent with `_require_sqlite` guard in the CLI. Tests exist for the inner `pin_sources` function but not for the `api.pin_sources` wrapper specifically — coverage is indirect.

### reasons_lib/api.py:pin_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Same pattern as `pin_sources`. Correctly counts errors separately from successes. No PG dispatch, guarded by `_require_sqlite` in CLI. Same note about api-level wrapper lacking direct tests.

### reasons_lib/pg.py:check_stale (modified)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `git_aware` parameter is accepted but completely ignored — `sha_bumped` is hardcoded to `0`. This is a stub. It's not incorrect per se (the feature just doesn't work with PG), but the function silently accepts `git_aware=True` and does nothing with it. A user calling `check_stale(git_aware=True, pg_conninfo=...)` would get no error and no benefit. Should either raise/warn or be documented as a no-op. The PG path doesn't have `pin_sources`/`pin_update` at all, which is at least guarded by `_require_sqlite` in the CLI.

### reasons_lib/cli.py:cmd_pin_sources
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly wired to argparse with `--force` and `--pin-urls` flags. Guarded by `_require_sqlite`. Output formatting is clear. The `re` variable shadows the builtin `re` module but it's a local scope so no issue in practice — though it's a naming collision with the `re` import at the module level of `check_stale.py`, not this file. No CLI-level tests but the command is straightforward dispatch.

### reasons_lib/cli.py:cmd_pin_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly wired. Uses positional `node_ids` arg. Handles error display. No CLI-level tests.

### reasons_lib/cli.py:cmd_check_stale (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Uses `getattr(args, "git", False)` which is safe. Prints sha-bump count when nonzero. Clean integration.

### tests/test_pin_sources.py (new file)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test coverage with a real git repo fixture. Tests cover: `get_file_commit_sha` (tracked, untracked, non-git, changed), `file_changed_since` (no change, change, other file), `pin_source_url` (branch, non-github, slashed branch), `pin_sources` (pin, backfill, skip pinned, force, skip OUT, skip no-source, skip missing, URL rewrite), `pin_update` (update, missing node, no source, source not found, hash refresh), and `check_stale` with `git_aware` (fresh, stale, sha-bump, fallback, disabled). Good edge case coverage.

### tests/test_check_stale.py (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All unpacking changed from `results, _` to `results, *_` to accommodate the new third return value. This is a clean, minimal adaptation. Tests that explicitly check `upgraded` use `results, upgraded, *_`. Existing assertions are preserved and still valid.

### tests/test_check_stale_issue25.py (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same `*_` unpacking pattern. All assertions unchanged and still valid.

### tests/test_pg_dispatch.py (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Mock assertions updated to include `git_aware=False` in the expected call. Matches the new signature correctly.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Did not inspect the `build/` directory copies for staleness against the source (the observation data shows they're out of date, missing `git_aware` in `build/lib/reasons_lib/pg.py`). Could not verify the `add_node` signature accepts `source_url` keyword arg (used in `pin_urls` test) — inferred from the Node dataclass having a `source_url` field.

### Feature Requests
- Include the full function body (not just diff hunks) for every modified function in the diff, without needing a separate observation request
- Flag when `build/` or `dist/` copies of source files are stale relative to the source — this review caught it only because the observation data surfaced it

## gemini [PASS]

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `check_stale` function correctly handles the new `git_aware` flag and dispatches it to the PostgreSQL backend if applicable. The new `pin_sources` and `pin_update` functions are well-integrated into the API layer, following the established pattern of using `_with_network` for state management and `_pg_dispatch` (where omitted, correctly handled by CLI-level checks).

### reasons_lib/check_stale.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation of `git_aware` staleness checking is robust. It uses `git log` and `git diff` to detect changes since a pinned SHA, with a safe fallback to content hashing if git commands fail or time out. The auto-bumping logic for `pinned_sha` when content is unchanged but the commit has moved is a smart optimization. The URL pinning logic handles branch names with slashes correctly when `source_path` is provided.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New subcommands `pin-sources` and `pin-update` are correctly added to the CLI. The `--git` flag for `check-stale` is properly wired. The use of `_require_sqlite` for the new pinning commands is appropriate since they are not yet implemented for the PostgreSQL backend.

### reasons_lib/pg.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `check_stale` signature was updated to include `git_aware` for compatibility with the API dispatcher. While it currently returns `sha_bumped: 0` (effectively a stub for the git-aware logic in PG), it maintains system integrity and avoids breaking existing PG-based workflows.

### tests/test_pin_sources.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

This new test file provides comprehensive coverage for the git-aware features, including edge cases like untracked files, changes to other files, and URL rewriting. It uses a temporary git repository fixture to ensure isolation and reproducibility.

### tests/test_check_stale.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Existing tests were correctly updated to handle the modified return type of `check_stale` (now a 3-tuple). The use of `*_` to unpack remaining values ensures backward compatibility for these tests.

### Self-Review
**Limitations:** I verified the `Node` metadata handling by observing how `pinned_sha` and `verified_at` are accessed/stored in `check_stale.py`, but I rely on the assumption that the `Storage` layer correctly persists these arbitrary metadata fields. Given existing tests pass, this assumption seems sound.

PR #177, opened 2026-06-04
Add propose-update command for structured belief amendments

### LINKED ISSUE #171: Add reasons propose-update command for belief amendments
A structured format for proposing belief amendments, designed to drop into a PR-style review flow.

## Proposed output format

1. Target node ID and current text
2. Proposed new text (or retract)
3. Failure mode by taxonomy: smuggled premise, unsupported superlative, false causal link, domain conflation, contradicted by source, stale
4. Update basis: source-divergence (mechanically verifiable, cheap to accept) / detected-contradiction (nogood) / prior-knowledge (needs scrutiny, highest risk of shared blind spot)
5. Evidence with reachable URL where available
6. Cascade section: list of dependents that need re-review, with prediction of which survive and which fall

## Why the basis field matters

Source-divergence proposals are cheap to accept -- the reviewer can check the same link. Prior-knowledge proposals deserve scrutiny because the proposer and reviewer may share the same blind spot. The field lets reviewers triage instantly.

## Usage

Could be invoked by an LLM agent reviewing beliefs, or by a human proposing changes. Output is a markdown file structured as a PR body: diff at the top, blast radius at the bottom.

## Origin

Opus 4.8 review of expert.ftl2.com -- designed the format while discussing how Claude web could propose belief updates via markdown files.

benthomasson:
## Summary

- Adds `reasons propose-update` command that sends beliefs to an LLM for evaluation and generates structured markdown proposals for updates or retractions
- Proposals include failure mode taxonomy (smuggled-premise, unsupported-superlative, false-causal-link, domain-conflation, contradicted-by-source, stale) and update basis (source-divergence, detected-contradiction, prior-knowledge) for reviewer triage
- Cascade impact computed via existing `what_if_retract()` infrastructure

Closes #171

## Test plan

- [x] 30 unit tests covering format, parse, markdown rendering, batching, error handling
- [ ] Integration test with live LLM: `uv run reasons propose-update --stale-only`
- [ ] Verify cascade section matches `reasons what-if retract` output

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2540,7 +2540,7 @@ def review_beliefs(
 def propose_update(
     belief_ids: list[str] | None = None,
     model: str = "claude",
-    timeout: int = 300,
+    timeout: int = 600,
     stale_only: bool = False,
     namespace: str | None = None,
     sample: int | None = None,
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2537,6 +2537,78 @@ def review_beliefs(
     }
 
 
+def propose_update(
+    belief_ids: list[str] | None = None,
+    model: str = "claude",
+    timeout: int = 300,
+    stale_only: bool = False,
+    namespace: str | None = None,
+    sample: int | None = None,
+    on_batch: Callable | None = None,
+    db_path: str = DEFAULT_DB,
+) -> dict:
+    """Propose structured updates or retractions for beliefs.
+
+    Sends beliefs to an LLM for evaluation, classifying each by failure
+    mode and update basis. Computes cascade impact for retract proposals.
+
+    Returns: {"proposals": [...], "reviewed": int, "timestamp": str}
+    """
+    from datetime import datetime
+
+    from .propose_update import propose_updates as _propose
+
+    result = export_network(db_path=db_path)
+    nodes = result.get("nodes", {})
+
+    candidates = {
+        k: v for k, v in nodes.items()
+        if v.get("truth_value") == "IN"
+    }
+
+    if belief_ids:
+        candidates = {k: v for k, v in candidates.items() if k in belief_ids}
+
+    if stale_only:
+        candidates = {
+            k: v for k, v in candidates.items()
+            if (v.get("metadata") or {}).get("stale_reason")
+        }
+
+    if namespace is not None:
+        if namespace == "":
+            candidates = {k: v for k, v in candidates.items() if ":" not in k}
+        else:
+            candidates = {
+                k: v for k, v in candidates.items()
+                if k.startswith(f"{namespace}:")
+            }
+
+    if sample is not None and len(candidates) > sample:
+        import random
+        sampled_keys = random.sample(sorted(candidates.keys()), sample)
+        candidates = {k: candidates[k] for k in sampled_keys}
+
+    review_ids = sorted(candidates.keys())
+    proposals = _propose(nodes, belief_ids=review_ids, model=model,
+                         timeout=timeout, on_batch=on_batch)
+
+    cascades = {}
+    for p in proposals:
+        try:
+            cascades[p["id"]] = what_if_retract(p["id"], db_path=db_path)
+        except KeyError:
+            pass
+
+    now = datetime.now().isoformat(timespec="seconds")
+    return {
+        "proposals": proposals,
+        "cascades": cascades,
+        "reviewed": len(review_ids),
+        "timestamp": now,
+    }
+
+
 def repair_smuggled(
     review_file: str | None = None,
     belief_ids: list[str] | None = None,
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2590,11 +2590,20 @@ def propose_update(
         candidates = {k: candidates[k] for k in sampled_keys}
 
     review_ids = sorted(candidates.keys())
+
+    # Enrich nodes with dependents (export_network doesn't include them)
+    with _with_network(db_path) as net:
+        for nid in nodes:
+            if nid in net.nodes:
+                nodes[nid]["dependents"] = sorted(net.nodes[nid].dependents)
+
     proposals = _propose(nodes, belief_ids=review_ids, model=model,
                          timeout=timeout, on_batch=on_batch)
 
     cascades = {}
     for p in proposals:
+        if p["id"] not in nodes:
+            continue
         try:
             cascades[p["id"]] = what_if_retract(p["id"], db_path=db_path)
         except KeyError:
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -1575,20 +1575,15 @@ def cmd_propose_update(args):
 
     print(f"\nReviewed {result['reviewed']} beliefs, {len(proposals)} update(s) proposed")
 
-    nodes = None
-    if output_format == "markdown" or args.output:
+    if output_format == "markdown":
         net_result = api.export_network(db_path=args.db)
         nodes = net_result.get("nodes", {})
-
-    output_file = args.output or "proposed-updates.md"
-    if output_format == "json":
-        output_file = args.output or "proposed-updates.json"
-
-    if output_format == "markdown":
+        output_file = args.output or "proposed-updates.md"
         content = format_proposals_file(proposals, nodes=nodes, cascades=cascades)
         Path(output_file).write_text(content)
         print(f"\nWrote proposals to {output_file}")
     else:
+        output_file = args.output or "proposed-updates.json"
         Path(output_file).write_text(_json.dumps({
             "timestamp": ts,
             "model": model,
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -1520,6 +1520,95 @@ def cmd_review_beliefs(args):
                 print(f"  ERROR retracting {r['id']}: {e}", file=sys.stderr)
 
 
+def cmd_propose_update(args):
+    _require_sqlite(args, "propose-update")
+    import json as _json
+    from datetime import datetime
+
+    from .propose_update import format_proposals_file
+
+    model = getattr(args, "model", None) or "claude"
+    ts = datetime.now().isoformat(timespec="seconds")
+    write_report = not args.no_report
+    output_format = getattr(args, "format", "markdown") or "markdown"
+
+    report_path = None
+    if write_report:
+        report_dir = Path(args.report_dir)
+        report_dir.mkdir(parents=True, exist_ok=True)
+        report_path = report_dir / f"propose-update-{ts.replace(':', '')}.json"
+
+    def on_batch(results):
+        if report_path is not None:
+            report_path.write_text(_json.dumps({
+                "timestamp": ts,
+                "status": "partial",
+                "model": model,
+                "proposals": results,
+            }, indent=2))
+
+    result = api.propose_update(
+        belief_ids=args.ids or None,
+        model=model,
+        timeout=args.timeout,
+        stale_only=args.stale_only,
+        namespace=args.namespace,
+        sample=args.sample,
+        on_batch=on_batch,
+        db_path=args.db,
+    )
+
+    proposals = result["proposals"]
+    cascades = result.get("cascades", {})
+
+    if not proposals:
+        print("No updates proposed.")
+        return
+
+    for p in proposals:
+        action = p["action"].upper()
+        basis = p.get("basis", "")
+        fm = p.get("failure_mode", "")
+        print(f"  [{action}] {p['id']}  ({fm}, {basis})")
+        if p.get("comment"):
+            print(f"    {p['comment']}")
+
+    print(f"\nReviewed {result['reviewed']} beliefs, {len(proposals)} update(s) proposed")
+
+    nodes = None
+    if output_format == "markdown" or args.output:
+        net_result = api.export_network(db_path=args.db)
+        nodes = net_result.get("nodes", {})
+
+    output_file = args.output or "proposed-updates.md"
+    if output_format == "json":
+        output_file = args.output or "proposed-updates.json"
+
+    if output_format == "markdown":
+        content = format_proposals_file(proposals, nodes=nodes, cascades=cascades)
+        Path(output_file).write_text(content)
+        print(f"\nWrote proposals to {output_file}")
+    else:
+        Path(output_file).write_text(_json.dumps({
+            "timestamp": ts,
+            "model": model,
+            "reviewed": result["reviewed"],
+            "proposals": proposals,
+            "cascades": {k: v for k, v in cascades.items()},
+        }, indent=2))
+        print(f"\nWrote proposals to {output_file}")
+
+    if write_report and report_path is not None:
+        report_path.write_text(_json.dumps({
+            "timestamp": ts,
+            "status": "complete",
+            "model": model,
+            "reviewed": result["reviewed"],
+            "proposals": proposals,
+        }, indent=2))
+        print(f"  Report: {report_path}")
+
+
 def cmd_repair_smuggled(args):
     _require_sqlite(args, "repair-smuggled")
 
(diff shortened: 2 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#177
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## Disagreements

- **tests/test_propose_update.py** [MEDIUM]: claude: CONCERN, gemini: BLOCK
- **reasons_lib/propose_update.py** [LOW]: claude: CONCERN, gemini: PASS
- **reasons_lib/api.py:propose_update** [LOW]: claude: CONCERN, gemini: PASS
- **reasons_lib/cli.py:cmd_propose_update** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### reasons_lib/propose_update.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

**Bug — `dependents` field never populated (line ~130-145).** `format_belief_for_update` reads `node.get("dependents", [])` and builds a "Dependents (N):" section. But `export_network()` — the source of the `nodes` dict — does not include `dependents` in its output. This means the dependent information is **silently omitted** from every LLM prompt. The code and tests work without errors because `.get("dependents", [])` defaults to empty, but the feature as designed is degraded — the LLM won't see cascade implications when deciding whether to propose retractions.

The test fixture `sample_nodes` manually includes `"dependents": ["derived-b"]`, so the test passes, but this diverges from real data flow. The test is testing a path that production never exercises.

**Fix:** Either add `"dependents": sorted(n.dependents)` to `export_network`'s node dict, or compute dependents in `propose_update` by scanning justification antecedents.

**Minor — `parse_update_proposals` returns `[]` for valid-but-empty LLM proposals via the fallthrough path** rather than the `if results:` branch. The behavior is correct but slightly fragile: if a future LLM response contains `[{"unrelated": "data"}]` (no `id` key) followed by a valid array, the first array's empty-results path falls through and the scan continues, which is actually good recovery behavior. Not a bug, just worth noting the control flow.

**FAILURE_MODES set is defined but never validated.** `parse_update_proposals` validates `basis` (defaults to `"prior-knowledge"` on invalid values) and `action` (defaults to `"update"`), but `failure_mode` is passed through unchecked. If the LLM returns a non-standard failure mode string, it'll persist into the output. Consider validating it the same way `basis` is validated.

### reasons_lib/api.py:propose_update
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

**Cascade computation for "update" actions is unnecessary.** The loop at lines 2594-2598 calls `what_if_retract` for **every** proposal, including "update" proposals. An "update" doesn't retract the node — it changes its text. Computing retraction cascades for updates is misleading and wastes DB reads. Should only compute cascades for `action == "retract"` proposals.

**No tests for `api.propose_update` itself.** The test file covers `reasons_lib.propose_update.propose_updates` (the inner module) but not `reasons_lib.api.propose_update` (the outer API function). The filtering logic (stale_only, namespace, sample) in the API layer is untested. This is the same pattern as `review_beliefs` so it's consistent with the codebase, but it means the namespace filter, stale-only filter, and sample logic are unverified.

**`what_if_retract` KeyError silently swallowed (line 2596-2598).** If the LLM hallucinates a belief ID that doesn't exist in the DB, the `except KeyError: pass` silently ignores it. This is fine for cascade computation specifically, but the hallucinated ID will still appear in the proposals output. No validation that `p["id"]` actually exists in `nodes`.

### reasons_lib/cli.py:cmd_propose_update
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

**`--format` argument name collides with Python builtin.** `getattr(args, "format", "markdown")` works but `format` is a Python builtin. argparse handles this correctly so it's not broken, but it's a minor readability concern (the existing `review-beliefs` command doesn't have this issue).

**Unconditional markdown file write.** When `output_format == "markdown"` (the default), the code always writes to `proposed-updates.md` or the `-o` path without asking. For a command that "proposes" changes (not applies them), silently writing a file on every run is mildly surprising — but follows the pattern of other commands in this CLI.

**`nodes` variable computed conditionally but could be None.** At line 1580, `nodes = None` initially. Then at line 1581: `if output_format == "markdown" or args.output:` — if neither is true (e.g., `--format json` with no `-o` flag), nodes stays `None`, but then the `else` branch at line 1590 writes JSON without needing `nodes`, so this is fine in practice. However, the condition `args.output` is redundant with `output_format == "markdown"` in the `if` block since the default `output_file` is set regardless.

**No CLI-level tests.** The CLI handler has no integration tests. This is consistent with the codebase pattern (other CLI commands also lack CLI-level tests), but means the argparse wiring, report writing, and output formatting are unverified.

### reasons_lib/cli.py (argparse registration)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Parser setup is correct and consistent with other subcommands. Default timeout of 600 in argparse differs from the 300 default in `api.propose_update`, but argparse's value takes precedence at the CLI level, so the API default doesn't matter for CLI invocations. The command dispatch at line 2326 correctly maps `"propose-update"` to `cmd_propose_update`.

### tests/test_propose_update.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Good coverage of the core module functions. The test fixture includes `"dependents"` keys that real `export_network` data does not contain (see main concern above), so `TestFormatBeliefForUpdate` tests a code path that production never exercises.

**Missing coverage:**
- `api.propose_update` is not tested at all (filtering, cascade computation, integration with `what_if_retract`)
- No test for multiple justifications (the `len(justs) > 1` branch in `format_belief_for_update`)
- No test that `parse_update_proposals` handles a response with multiple `[` characters where the first is not the JSON array
- No CLI-level tests

**Good coverage of:**
- All `format_belief_for_update` variants (premise, derived, outlist, stale, missing)
- `parse_update_proposals` edge cases (malformed, prose-wrapped, invalid fields, missing id)
- `propose_updates` batching, error handling, callbacks, empty/nonexistent IDs
- Markdown formatting with/without cascades

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Could not verify the `review_beliefs` implementation's handling of similar patterns to confirm consistency claims. Did not have the full `cli.py` to verify no argparse conflicts or ordering issues with the new subparser.

### Feature Requests
- Include the return type/shape of functions called by the changed code (e.g., `export_network`'s actual output schema) as an automatic observation when the changed code accesses specific keys from their results
- Flag test fixtures whose shape diverges from the actual production data (e.g., fixture includes keys that the real data source doesn't produce)

## gemini [BLOCK]

### reasons_lib/propose_update.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The core implementation of the LLM-driven update proposal logic is sound. The prompt is well-structured and uses a clear taxonomy for failure modes and bases. The parser is robust, using `raw_decode` to surgically extract the JSON array from LLM responses that might include conversational preamble or postamble. The markdown formatting provides clear, actionable `reasons` commands for the user to apply the suggested changes.

### reasons_lib/api.py:propose_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The API integration correctly bridges the network state with the proposal logic. The candidate filtering logic (IDs, stale-only, namespace, sampling) is consistent with existing patterns in the codebase (e.g., `review_beliefs`). A significant value-add is the inclusion of `what_if_retract` calls for each proposal, which allows the return of cascade impact data so the user can understand the consequences of a retraction before accepting it.

### reasons_lib/cli.py:cmd_propose_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The CLI command is correctly wired and follows the project's established patterns for report generation and output formatting. The use of an `on_batch` callback to write partial JSON reports is an excellent touch for long-running LLM batch processes, ensuring data is not lost if the process is interrupted. The command also correctly requires SQLite, as the PG dispatch for this specific complex flow hasn't been implemented yet.

### tests/test_propose_update.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The test file is fundamentally broken and will fail to execute. All instances of the `@patch` decorator (lines 326, 347, 356, 366, 374, 383) have been replaced with the mangled text `@tests/test_pg_dispatch.py(...)`. This is likely the result of an incorrect automated search-and-replace during development. Since `tests/test_pg_dispatch.py` is a file path and not a valid Python identifier or callable decorator in this context, the file will raise a `SyntaxError` or `NameError` immediately upon discovery by `pytest`.

### Self-Review
**Limitations:** I could not verify if `Path` is imported at the top of `reasons_lib/cli.py`. While it is used in `cmd_propose_update`, it is not imported locally like `json` and `datetime` are. If it is not a global import in that file, the command will fail. However, the broken decorators in the test file are a definitive blocker that takes precedence.

### Feature Requests
- It would be helpful to see the top-level imports of a file when reviewing new functions that use common libraries like `pathlib` or `typing`, to ensure they don't need to be added.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#177
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## Disagreements

- **tests/test_propose_update.py** [HIGH]: claude: PASS, gemini: BLOCK
- **reasons_lib/api.py:propose_update** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### reasons_lib/propose_update.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Well-structured module following the same pattern as the existing `review.py`. The `parse_update_proposals` scanner correctly handles prose-wrapped JSON, empty arrays, and invalid fields. The failure-mode and basis taxonomies are validated with fallback defaults. Batching logic in `propose_updates` is correct — error handling catches per-batch failures and continues. One minor nit: invalid `failure_mode` defaults to `""` rather than `"unknown"`, which produces `**Failure mode:** ` (trailing blank) in markdown output — cosmetic, not a bug.

### reasons_lib/api.py:propose_update
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Two concerns. (1) **Cascades are computed for all proposals regardless of action type** — `what_if_retract` runs even for `"update"` proposals where the node isn't being retracted. The `format_proposal_markdown` then shows "Cascade impact: N dependent(s) affected — will go OUT" for updates, which is misleading since updating text doesn't retract the node. Consider gating cascades on `p["action"] == "retract"`. (2) **No test coverage** — the API function (`api.propose_update`) is not exercised by any test. The tests only cover the lower-level `propose_updates()` from `propose_update.py`. The filtering logic (stale_only, namespace, sample), dependents enrichment, and cascade assembly are all untested at the API layer. The dependents enrichment loop (`for nid in nodes`) correctly opens a separate `_with_network` context, and the mutation of the `nodes` dict is safe since `export_network` returns a fresh dict.

### reasons_lib/cli.py:cmd_propose_update
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

CLI handler follows the established pattern (see `cmd_review_beliefs`). The subparser is registered, the dispatch table entry is added. Two observations: (1) **Timeout default mismatch** — CLI argparser defaults to `600` while the API function signature defaults to `300`. Not a runtime bug since the CLI always passes its value explicitly, but the inconsistency could surprise someone reading the API docs. (2) **No CLI-level tests** — `cmd_propose_update` is untested. The `on_batch` callback that writes partial JSON reports is untested. The format-switching logic (markdown vs JSON output) is untested.

### reasons_lib/cli.py (argparser registration)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Parser registration is correct. Arguments mirror the API parameters. The dispatch table entry `"propose-update": cmd_propose_update` at line 2321 wires it in. The `_require_sqlite` guard correctly blocks PostgreSQL usage, consistent with other LLM-driven commands.

### tests/test_propose_update.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Strong unit test coverage for the `propose_update.py` module — 30 tests across 6 classes covering formatting, parsing, batching, error handling, edge cases (empty/nonexistent IDs, malformed JSON, invalid field defaults). The mock strategy (`@patch("reasons_lib.propose_update.invoke_model")`) is correct and avoids real LLM calls. The `sample_nodes` fixture is well-designed with premise, derived, outlist, and stale nodes. The test gap is at the API and CLI layers, not here.

### Self-Review
**Limitations:** Could not verify the full `review_beliefs` / `cmd_review_beliefs` patterns beyond the observation excerpts — the review relies on the observation data for structural comparison. Could not run the test suite to confirm all 438 lines of test pass. Did not see the `reasons update` command referenced in the generated `**Command:**` lines — cannot verify it exists (the diff only shows `propose-update`, not an `update` subcommand that would actually apply the proposed text change).

### Feature Requests
- Include the existing sibling module (`review.py`) in observations when a new module clearly parallels it — helps verify pattern consistency
- Flag when a generated CLI command (like `reasons update NODE "text"`) references a subcommand not visible in the diff or dispatch table — potential broken user-facing instruction

## gemini [BLOCK]

### reasons_lib/api.py:propose_update
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation correctly orchestrates the proposal generation process. It mirrors the filtering logic used in `review_beliefs`, enriches nodes with necessary dependency information that `export_network` lacks, and integrates cascade simulation for better user feedback. While the multiple database re-opens (once for enrichment and once per proposal for cascades) could be optimized, it is functional and consistent with the existing API style.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The CLI command is properly integrated into the subparser and successfully handles output formatting (Markdown/JSON) and reporting. The use of `on_batch` for partial report generation is a good practice for long-running LLM tasks.

### reasons_lib/propose_update.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The core logic for formatting LLM prompts and parsing responses is robust. The use of `json.JSONDecoder.raw_decode` is an excellent choice for resiliently extracting JSON arrays from potentially conversational LLM output. The batching logic correctly handles progress reporting and partial results.

### tests/test_propose_update.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The tests are broken and will not run. Every test case in the `TestProposeUpdates` class (starting at line 360) uses an invalid Python decorator: `@tests/test_pg_dispatch.py(...)`. This is a syntax error (dots and file extensions are not valid in decorator attribute access) and likely a botched find-and-replace for `@patch`. Since the tests cannot execute, the new feature is effectively untested in this PR.

### Self-Review
**Limitations:** I was able to verify the broken tests directly from the provided diff and the `propose_update_callers` observation which confirmed the invalid decorator string in the file. I have sufficient context on the `Network` and `Node` structures from surrounding function bodies.

### Feature Requests
- **Batch Optimization**: Modify internal API methods (like `what_if_retract`) to optionally accept an existing `Network` object to avoid repeated database I/O when called in loops (e.g., in `propose_update`).
- **Dependency Inclusion**: Update `export_network` to optionally include `dependents` in the node payload to avoid the need for separate enrichment steps in callers.

PR #185, opened 2026-06-08
Add cost/token tracking via --output-format json (#184)

### LINKED ISSUE #184: Add cost/token tracking via --output-format json
Add cost/token tracking to LLM operations by switching CLI invocations to `--output-format json`.

## Problem

LLM-calling operations (derive, review-beliefs, propose-update, ask, research) consume tokens that cost money, but there is no visibility into per-operation or cumulative costs. Users cannot estimate the cost of running a pipeline or compare model costs.

## Approach

Both `claude -p` and `gemini -p` support `--output-format json` which returns structured output including token counts and (for Claude) cost:

**Claude:**
```json
{
  "result": "The LLM response text...",
  "total_cost_usd": 0.217,
  "usage": {
    "input_tokens": 3,
    "output_tokens": 7,
    "cache_creation_input_tokens": 34564,
    "cache_read_input_tokens": 0
  }
}
```

**Gemini:**
```json
{
  "response": "The LLM response text...",
  "stats": {
    "models": {
      "gemini-2.5-flash-lite": {
        "tokens": {"input": 1707, "candidates": 32, "total": 1833}
      }
    }
  }
}
```

### 1. Switch to JSON output in `llm.py`

In `invoke_model()` (~20 lines of change):
- Add `--output-format json` to CLI commands in `MODEL_COMMANDS` and `resolve_model_cmd()`
- Parse the JSON response, extract text from `result` (Claude) or `response` (Gemini)
- Extract token/cost data and accumulate in a module-level tracker

### 2. Cost accumulator

Simple dataclass or dict that tracks per-call and cumulative stats:
- `input_tokens`, `output_tokens`, `total_cost_usd`
- Per-model breakdown
- Accessible via `llm.get_cost_summary()` or similar

### 3. CLI integration

- Print cost summary at the end of multi-call operations (derive, review-beliefs, propose-update)
- Optional `--show-cost` flag or always-on summary line
- Example output: `Cost: $0.42 (12,340 input + 2,100 output tokens across 3 calls)`

## What does NOT change

- The API-based path (`api:`, `vertex:`) already uses LangChain which has its own token tracking via callbacks/Langfuse
- The `ollama:` path has no cost (local inference)
- No new dependencies needed — this uses stdlib `json` parsing on existing CLI output

## Scope

This is a lightweight, ~30-line change to `llm.py` plus minor CLI output additions. No architectural changes, no new dependencies.

benthomasson:
## Summary
- CLI models (claude, gemini) now use --output-format json to capture token counts and costs from wrapper metadata without affecting model responses
- New cost accumulator records per-call and per-model stats across all invoke_model calls
- format_cost_summary() prints a human-readable cost line to stderr after multi-call commands
- JSON parsing handles both Claude and Gemini formats with graceful fallback for non-JSON output

## Test plan
- [x] 51 tests in test_llm.py pass (12 new tests for JSON parsing, cost tracking, and edge cases)
- [x] Full suite: 1328 passed, 0 failed
- [x] Edge case: JSON arrays from mocked responses handled gracefully

Closes #184

### DIFF
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -1191,6 +1191,8 @@ def _write_derive_report(report_state, status):
 def cmd_derive(args):
     _require_sqlite(args, "derive")
     from datetime import datetime
+    from .llm import reset_cost_tracker, format_cost_summary
+    reset_cost_tracker()
 
     prompt_template = None
     if args.prompt_file:
@@ -1287,6 +1289,10 @@ def cmd_derive(args):
     if report_state is not None:
         _write_derive_report(report_state, "complete")
         print(f"  Report: {report_state['report_path']}")
+
+    cost = format_cost_summary()
+    if cost:
+        print(f"  {cost}", file=sys.stderr)
 
 
 def cmd_accept(args):
@@ -1431,6 +1437,8 @@ def cmd_review_beliefs(args):
     _require_sqlite(args, "review-beliefs")
     import json
     from datetime import datetime
+    from .llm import reset_cost_tracker, format_cost_summary
+    reset_cost_tracker()
 
     model = getattr(args, "model", None) or "claude"
     ts = datetime.now().isoformat(timespec="seconds")
@@ -1552,11 +1560,17 @@ def cmd_review_beliefs(args):
             except Exception as e:
                 print(f"  ERROR retracting {r['id']}: {e}", file=sys.stderr)
 
+    cost = format_cost_summary()
+    if cost:
+        print(f"  {cost}", file=sys.stderr)
+
 
 def cmd_propose_update(args):
     _require_sqlite(args, "propose-update")
     import json as _json
     from datetime import datetime
+    from .llm import reset_cost_tracker, format_cost_summary
+    reset_cost_tracker()
 
     from .propose_update import format_proposals_file
 
@@ -1636,6 +1650,10 @@ def cmd_propose_update(args):
         }, indent=2))
         print(f"  Report: {report_path}")
 
+    cost = format_cost_summary()
+    if cost:
+        print(f"  {cost}", file=sys.stderr)
+
 
 def cmd_repair_smuggled(args):
     _require_sqlite(args, "repair-smuggled")
@@ -1681,6 +1699,8 @@ def cmd_repair_smuggled(args):
 
 def cmd_research(args):
     _require_sqlite(args, "research")
+    from .llm import reset_cost_tracker, format_cost_summary
+    reset_cost_tracker()
 
     model = getattr(args, "model", None) or "claude"
     review_file = getattr(args, "review_file", None)
@@ -1722,6 +1742,10 @@ def cmd_research(args):
     if args.dry_run:
         print("\n  (dry run -- no changes applied)")
 
+    cost = format_cost_summary()
+    if cost:
+        print(f"  {cost}", file=sys.stderr)
+
 
 def cmd_contradictions(args):
     _require_sqlite(args, "detect-contradictions")
@@ -1747,6 +1771,9 @@ def cmd_contradictions(args):
         else:
             print("No nogoods to apply.")
         return
+
+    from .llm import reset_cost_tracker, format_cost_summary
+    reset_cost_tracker()
 
     model = getattr(args, "model", None) or "claude"
     output = args.output
@@ -1787,6 +1814,10 @@ def cmd_contradictions(args):
     elif not args.auto_apply:
         print(f"\nWrote {output} — review, then run:")
         print(f"  reasons contradictions --accept {output}")
+
+    cost = format_cost_summary()
+    if cost:
+        print(f"  {cost}", file=sys.stderr)
 
 
 def cmd_namespaces(args):
--- a/reasons_lib/llm.py
+++ b/reasons_lib/llm.py
@@ -6,8 +6,12 @@ and API-based providers via 'api:<model>' and 'vertex:<model>'.
 
 CLI-based models pipe prompts to stdin and read responses from stdout.
 API-based models use LangChain adapters with optional Langfuse tracing.
+
+Cost tracking: CLI models use --output-format json to capture token
+counts and costs. Use get_cost_summary() to retrieve accumulated stats.
 """
 
+import json
 import os
 import shutil
 import subprocess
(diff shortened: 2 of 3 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#185
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/llm.py — Cost tracker module globals & functions
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The module-level `_cost_tracker` dict, `reset_cost_tracker()`, `get_cost_summary()`, `format_cost_summary()`, and `_record_cost()` are straightforward accumulators. Using a mutable module-level dict that gets mutated in-place (rather than reassigned) means `reset_cost_tracker()` correctly resets the shared object. Tests cover reset, accumulation, by-model tracking, and format edge cases.

### reasons_lib/llm.py:_parse_cli_json
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Two issues: (1) The gemini branch guard `model.startswith("gemini") or model.startswith("gemini:")` is redundant — anything starting with `"gemini:"` already starts with `"gemini"`. Not a bug, just dead code in the condition. (2) When the Claude JSON is valid but has no `"result"` key, `data.get("result", output)` falls back to `output` — which is the raw JSON string, not the original text. The test `test_json_missing_result_returns_raw` asserts this exact behavior, so it's intentional, but it means the caller gets a JSON string instead of usable text, which could silently corrupt downstream processing. The fallback to `output` also means `_record_cost` is still called with whatever `usage`/`total_cost_usd` data exists in the dict, recording stats for a "failed" parse. Minor: if the response text itself is valid JSON (e.g., the LLM returns a JSON object as its answer), the outer `json.loads` will succeed and `data.get("result", output)` may return the wrong thing — though this is unlikely given that `--output-format json` wraps the result.

### reasons_lib/llm.py:MODEL_COMMANDS and resolve_model_cmd
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adding `--output-format json` to Claude and `-o json` to Gemini commands is clean. The `resolve_model_cmd` updates for `claude:*` and `gemini:*` submodels are consistent. Ollama is correctly excluded (no JSON output mode). All existing tests updated to match the new command arrays.

### reasons_lib/llm.py:invoke_model
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The ollama early-return before `_parse_cli_json` is correct — ollama doesn't have `--output-format json`. The thinking-marker stripping for ollama still happens before the early return. For Claude/Gemini, `_parse_cli_json` is called on the raw stdout which now contains JSON. The graceful fallback in `_parse_cli_json` means non-JSON output (e.g., from older CLI versions) won't break. API models (`api:`, `vertex:`) bypass CLI entirely so are unaffected.

### reasons_lib/cli.py — Cost tracking integration in cmd_derive, cmd_review_beliefs, cmd_propose_update, cmd_research, cmd_contradictions
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

The pattern (import + reset at top, format + print to stderr at bottom) is applied consistently to 5 commands. However, **3 LLM-calling commands are missing cost tracking**: `cmd_ask` (calls `ask.ask()` → `invoke_model`), `cmd_repair_smuggled` (calls `api.repair_smuggled()` → `invoke_model`), and `cmd_list_negative` (calls `api.list_negative()` → `invoke_model`). This is an incomplete rollout. Additionally, there are no integration tests verifying that cost summaries actually appear on stderr for any command — only unit tests for the formatting functions themselves.

### reasons_lib/cli.py — `_write_report` placement of cost printing
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

In `cmd_review_beliefs`, the cost summary print is placed after the nested `_write_report` function definition block (line 1563), but inside the outer function. The `_write_report` helper at line 1481 writes the report file — the cost data is NOT included in the report JSON, only printed to stderr. This is fine for now, but if the user wants cost data in the report (e.g., for `--output-format json` in the CLI itself), it would need to be wired into `_build_report`. Not a bug, but a gap.

### tests/test_llm.py — TestParseCliJson and TestCostTracker
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive unit tests covering: Claude JSON parsing with and without cache tokens, Gemini JSON parsing with nested stats structure, invalid JSON fallback, missing `result` key fallback, JSON array fallback, cost tracker reset/accumulate/by-model, format with cost, format without cost, format when empty. The `setup_method` in `TestCostTracker` properly resets state between tests. Tests for existing `invoke_model` behavior correctly updated to wrap mock stdout in JSON.

### tests/test_llm.py — Updated existing tests
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All existing tests that mock `subprocess.run` stdout have been updated to return `json.dumps({"result": "...", "usage": {}, "total_cost_usd": 0.0})` instead of plain text. The `test_claude_does_not_strip_thinking` test correctly embeds the thinking markers inside the JSON `result` field, verifying that thinking-stripping no longer applies to Claude (only ollama). This is a subtle but correct behavioral change — with `--output-format json`, Claude CLI handles thinking internally and the `result` field contains the full output.

### Self-Review
**Limitations:** Could not verify whether `gemini -o json` is the correct flag for Gemini CLI JSON output (vs `--output-format json`). Could not verify the exact JSON schema that `claude --output-format json` produces to confirm field names (`result`, `usage`, `total_cost_usd`) are correct, though the PR title references issue #184 which presumably validated this. Did not have access to the full `cmd_ask` or `cmd_repair_smuggled` or `cmd_list_negative` function bodies to confirm they definitely call LLM functions — relied on the observation data showing import/call patterns.

### Feature Requests
- Include the linked GitHub issue body when an issue number is referenced in the branch name or PR title
- Show the full function body for functions that callers reference (e.g., `cmd_ask`) when they're identified as missing integration points

## gemini [CONCERN]

### reasons_lib/llm.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation of the cost tracker is sound. It uses a global dictionary to accumulate stats and correctly updates it by parsing JSON output from supported CLI tools (Claude and Gemini). The fallback mechanism in `_parse_cli_json` ensures that non-JSON output (e.g., from older CLI versions or errors) is still returned as raw text, preventing breakages. The test suite for these changes is comprehensive, covering various JSON structures, edge cases (invalid JSON), and model-specific parsing logic.

### reasons_lib/cli.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

The integration into major CLI commands (`derive`, `review-beliefs`, `propose-update`, `research`, `contradictions`) is correct and follows a consistent pattern of resetting the tracker at the start and printing to `stderr` at the end. However, cost tracking was not added to `cmd_ask` (the `reasons ask` command) and `cmd_repair` (the `reasons repair` command), both of which are common entry points that invoke LLMs. Additionally, `_invoke_api` (used for Vertex/LangChain models) does not currently report its usage to the tracker, meaning costs for API-based models will remain at 0 in the summary.

### tests/test_llm.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new tests are excellent. They mock subprocess outputs with the expected JSON format to verify that `invoke_model` correctly triggers cost recording. The addition of `TestParseCliJson` and `TestCostTracker` provides thorough unit-level verification of the core logic.

### Self-Review
**Limitations:** I could not verify the exact output format of the `gemini` and `claude` CLI tools to ensure the parsing logic matches their current production output, but the tests assume a specific structure that seems reasonable for those tools. I also could not see the implementation of `_invoke_api` to confirm how difficult it would be to integrate cost tracking there.

### Feature Requests
- Include the `cmd_ask` and `cmd_repair` functions in `reasons_lib/cli.py` in the diff or observations to facilitate full integration review.
- Provide the implementation of `_invoke_api` in `reasons_lib/llm.py` to check if token usage can be extracted from LangChain/Vertex responses.

PR #192, opened 2026-06-09
Add repair-premises and --parallel support (#190)

### LINKED ISSUE #190: Add review-premises command to evaluate premises against source material
## Problem

`review-beliefs` filters for derived beliefs only (nodes with justifications) and skips premises entirely. This is architecturally correct — it evaluates whether conclusions follow from antecedents. But premises have no antecedents. There is no automated step that verifies whether premises accurately reflect their source material.

## Evidence

Phase 2 of the propose-beliefs accuracy experiment (handbook-expert, 100 sampled premises) found:

- **ACCEPT precision: 92%** — 8% of premises contain fabricated details
- **Dominant error type: misread source (7/9 errors)** — the proposer adds plausible details not in the source (e.g., "stored in Redis" when the source never mentions Redis, "JIRA review" when the source says Slack channel)
- **Second error type: overgeneralized (2/9)** — infers rationale or ordering the source does not state
- **Zero factually-wrong errors** — the proposer does not contradict the source, it embellishes it

These errors pass every downstream check. `review-beliefs` evaluates "does the conclusion follow from the premises?" — and it does. The premises are just wrong. This is the "coherence is not correctness" gap applied to the front gate.

At 8% error rate across 3,783 premises in handbook-expert, roughly 300 premises contain fabricated details that no existing pipeline step catches.

## Proposed Solution

Add a `review-premises` command (or a `--premises` mode on `review-beliefs`) that:

1. Takes premise IDs (or defaults to all IN premises with no justifications)
2. For each premise, loads the source document referenced in the node's `source` field
3. Asks an LLM: "Does this source document actually say [claim]?"
4. Evaluates: factually correct, well-supported, error type classification
5. Supports `--dry-run` (report only) and `--auto-retract` (retract flagged premises)
6. Writes results in the same format as `review-beliefs` (markdown output, JSON reports)

The evaluation question for premises is different from derived beliefs:
- **Derived:** "Does this conclusion follow from these antecedents?" (logical validity)
- **Premises:** "Does the source material actually say this?" (factual accuracy)

## Pipeline Integration

This would extend the expert-agent-builder pipeline from:

```
propose-beliefs -> human review -> accept-beliefs
```

to:

```
propose-beliefs -> review-premises -> human review -> accept-beliefs
```

Same generate-then-critique architecture as derive-then-review, applied one step earlier.

## References

- Experiment design: `beliefs-pi/entries/2026/06/08/experiment-design-propose-beliefs-accuracy.md`
- Phase 2 results: `beliefs-pi/entries/2026/06/09/propose-beliefs-accuracy-phase2-results.md`
- `review-beliefs` derived-only filter: `reasons_lib/api.py` line 2541-2546

benthomasson:
## Summary
- New `reasons repair-premises` command: takes review results and either rewrites inaccurate premises from source material or retracts unsupported claims
- `--parallel N` added to both `review-premises` and `repair-premises` for concurrent LLM calls via ThreadPoolExecutor
- Repair uses `--review-file` (JSON from review-premises) or re-reviews specific `belief_ids` inline
- Follows same patterns as existing review/repair pipeline: JSON reports, `--dry-run`, cost tracking

## Usage
```bash
# Review then repair pipeline
reasons review-premises --sample 50 --parallel 3
reasons repair-premises --review-file reviews/review-premises-*.json --parallel 3

# Direct repair by IDs
reasons repair-premises obs-1 obs-2 --dry-run

# Parallel review
reasons review-premises --parallel 4
```

## Test plan
- [x] 21 tests for repair-premises (module, API, CLI, parallel)
- [x] 21 tests for review-premises (existing + parallel batch test)
- [x] Full suite: 1382 passed, 0 failed

Closes #190

### DIFF
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -2626,6 +2626,7 @@ def review_premises(
     visible_to: list[str] | None = None,
     dry_run: bool = False,
     on_batch: Callable | None = None,
+    parallel: int = 0,
     db_path: str = DEFAULT_DB,
 ) -> dict:
     """Review premises against their source material for factual accuracy.
@@ -2702,7 +2703,8 @@ def review_premises(
         review_ids.append(nid)
 
     results = _review(nodes, premise_ids=review_ids, source_contents=source_contents,
-                      model=model, timeout=timeout, on_batch=on_batch)
+                      model=model, timeout=timeout, on_batch=on_batch,
+                      parallel=parallel)
 
     now = datetime.now().isoformat(timespec="seconds")
 
@@ -2732,6 +2734,114 @@ def review_premises(
         "total_premises": total_premises,
         "skipped_no_source": skipped,
         "timestamp": now,
+    }
+
+
+def repair_premises(
+    review_file: str | None = None,
+    belief_ids: list[str] | None = None,
+    model: str = "claude",
+    timeout: int = 300,
+    dry_run: bool = False,
+    parallel: int = 0,
+    on_result: Callable | None = None,
+    db_path: str = DEFAULT_DB,
+) -> dict:
+    """Repair inaccurate premises by rewriting from source or retracting.
+
+    Takes review results (from review_file or by re-reviewing belief_ids)
+    and either rewrites premises to match source material or retracts them.
+
+    Returns: {"results": [...], "total_inaccurate": int, "rewritten": int,
+              "retracted": int, "failed": int}
+    """
+    import sys
+    from pathlib import Path
+
+    from .check_stale import resolve_source_path
+    from .repair_premises import repair_premises as _repair
+
+    if review_file:
+        import json as json_mod
+        data = json_mod.loads(Path(review_file).read_text())
+        review_results_list = data.get("results", [])
+    elif belief_ids:
+        review_result = review_premises(
+            belief_ids=belief_ids, model=model, timeout=timeout,
+            dry_run=True, db_path=db_path,
+        )
+        review_results_list = review_result.get("results", [])
+    else:
+        raise ValueError("Either review_file or belief_ids must be provided")
+
+    inaccurate = [r for r in review_results_list if not r.get("accurate", True)]
+    if not inaccurate:
+        return {"results": [], "total_inaccurate": 0, "rewritten": 0,
+                "retracted": 0, "failed": 0}
+
+    review_map = {r["id"]: r for r in inaccurate}
+
+    result = export_network(db_path=db_path)
+    nodes = result.get("nodes", {})
+    repos = result.get("repos", {})
+    repos = {k: Path(v) if isinstance(v, str) else v for k, v in repos.items()}
+    db_dir = Path(db_path).parent if db_path else None
+
+    source_contents = {}
+    repair_ids = []
+
+    for nid in sorted(review_map.keys()):
+        if nid not in nodes:
+            continue
+        source = nodes[nid].get("source", "")
+        if not source:
+            continue
+        if source not in source_contents:
+            agent = nodes[nid].get("metadata", {}).get("agent")
+            path = resolve_source_path(source, repos=repos, db_dir=db_dir, agent=agent)
+            if path and path.exists():
+                try:
+                    source_contents[source] = path.read_text()
+                except (OSError, UnicodeDecodeError):
+                    continue
+            else:
+                print(f"  SKIP {nid}: source not found: {source}", file=sys.stderr)
+                continue
+        repair_ids.append(nid)
+
+    results = _repair(nodes, premise_ids=repair_ids, source_contents=source_contents,
+                      review_results=review_map, model=model, timeout=timeout,
+                      parallel=parallel, on_result=on_result)
+
+    if not dry_run:
+        for r in results:
+            nid = r["id"]
+            action = r.get("action")
+            if action == "rewrite" and r.get("corrected_text"):
+                try:
+                    update_node(nid, text=r["corrected_text"], db_path=db_path)
+                except Exception as e:
+                    print(f"  ERROR updating {nid}: {e}", file=sys.stderr)
+                    r["action"] = "error"
+            elif action == "retract":
+                try:
+                    retract_node(nid,
+                                 reason=f"repair-premises: {r.get('rationale', 'inaccurate')}",
+                                 db_path=db_path)
+                except Exception as e:
+                    print(f"  ERROR retracting {nid}: {e}", file=sys.stderr)
+                    r["action"] = "error"
+
+    rewritten = sum(1 for r in results if r.get("action") == "rewrite")
+    retracted = sum(1 for r in results if r.get("action") == "retract")
+    failed = sum(1 for r in results if r.get("action") == "error")
+
+    return {
+        "results": results,
+        "total_inaccurate": len(inaccurate),
+        "rewritten": rewritten,
+        "retracted": retracted,
+        "failed": failed,
     }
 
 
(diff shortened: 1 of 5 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#192
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **tests/test_repair_premises.py** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### reasons_lib/repair_premises.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean new module following the same patterns as `review_premises.py`. The prompt template uses `{{` for JSON braces (correct for `.format()`), `parse_repair_response` uses the same JSON-in-prose extraction pattern as `parse_premise_review_response`, and `_repair_one` wraps exceptions into error dicts so `future.result()` never raises in parallel mode. One-per-LLM-call granularity (vs batching in review) is a reasonable tradeoff — it simplifies error isolation and parallelism. The `on_result` callback receives the accumulating `all_results` list, which in parallel mode will be in completion order not input order; this is fine for partial reports since results carry their `id`.

### reasons_lib/review_premises.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good refactor — `_process_batch` extracts the batch processing logic cleanly so both sequential and parallel paths share it. The `as_completed` loop runs in the main thread, so `all_results.extend()` and `on_batch()` are single-threaded. Exception handling in the parallel path mirrors the sequential path (`WARN` + continue). The batch numbering in parallel mode reflects submission order, not completion order, which is correct for the log message. Note: result ordering is non-deterministic in parallel mode (completion order), but since results carry `id` fields this is acceptable.

### reasons_lib/api.py:repair_premises
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Two issues. (1) When `belief_ids` is provided without `review_file`, the function calls `review_premises(..., dry_run=True)` to get review results, then proceeds to repair. But it passes no `parallel` argument to the inner `review_premises` call, so the review step is always sequential regardless of the user's `--parallel` flag. Minor perf gap, not a correctness bug. (2) The `except Exception` blocks around `update_node` and `retract_node` (lines ~2822-2833) silently downgrade the action to `"error"` and print to stderr, but the downstream summary counts an `"error"` in the `failed` bucket — good. However, the error handling doesn't surface *which* exception class hit; for debugging, the `str(e)` in the rationale is sufficient. The source-resolution logic is duplicated from `review_premises()`, but extracting it would be a refactor beyond scope.

### reasons_lib/api.py:review_premises (parallel param addition)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new `parallel` parameter is threaded through to `_review()` correctly. Existing callers default to `parallel=0` (sequential), preserving backward compatibility.

### reasons_lib/cli.py:cmd_repair_premises
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Follows the exact pattern of `cmd_review_premises`. Argparse registration is complete (IDs, `--review-file`, `--model`, `--timeout`, `--dry-run`, `--parallel`, `--report-dir`, `--no-report`). Dispatch table wired. The local `_write_report` correctly scopes to this function. One minor note: the `--parallel` flag was also correctly added to the `review-premises` subparser. Test coverage is limited to checking the function exists (`TestCliDispatch`), not exercising the CLI handler end-to-end, but this matches the existing test pattern for `cmd_review_premises`.

### tests/test_repair_premises.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good breadth — unit tests for formatting, parsing, the repair loop (sequential + parallel + errors + callbacks), API-level integration (review file, rewrite, retract, dry run, validation), CLI registration, and review-premises parallel support. However, two tests use conditional assertions that can silently pass: `test_rewrite_updates_node` at line 244 (`if "obs-1" in net["nodes"]`) and `test_retract_removes_node` at line 261 (`if "obs-2" in net["nodes"]`). Since `update_node` and `retract_node` don't remove nodes from the DB, these nodes will always be present — the `if` guards are dead code that could mask a real regression if the node *were* somehow missing. These should be unconditional `assert` statements.

### Self-Review
**Limitations:** Could not see the full `invoke_model` function body to confirm exactly how `_cost_tracker` is mutated — inferred from the `reset_cost_tracker` structure. Could not see `_with_network` context manager to verify that `update_node` and `retract_node` are safe to call sequentially (each opens its own DB connection). The `build/lib/` copies suggest a stale build directory that may confuse imports if not cleaned.

### Feature Requests
- Include the full body of functions called by the diff (e.g., `invoke_model`) when thread safety is relevant, not just the direct callers
- Flag conditional assertions in test code (`if x: assert y`) as a pattern worth highlighting automatically

## gemini [PASS]

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new `repair_premises` function correctly coordinates the repair workflow: it handles both direct belief IDs and review report files, resolves source material paths, and applies the requested modifications (text updates or retractions) to the database. The addition of the `parallel` parameter to `review_premises` is properly implemented and passes through to the underlying processing logic. While there are redundant local imports (e.g., `sys`, `Path`, `json`) that are already present at the module level, this appears to be a consistent pattern in this codebase.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The CLI implementation properly wires the new repair command, including argument parsing for parallelism, model selection, and reporting. It correctly handles cost tracking and provides clear console output for the actions taken. The update to `cmd_review_premises` to support parallelism is also correct.

### reasons_lib/repair_premises.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

This new module encapsulates the LLM interaction logic well. The prompt is specific and requires structured JSON output. The `parse_repair_response` function is robust, using `raw_decode` to extract the JSON array even if the LLM includes conversational prose. The implementation of `ThreadPoolExecutor` for parallel repairs is sound, though users should be aware that cost tracking in `llm.py` may have minor race conditions when updated from multiple threads (this is acceptable for non-critical telemetry).

### reasons_lib/review_premises.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Refactoring the batch processing into `_process_batch` allows for a clean implementation of concurrent batch reviewing. The use of `as_completed` ensures that results are collected as soon as they are ready, and the logic correctly handles potential exceptions during LLM calls by catching them and printing a warning to stderr.

### tests/test_repair_premises.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The test suite is comprehensive, covering unit-level formatting and parsing, the core repair loop (both sequential and parallel), and the high-level API functions. The use of mocks for LLM calls and file system operations is appropriate and thorough.

### Self-Review
**Limitations:** I was able to verify the full bodies of the primary modified functions and their callers. I confirmed that the signature change for `review_premises` in `api.py` does not break existing callers as they primarily use keyword arguments for the shifted `db_path` parameter.

PR #211, opened 2026-06-15
Add lifecycle timestamps to Node dataclass

### LINKED ISSUE #210: Add created_at and updated_at timestamps to Node dataclass
## Summary

Beliefs have no timestamps. The `date` field on the Node dataclass exists but is only populated during import -- `add_node` never sets it automatically. There is no way to know when a belief was created or last modified.

## Proposed change

Add `created_at`, `updated_at`, `reviewed_at`, `verified_at`, and `retracted_at` as first-class fields on the Node dataclass (not metadata).

### Behavior

- **`add_node`**: Set `created_at` and `updated_at` to current ISO timestamp.
- **`update_node`**, **`set_metadata`**: Update `updated_at`.
- **`retract_node`**: Set `retracted_at` and update `updated_at`.
- **`review-beliefs`**: Set `reviewed_at` on each reviewed belief.
- **`verify`** (code-expert): Set `verified_at` on each verified belief.
- **`show_node`**: Include all five timestamps in output.
- **`export` / `import`**: Round-trip all fields.

### Schema changes

- **SQLite (`storage.py`)**: Add `created_at`, `updated_at`, `reviewed_at`, `verified_at`, `retracted_at` TEXT columns. Migration for existing databases.
- **PostgreSQL (`pg.py`)**: Add corresponding `TIMESTAMPTZ` columns to `rms_nodes`.

## Why not metadata?

Timestamps are universal (every node has them), queryable (sort by creation date, find stale beliefs), and should survive export/import reliably. Metadata is for optional, domain-specific key-value pairs. Timestamps belong alongside `source`, `date`, and other first-class fields.

Note: `review-beliefs` currently stores `last_reviewed` in metadata (`api.py:2730`). This should migrate to the `reviewed_at` field.

## Affected commands

- `add` / `add_node` -- set `created_at` and `updated_at`
- `update` / `update_node` -- update `updated_at`
- `set-metadata` / `set_metadata` -- update `updated_at`
- `retract` / `retract_node` -- set `retracted_at`, update `updated_at`
- `review-beliefs` -- set `reviewed_at` (replace metadata `last_reviewed`)
- `verify` (code-expert) -- set `verified_at`
- `show` / `show_node` -- display timestamps
- `export` / `import` -- round-trip
- `list` -- optionally sortable by timestamp

benthomasson:
## Summary
- Adds created_at, updated_at, reviewed_at, verified_at, retracted_at as first-class fields on the Node dataclass (closes #210)
- Auto-sets created_at/updated_at on add_node, updated_at on mutations (update_node, set_metadata), retracted_at/updated_at on retract, reviewed_at/updated_at on review_beliefs
- Migrates review_beliefs from metadata[last_reviewed] to node.reviewed_at
- Updates SQLite schema + migration, PG schema + migration, storage save/load, API (show/export/import), CLI show output
- 7 new lifecycle timestamp tests, all 1497 tests passing

## Test plan
- [x] TestLifecycleTimestamps::test_add_node_sets_created_at
- [x] TestLifecycleTimestamps::test_update_node_sets_updated_at
- [x] TestLifecycleTimestamps::test_set_metadata_sets_updated_at
- [x] TestLifecycleTimestamps::test_retract_sets_retracted_at
- [x] TestLifecycleTimestamps::test_show_node_includes_all_timestamps
- [x] TestLifecycleTimestamps::test_export_includes_timestamps
- [x] TestLifecycleTimestamps::test_import_roundtrips_timestamps
- [x] Full test suite: 1497 passed, 166 skipped

### DIFF
--- a/reasons_lib/__init__.py
+++ b/reasons_lib/__init__.py
@@ -39,6 +39,11 @@ class Node:
     source_hash: str = ""
     date: str = ""
     metadata: dict = field(default_factory=dict)
+    created_at: str = ""
+    updated_at: str = ""
+    reviewed_at: str = ""
+    verified_at: str = ""
+    retracted_at: str = ""
 
 
 @dataclass
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -13,6 +13,7 @@ import re
 import sqlite3
 import sys
 from collections.abc import Callable
+from datetime import datetime, timezone
 from itertools import combinations
 from pathlib import Path
 
@@ -627,6 +628,11 @@ def show_node(node_id: str, visible_to: list[str] | None = None, db_path: str = 
             ],
             "dependents": sorted(node.dependents),
             "metadata": node.metadata,
+            "created_at": node.created_at,
+            "updated_at": node.updated_at,
+            "reviewed_at": node.reviewed_at,
+            "verified_at": node.verified_at,
+            "retracted_at": node.retracted_at,
         }
 
 
@@ -808,6 +814,8 @@ def update_node(
             meta["example"] = example
             node.metadata = meta
             updated.append("example")
+        if updated:
+            node.updated_at = datetime.now(timezone.utc).isoformat(timespec="seconds")
         return {"node_id": node_id, "updated_fields": updated}
 
 
@@ -829,6 +837,7 @@ def set_metadata(
         meta = node.metadata or {}
         meta[key] = value
         node.metadata = meta
+        node.updated_at = datetime.now(timezone.utc).isoformat(timespec="seconds")
         return {"node_id": node_id, "key": key}
 
 
@@ -950,6 +959,11 @@ def export_network(visible_to: list[str] | None = None, db_path: str = DEFAULT_D
                 "source_hash": n.source_hash,
                 "date": n.date,
                 "metadata": {k: v for k, v in n.metadata.items() if not k.startswith("_")},
+                "created_at": n.created_at,
+                "updated_at": n.updated_at,
+                "reviewed_at": n.reviewed_at,
+                "verified_at": n.verified_at,
+                "retracted_at": n.retracted_at,
             }
             for nid, n in sorted(net.nodes.items())
             if visible_to is None or _is_visible(n, visible_to)
@@ -1258,7 +1272,12 @@ def import_json(json_file: str, db_path: str = DEFAULT_DB,
                         source_hash=ndata.get("source_hash", ""),
                         date=ndata.get("date", ""),
                         metadata=ndata.get("metadata", {}),
+                        created_at=ndata.get("created_at", ""),
+                        updated_at=ndata.get("updated_at", ""),
                     )
+                    node.reviewed_at = ndata.get("reviewed_at", "")
+                    node.verified_at = ndata.get("verified_at", "")
+                    node.retracted_at = ndata.get("retracted_at", "")
                     # Restore exact truth value (may differ from computed if retracted)
                     target_tv = ndata.get("truth_value", "IN")
                     if node.truth_value != target_tv:
@@ -1287,7 +1306,7 @@ def import_json(json_file: str, db_path: str = DEFAULT_DB,
                             )
                             for j in jlist
                         ]
-                    net.add_node(
+                    node = net.add_node(
                         id=nid,
                         text=ndata.get("text", ""),
                         justifications=justifications,
@@ -1296,9 +1315,14 @@ def import_json(json_file: str, db_path: str = DEFAULT_DB,
                         source_hash=ndata.get("source_hash", ""),
                         date=ndata.get("date", ""),
                         metadata=ndata.get("metadata", {}),
+                        created_at=ndata.get("created_at", ""),
+                        updated_at=ndata.get("updated_at", ""),
                     )
+                    node.reviewed_at = ndata.get("reviewed_at", "")
+                    node.verified_at = ndata.get("verified_at", "")
+                    node.retracted_at = ndata.get("retracted_at", "")
                     target_tv = ndata.get("truth_value", "IN")
-                    if net.nodes[nid].truth_value != target_tv:
+                    if node.truth_value != target_tv:
                         if target_tv == "OUT":
                             net.retract(nid)
                         else:
@@ -2265,12 +2289,12 @@ def list_nodes(
                             status=status, premises_only=premises_only,
                             has_dependents=has_dependents, namespace=namespace,
                             visible_to=visible_to, label=label)
-    from datetime import datetime, timedelta
+    from datetime import timedelta
 
     with _with_network(db_path) as net:
         memo = {} if (min_depth is not None or max_depth is not None) else None
         if not_reviewed_since is not None:
-            cutoff = datetime.now() - timedelta(days=not_reviewed_since)
+            cutoff = datetime.now(timezone.utc) - timedelta(days=not_reviewed_since)
         else:
             cutoff = None
         nodes = []
@@ -2298,12 +2322,12 @@ def list_nodes(
             if never_reviewed:
                 if not node.justifications:
                     continue
-                if node.metadata.get("last_reviewed"):
+                if node.reviewed_at or node.metadata.get("last_reviewed"):
                     continue
             if cutoff is not None:
                 if not node.justifications:
                     continue
-                last = node.metadata.get("last_reviewed")
+                last = node.reviewed_at or node.metadata.get("last_reviewed", "")
                 if last:
                     try:
                         if datetime.fromisoformat(last) >= cutoff:
@@ -2317,7 +2341,7 @@ def list_nodes(
                 "justification_count": len(node.justifications),
                 "dependent_count": len(node.dependents),
                 "challenges": node.metadata.get("challenges", []),
-                "last_reviewed": node.metadata.get("last_reviewed"),
+                "last_reviewed": node.reviewed_at or node.metadata.get("last_reviewed"),
                 "review_result": node.metadata.get("review_result"),
                 "source_type": node.metadata.get("source_type", ""),
             })
@@ -2692,14 +2716,13 @@ def review_beliefs(
     """Review derived beliefs for validity, sufficiency, and necessity.
 
     Uses an LLM to evaluate whether each derived belief's reasoning
-    from antecedents to conclusion is sound. Writes last_reviewed and
-    review_result metadata to each reviewed node (unless dry_run=True).
+    from antecedents to conclusion is sound. Sets reviewed_at and updated_at
+    timestamps and writes review_result metadata to each reviewed node
+    (unless dry_run=True).
 
     Returns: {"results": [...], "reviewed": int, "invalid": int,
               "insufficient": int, "unnecessary": int, "total_derived": int}
     """
-    from datetime import datetime
-
     from .derive import _get_depth
     from .review import review_beliefs as _review
 
@@ -2761,14 +2784,15 @@ def review_beliefs(
     results = _review(nodes, belief_ids=review_ids, model=model, timeout=timeout,
                       on_batch=on_batch)
 
-    now = datetime.now().isoformat(timespec="seconds")
+    now = datetime.now(timezone.utc).isoformat(timespec="seconds")
 
     if not dry_run and results:
         result_map = {r["id"]: r for r in results}
         with _with_network(db_path, write=True) as net:
             for nid in review_ids:
                 if nid in net.nodes and nid in result_map:
-                    net.nodes[nid].metadata["last_reviewed"] = now
+                    net.nodes[nid].reviewed_at = now
+                    net.nodes[nid].updated_at = now
                     net.nodes[nid].metadata["review_result"] = _classify_review_result(result_map[nid])
 
     invalid = sum(1 for r in results if not r.get("valid", True))
@@ -2808,8 +2832,6 @@ def review_premises(
               "skipped_no_source": int}
     """
     import sys
-    from datetime import datetime
-    from pathlib import Path
 
     from .check_stale import resolve_source_path
     from .review_premises import review_premises as _review
@@ -3030,8 +3052,6 @@ def propose_update(
 
     Returns: {"proposals": [...], "reviewed": int, "timestamp": str}
     """
-    from datetime import datetime
-
     from .propose_update import propose_updates as _propose
 
     result = export_network(db_path=db_path)
--- a/reasons_lib/api.py
+++ b/reasons_lib/api.py
@@ -1275,9 +1275,6 @@ def import_json(json_file: str, db_path: str = DEFAULT_DB,
                         created_at=ndata.get("created_at", ""),
                         updated_at=ndata.get("updated_at", ""),
                     )
-                    node.reviewed_at = ndata.get("reviewed_at", "")
-                    node.verified_at = ndata.get("verified_at", "")
-                    node.retracted_at = ndata.get("retracted_at", "")
                     # Restore exact truth value (may differ from computed if retracted)
                     target_tv = ndata.get("truth_value", "IN")
                     if node.truth_value != target_tv:
@@ -1285,6 +1282,11 @@ def import_json(json_file: str, db_path: str = DEFAULT_DB,
                             net.retract(nid)
                         else:
                             net.assert_node(nid)
+                    # Restore exact timestamps AFTER retract/assert to avoid overwrite
+                    node.reviewed_at = ndata.get("reviewed_at", "")
+                    node.verified_at = ndata.get("verified_at", "")
+                    node.retracted_at = ndata.get("retracted_at", "")
+                    node.updated_at = ndata.get("updated_at", "")
                     added.add(nid)
                     nodes_imported += 1
                 else:
@@ -1318,15 +1320,16 @@ def import_json(json_file: str, db_path: str = DEFAULT_DB,
                         created_at=ndata.get("created_at", ""),
                         updated_at=ndata.get("updated_at", ""),
                     )
-                    node.reviewed_at = ndata.get("reviewed_at", "")
-                    node.verified_at = ndata.get("verified_at", "")
-                    node.retracted_at = ndata.get("retracted_at", "")
                     target_tv = ndata.get("truth_value", "IN")
                     if node.truth_value != target_tv:
                         if target_tv == "OUT":
                             net.retract(nid)
                         else:
                             net.assert_node(nid)
+                    node.reviewed_at = ndata.get("reviewed_at", "")
+                    node.verified_at = ndata.get("verified_at", "")
+                    node.retracted_at = ndata.get("retracted_at", "")
+                    node.updated_at = ndata.get("updated_at", "")
                     added.add(nid)
                     nodes_imported += 1
                 break
@@ -2330,7 +2333,10 @@ def list_nodes(
                 last = node.reviewed_at or node.metadata.get("last_reviewed", "")
                 if last:
                     try:
-                        if datetime.fromisoformat(last) >= cutoff:
+                        dt = datetime.fromisoformat(last)
+                        if dt.tzinfo is None:
+                            dt = dt.replace(tzinfo=timezone.utc)
+                        if dt >= cutoff:
                             continue
                     except ValueError:
                         pass
--- a/reasons_lib/cli.py
+++ b/reasons_lib/cli.py
@@ -290,6 +290,17 @@ def cmd_show(args):
         indented = node["metadata"]["example"].replace("\n", "\n  ")
         print(f"\nExample:\n  {indented}")
 
+    timestamps = []
+    for ts_key in ("created_at", "updated_at", "reviewed_at", "verified_at", "retracted_at"):
+        ts_val = node.get(ts_key, "")
+        if ts_val:
+            label = ts_key.replace("_", " ").title().replace(" At", "")
+            timestamps.append((label, ts_val))
+    if timestamps:
+        print()
+        for label, val in timestamps:
+            print(f"{label}: {val}")
+
     if node["dependents"]:
         print(f"\nDependents: {', '.join(node['dependents'])}")
 
--- a/reasons_lib/network.py
+++ b/reasons_lib/network.py
@@ -9,7 +9,7 @@ Implements Doyle's (1979) TMS algorithm:
 """
 
 from collections import deque
-from datetime import datetime
+from datetime import datetime, timezone
 
 from . import Node, Justification, Nogood
 
@@ -79,6 +79,11 @@ class Network:
         source_hash: str = "",
         date: str = "",
         metadata: dict | None = None,
+        created_at: str = "",
+        updated_at: str = "",
+        reviewed_at: str = "",
+        verified_at: str = "",
+        retracted_at: str = "",
     ) -> Node:
         """Add a node to the network and propagate.
 
@@ -87,6 +92,11 @@ class Network:
         """
         if id in self.nodes:
             raise ValueError(f"Node '{id}' already exists")
+
+        if not created_at:
+            now = datetime.now(timezone.utc).isoformat(timespec="seconds")
+            created_at = now
+            updated_at = now
 
         node = Node(
             id=id,
@@ -97,6 +107,11 @@ class Network:
             source_hash=source_hash,
             date=date,
             metadata=metadata or {},
+            created_at=created_at,
+            updated_at=updated_at,
+            reviewed_at=reviewed_at,
+            verified_at=verified_at,
+            retracted_at=retracted_at,
         )
 
         # Register as dependent of antecedents (inlist) and outlist nodes.
@@ -135,12 +150,17 @@ class Network:
             raise KeyError(f"Node '{node_id}' not found")
 
         node = self.nodes[node_id]
+        now = datetime.now(timezone.utc).isoformat(timespec="seconds")
         if node.truth_value == "OUT":
             node.metadata["_retracted"] = True
+            node.retracted_at = now
+            node.updated_at = now
             return []
 
         node.truth_value = "OUT"
         node.metadata["_retracted"] = True
+        node.retracted_at = now
+        node.updated_at = now
         if reason:
             node.metadata["retract_reason"] = reason
         changed = [node_id]
--- a/reasons_lib/network.py
+++ b/reasons_lib/network.py
@@ -184,6 +184,8 @@ class Network:
 
         node.truth_value = "IN"
         node.metadata.pop("_retracted", None)
+        node.retracted_at = ""
+        node.updated_at = datetime.now(timezone.utc).isoformat(timespec="seconds")
         changed = [node_id]
         self._log("assert", node_id, "IN")
 
--- a/reasons_lib/pg.py
+++ b/reasons_lib/pg.py
@@ -30,6 +30,10 @@ CREATE TABLE IF NOT EXISTS rms_nodes (
     date TEXT DEFAULT '',
     metadata JSONB DEFAULT '{}',
     created_at TIMESTAMPTZ DEFAULT now(),
+    updated_at TIMESTAMPTZ,
+    reviewed_at TIMESTAMPTZ,
+    verified_at TIMESTAMPTZ,
+    retracted_at TIMESTAMPTZ,
     PRIMARY KEY (id, project_id)
 );
 
@@ -132,6 +136,14 @@ class PgApi:
             )
             if not cur.fetchone():
                 cur.execute("ALTER TABLE rms_nodes ADD COLUMN source_url TEXT DEFAULT ''")
+            for col in ("updated_at", "reviewed_at", "verified_at", "retracted_at"):
+                cur.execute(
+                    "SELECT column_name FROM information_schema.columns "
+                    "WHERE table_name = 'rms_nodes' AND column_name = %s",
+                    (col,),
+                )
+                if not cur.fetchone():
+                    cur.execute(f"ALTER TABLE rms_nodes ADD COLUMN {col} TIMESTAMPTZ")
             now = datetime.now(timezone.utc).isoformat(timespec="seconds")
             pname = project_name or self.project_id
             for key, val in [
@@ -153,22 +165,25 @@ class PgApi:
 
     def _add_node_raw(self, cur, node_id, text, justifications=None,
                       source="", source_url="", source_hash="", date="",
-                      metadata=None):
+                      metadata=None, created_at="", updated_at=""):
         """Insert a node with pre-parsed justification dicts.
 
         Does NOT commit, validate refs, or inherit access tags.
         Caller is responsible for transaction management.
         """
         pid = self.project_id
-        now = date or datetime.now().isoformat(timespec="seconds")
+        now_ts = datetime.now(timezone.utc)
+        now = date or now_ts.isoformat(timespec="seconds")
         meta = metadata or {}
+        ts_created = created_at or now_ts.isoformat(timespec="seconds")
+        ts_updated = updated_at or ts_created
 
         cur.execute(
             "INSERT INTO rms_nodes (id, project_id, text, source, source_url, "
-            "source_hash, date, metadata) "
-            "VALUES (%s, %s, %s, %s, %s, %s, %s, %s)",
+            "source_hash, date, metadata, created_at, updated_at) "
+            "VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s)",
             (node_id, pid, text, source, source_url, source_hash, now,
-             json.dumps(meta)),
+             json.dumps(meta), ts_created, ts_updated),
         )
 
         if justifications:
@@ -358,18 +373,22 @@ class PgApi:
             if reason:
                 metadata["retract_reason"] = reason
 
+            now = datetime.now(timezone.utc).isoformat(timespec="seconds")
+
             if old_value == "OUT":
                 cur.execute(
-                    "UPDATE rms_nodes SET metadata = %s WHERE id = %s AND project_id = %s",
-                    (json.dumps(metadata), node_id, pid),
+                    "UPDATE rms_nodes SET metadata = %s, retracted_at = %s, updated_at = %s "
+                    "WHERE id = %s AND project_id = %s",
+                    (json.dumps(metadata), now, now, node_id, pid),
                 )
                 self.conn.commit()
                 return {"changed": [], "went_out": [], "went_in": []}
 
             cur.execute(
-                "UPDATE rms_nodes SET truth_value = 'OUT', metadata = %s "
+                "UPDATE rms_nodes SET truth_value = 'OUT', metadata = %s, "
+                "retracted_at = %s, updated_at = %s "
                 "WHERE id = %s AND project_id = %s",
-                (json.dumps(metadata), node_id, pid),
+                (json.dumps(metadata), now, now, node_id, pid),
             )
             self._log(cur, "retract", node_id, reason or "OUT")
 
@@ -657,7 +676,8 @@ class PgApi:
         pid = self.project_id
         with self.conn.cursor() as cur:
             cur.execute(
-                "SELECT id, text, truth_value, source, source_url, source_hash, metadata "
+                "SELECT id, text, truth_value, source, source_url, source_hash, metadata, "
+                "created_at, updated_at, reviewed_at, verified_at, retracted_at "
                 "FROM rms_nodes WHERE id = %s AND project_id = %s",
                 (node_id, pid),
             )
@@ -665,7 +685,8 @@ class PgApi:
             if not row:
                 raise KeyError(f"Node '{node_id}' not found")
 
-            nid, text, tv, source, source_url, source_hash, meta = row
+            nid, text, tv, source, source_url, source_hash, meta, \
+                created_at, updated_at, reviewed_at, verified_at, retracted_at = row
             if isinstance(meta, str):
                 meta = json.loads(meta)
 
@@ -689,6 +710,9 @@ class PgApi:
 
             dependents = sorted(self._find_dependents(cur, [node_id]))
 
+        def _fmt_ts(val):
+            return val.isoformat(timespec="seconds") if val else ""
+
         return {
             "id": nid,
             "text": text,
@@ -699,6 +723,11 @@ class PgApi:
             "justifications": justifications,
             "dependents": dependents,
             "metadata": meta,
+            "created_at": _fmt_ts(created_at),
+            "updated_at": _fmt_ts(updated_at),
+            "reviewed_at": _fmt_ts(reviewed_at),
+            "verified_at": _fmt_ts(verified_at),
+            "retracted_at": _fmt_ts(retracted_at),
         }
 
     def search(self, query, visible_to=None, format="markdown"):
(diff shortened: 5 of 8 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#211
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## Disagreements

- **reasons_lib/api.py:list_nodes** [LOW]: claude: PASS, gemini: CONCERN

## claude [BLOCK]

### reasons_lib/__init__.py:Node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Five new string fields with empty defaults. Dataclass field ordering is correct (all have defaults, appended after existing defaults). No issues.

### reasons_lib/network.py:add_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Auto-generates `created_at` and `updated_at` when `created_at` is empty. If `created_at` is supplied (e.g. import), both are passed through. The `if not created_at` guard is clean.

### reasons_lib/network.py:retract
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

`retract()` correctly sets `retracted_at` and `updated_at`. However, two asymmetries: (1) `assert_node()` (lines 173-192) does NOT set `updated_at` or clear `retracted_at` when restoring a node to IN. A node retracted at T1 and reasserted at T2 will still show `retracted_at=T1` and `updated_at=T1` — misleading. (2) `_propagate()` (lines 881-912) changes dependent truth values without setting any timestamps on them, so cascade-retracted nodes won't have `retracted_at` set. The second point may be intentional (retracted_at = explicit retraction only), but `updated_at` should arguably be set on any truth value change.

**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

After restoring timestamps from the export data, the code calls `net.retract(nid)` to restore truth_value=OUT. But `retract()` overwrites `retracted_at` and `updated_at` with `datetime.now()`. This means the original timestamps from the export are lost. The roundtrip test at line 838 only asserts `node["retracted_at"] != ""` — it doesn't verify the value matches the export, so the bug passes silently. The same issue exists in both code paths (topological-sort path at line ~1280 and fallback path at line ~1323). Fix: either set timestamps *after* the retract call, or use a lower-level mechanism to set truth_value without triggering timestamp updates.

### reasons_lib/api.py:update_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Only sets `updated_at` when `updated` list is non-empty, correctly guarding against no-op updates. UTC timestamps are consistent.

### reasons_lib/api.py:set_metadata
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Unconditionally sets `updated_at` on metadata change, which is correct since metadata always changes when this is called.

### reasons_lib/api.py:show_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All five timestamp fields added to the return dict.

### reasons_lib/api.py:export_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All five timestamps included in export. Matches the import expectations.

### reasons_lib/api.py:review_beliefs
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Migrated from `metadata["last_reviewed"]` to `node.reviewed_at`. Also sets `updated_at`. Removed redundant `from datetime import datetime` since it's already at module level. Docstring updated.

### reasons_lib/api.py:list_nodes
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Good backward compatibility — `node.reviewed_at or node.metadata.get("last_reviewed")` gracefully handles pre-migration data. The `cutoff` comparison at line 2322 now uses `timezone.utc`, which is correct. The output dict still emits `"last_reviewed"` key for API stability, populated from the new field with fallback.

### reasons_lib/storage.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Schema migration via `ALTER TABLE ADD COLUMN` with column-existence check is correct. Save/load correctly round-trips all five fields. The `has_timestamps` branch in `load()` is clean and backward-compatible with older DBs missing the columns.

### reasons_lib/pg.py:_add_node_raw
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Accepts `created_at` and `updated_at` with sensible defaults derived from `now_ts`. The INSERT includes both new columns. However, callers like `add_node`, `import_beliefs`, `import_agent`, `sync_agent` don't pass these params — they'll get auto-generated defaults, which is reasonable for new nodes.

### reasons_lib/pg.py:retract_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Both the already-OUT and transitioning-to-OUT paths set `retracted_at` and `updated_at`. Mirrors the network.py behavior. Same `assert_node` asymmetry concern applies (pg.py `assert_node` at lines 401-432 doesn't set any timestamps).

### reasons_lib/pg.py:show_node
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `_fmt_ts` helper is defined as a nested function after the `with` block closes. It handles `datetime` objects from PostgreSQL correctly (`.isoformat()`). However, it's defined at function scope but only used for the return dict, which is fine. Minor: if the DB has NULL timestamps (columns are nullable TIMESTAMPTZ with no default), `_fmt_ts(None)` correctly returns `""`.

### reasons_lib/pg.py:export_network
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

`_fmt_ts` is redefined inside the `for row in cur.fetchall()` loop body. It should be defined once outside the loop. Not a correctness bug but creates a new function object per row for no reason.

### reasons_lib/pg.py:import_json
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The timestamp UPDATE happens before the truth_value restoration loop. If the truth_value restoration code calls `retract_node()` (which sets `retracted_at`), the imported value would be overwritten — same class of bug as api.py. Need to verify whether pg.py's truth_value restoration uses raw SQL or `retract_node()`. The `_add_node_raw` calls correctly pass `created_at` and `updated_at`.

### reasons_lib/pg.py:init_db migration
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Column-existence check via `information_schema.columns` before ALTER TABLE is the correct pattern for PostgreSQL. All four columns are added (created_at already existed in the schema).

### reasons_lib/cli.py:cmd_show
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Clean display of non-empty timestamps. The label transform `"created_at" -> "Created"` via `replace("_", " ").title().replace(" At", "")` is correct for all five fields. No CLI-level tests for this display, but it's pure formatting.

### tests/test_api.py:TestLifecycleTimestamps
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Tests cover add, update, set_metadata, retract, show, export, and import roundtrip. However, `test_import_roundtrips_timestamps` at line 847 only asserts `node["retracted_at"] != ""` instead of `node["retracted_at"] == export["nodes"]["ts-g"]["retracted_at"]`. This masks the retract-overwrite bug. Missing tests: (1) assert_node behavior with timestamps, (2) cascade propagation timestamp behavior, (3) reviewed_at and verified_at being set/used.

### tests/test_review.py changes
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Tests correctly updated from `metadata["last_reviewed"]` to `node["reviewed_at"]`. The dry_run test now checks `not node["reviewed_at"]` instead of key absence, which is correct since the field always exists but is empty when unreviewed.

### Self-Review
**Limitations:** Could not see full pg.py import_json truth-value restoration code to confirm whether it uses `retract_node()` (which would trigger timestamp overwrite) or raw SQL (which wouldn't). Could not verify whether the test at test_review.py:739 (`assert "last_reviewed" in node`) still passes — it should via the backward-compat output in list_nodes, but didn't have full test context. Could not run the test suite to confirm all 211+ tests pass.

### Feature Requests
- Include the full function body for functions that span multiple diff hunks (especially import_json in pg.py which had critical context between hunks)
- Show the test assertion that a roundtrip test actually validates (not just that it runs) — flag weak assertions like `!= ""` when testing round-trip fidelity

## gemini [CONCERN]

### reasons_lib/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adding new timestamp fields (`created_at`, `updated_at`, `reviewed_at`, `verified_at`, `retracted_at`) to the `Node` dataclass with default empty strings (`""`) at the end of the class definition is correct and fully backward-compatible. This avoids default-argument ordering issues in Python dataclasses.

### reasons_lib/api.py:list_nodes
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

There is a timezone-awareness mismatch bug. `cutoff` is defined as an aware datetime object: `datetime.now(timezone.utc) - timedelta(...)`. However, legacy/older records in the database store `last_reviewed` as naive timestamp strings (since the previous implementation used `datetime.now().isoformat()`). 

When parsing these naive strings using `datetime.fromisoformat(last)`, Python returns a naive datetime object. Comparing naive and aware datetimes (`datetime.fromisoformat(last) >= cutoff`) raises a `TypeError: can't compare offset-naive and offset-aware datetimes`. Because the `try-except` block only catches `ValueError`, any legacy node with a naive timestamp will crash `list_nodes()`.

**Suggested Fix:**
Ensure the parsed datetime is converted to UTC or treated consistently:
```python
dt = datetime.fromisoformat(last)
if dt.tzinfo is None:
    dt = dt.replace(tzinfo=timezone.utc)
if dt >= cutoff:
    continue
```

### reasons_lib/api.py:import_json (and reasons_lib/pg.py:import_json)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The JSON import process does not fully preserve original timestamps for retracted nodes. 
In both the SQLite and PostgreSQL implementations, `import_json` first sets the imported timestamps (`created_at`, `updated_at`, `retracted_at`, etc.). However, it then restores the node's `truth_value` by calling `net.retract(nid)` or `self.retract_node(nid)`.

Calling `retract`/`retract_node` overwrites `retracted_at` and `updated_at` with the *current* system time (`now`), losing the historical timestamps from the imported data and violating the "lossless round-trip" guarantee of `import_json`. This behavior was masked in the test suite (`test_import_roundtrips_timestamps`) because the assertions only verified that `node["retracted_at"] != ""` instead of asserting exact match with the exported timestamp.

**Suggested Fix:**
To prevent logic methods from overwriting imported metadata, restore the truth value first and then apply the exact timestamps afterward:
```python
# SQLite api.py suggestion:
node = net.add_node(..., created_at=ndata.get("created_at", ""))
target_tv = ndata.get("truth_value", "IN")
if node.truth_value != target_tv:
    if target_tv == "OUT":
        net.retract(nid)
# Apply/overwrite exact historical timestamps AFTER retraction logic runs
node.reviewed_at = ndata.get("reviewed_at", "")
node.verified_at = ndata.get("verified_at", "")
node.retracted_at = ndata.get("retracted_at", "")
node.updated_at = ndata.get("updated_at", "")
```

### reasons_lib/network.py:assert_node (and reasons_lib/pg.py:assert_node)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

When a retracted node is restored/asserted back to `IN` via `assert_node`, the `updated_at` timestamp is not modified, and the outdated `retracted_at` timestamp is not cleared or invalidated. This leaves the database/network in an inconsistent state where a node's truth value is `"IN"`, but it still holds a stale retraction timestamp.

**Suggested Fix:**
In both `Network.assert_node` and `PgApi.assert_node`, update `updated_at` to the current timestamp and clear the `retracted_at` field (or set it to `""`):
```python
node.truth_value = "IN"
node.metadata.pop("_retracted", None)
node.retracted_at = ""
node.updated_at = datetime.now(timezone.utc).isoformat(timespec="seconds")
```

### reasons_lib/storage.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The schema migration logic correctly checks for the presence of the columns and applies the appropriate `ALTER TABLE` operations. The mapping of the new fields in both the serialization (`save`) and deserialization (`load`) logic is solid, ensuring proper direct SQLite persistence of the new timestamps.

### tests/test_api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new lifecycle tests verify standard timestamp operations correctly. However, as noted in the `import_json` review, the roundtrip assertion should be made more robust by asserting equality of `retracted_at` instead of merely asserting that it is non-empty.
```python
# Weak assertion in test:
assert node["retracted_at"] != ""

# Stronger/correct assertion:
assert node["retracted_at"] == export["nodes"]["ts-g"]["retracted_at"]
```

### Self-Review
**Limitations:** Reviewed without dynamic execution capabilities, but conducted a rigorous static analysis tracing of temporal comparisons, state consistency (retraction vs restoration), and serialization/deserialization flows.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#211
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **tests/test_review.py** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### reasons_lib/__init__.py:Node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean addition of 5 string fields with empty-string defaults. Backward-compatible — existing code that constructs `Node` without these fields gets safe defaults. All fields are consumed by storage, API, CLI, and PG layers.

### reasons_lib/network.py:add_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Auto-generates `created_at` and `updated_at` only when `created_at` is empty, preserving caller-supplied values during import. The `if not created_at` guard correctly treats both fresh creation and import-with-timestamps. Returns the created node (already did before), enabling callers to read back timestamps.

### reasons_lib/network.py:retract
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Sets `retracted_at` and `updated_at` on the directly retracted node — correct. However, `_propagate` (called on line 171) recomputes truth values for dependent nodes without setting `updated_at` on those cascade-affected nodes. A dependent that transitions from IN→OUT via cascade will have its `truth_value` changed but `updated_at` untouched. This means `updated_at` doesn't reflect the last time a node's state actually changed, only explicit mutations. Whether this is a bug depends on intent, but it's a semantic gap worth acknowledging.

### reasons_lib/network.py:assert_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Clears `retracted_at` (sets to "") and updates `updated_at`. Same cascade concern as `retract` — dependents restored by propagation won't get `updated_at` set. No dedicated test for assert clearing `retracted_at`, though the roundtrip test exercises a retracted node.

### reasons_lib/api.py:show_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Includes all 5 timestamp fields in output dict. `test_show_node_includes_all_timestamps` verifies presence of all keys.

### reasons_lib/api.py:update_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`updated_at` is set only when `if updated:` — so no-op calls don't produce spurious timestamp changes. Test `test_update_node_sets_updated_at` covers this.

### reasons_lib/api.py:set_metadata
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Sets `updated_at` unconditionally, which is correct since set_metadata always mutates. Test `test_set_metadata_sets_updated_at` covers this.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly passes `created_at` and `updated_at` to `net.add_node()`, then restores `reviewed_at`, `verified_at`, `retracted_at`, and `updated_at` AFTER retract/assert calls to avoid overwrite. The ordering (updated_at restored last) is correct. Both the "ready to add" and "no progress" branches have identical timestamp handling. The `node = net.add_node(...)` return value is now captured (line 1308), fixing a minor pre-existing issue. `test_import_roundtrips_timestamps` exercises the full path including retract.

### reasons_lib/api.py:list_nodes
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Migrates from `datetime.now()` (naive) to `datetime.now(timezone.utc)` (aware) for cutoff calculation. The naive/aware comparison fix at line 2337-2339 (`if dt.tzinfo is None: dt = dt.replace(tzinfo=timezone.utc)`) handles old metadata that stored naive timestamps. The `last_reviewed` output key is preserved as `node.reviewed_at or node.metadata.get("last_reviewed")` for backward compatibility — correct migration pattern. No new tests for the `not_reviewed_since` / `never_reviewed` filter paths with the new field.

### reasons_lib/api.py:review_beliefs
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Now sets `reviewed_at` and `updated_at` on reviewed nodes instead of `metadata["last_reviewed"]`. Uses UTC-aware timestamps. Docstring updated to match. Tests in `test_review.py` updated to check `reviewed_at` instead of `metadata["last_reviewed"]`.

### reasons_lib/api.py:export_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All 5 timestamps included in export output. `test_export_includes_timestamps` verifies `created_at` presence and non-emptiness.

### reasons_lib/cli.py:cmd_show
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Timestamp display logic is clean. The label transform (`ts_key.replace("_", " ").title().replace(" At", "")`) produces "Created", "Updated", "Reviewed", "Verified", "Retracted" — all correct. Only displays non-empty values. No CLI-level tests but this is presentation-only logic.

### reasons_lib/storage.py (schema + save + load)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Schema migration adds columns via ALTER TABLE with existence check. Save includes all 5 fields. Load has a `has_timestamps` backward-compat branch that falls back to empty strings for older DBs. The `or ""` guards on load (e.g., `created_at or ""`) handle NULL values from SQLite. The API-level roundtrip test exercises the storage layer indirectly.

### reasons_lib/pg.py:_add_node_raw
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

Signature now accepts `created_at` and `updated_at`. However, only 2 of the 9 production callers (both in `import_json`) pass these parameters — the `add_node` method at line 265 does not, so `_add_node_raw` auto-generates timestamps there. This is functionally correct but means callers like `import_beliefs`, `import_agent`, and `sync_agent` always auto-generate timestamps even if the source data had them. The `reviewed_at`, `verified_at`, and `retracted_at` columns are not written by `_add_node_raw` — they default to NULL in PG. The PG `import_json` restores them in a separate UPDATE (lines 2355-2369), which works but is a different pattern than the SQLite path.

### reasons_lib/pg.py:retract_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Sets `retracted_at` and `updated_at` in both the "already OUT" and "transitioning to OUT" SQL UPDATE paths. Matches the network.py behavior. No PG-specific tests in the diff but test_pg.py exists at 1340 lines.

### reasons_lib/pg.py:assert_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Sets `retracted_at = NULL` (correct PG convention vs "" for SQLite) and `updated_at` to now. Behavior matches the in-memory network layer.

### reasons_lib/pg.py:show_node + export_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Both methods select all 5 timestamp columns and use `_fmt_ts` to convert PG `datetime` objects to ISO strings (or "" for NULL). The `_fmt_ts` helper is defined locally in each method — minor duplication but not a correctness issue.

### reasons_lib/pg.py:import_json
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Passes `created_at` and `updated_at` to `_add_node_raw`. Restores remaining timestamps (`reviewed_at`, `verified_at`, `retracted_at`, `updated_at`) in a separate UPDATE after truth-value fixup. The SQL is correctly parameterized — no injection risk from the `f-string` since `ts_col` values are from a hardcoded tuple.

### tests/test_api.py:TestLifecycleTimestamps
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Seven tests cover the core timestamp lifecycle: creation, update, metadata, retraction, show, export, and import roundtrip. Missing coverage for: (1) `assert_node` clearing `retracted_at`, (2) `verified_at` being set/preserved, (3) cascade effects on timestamps, (4) the `never_reviewed` and `not_reviewed_since` filters with the new `reviewed_at` field. The roundtrip test is solid — it exports a retracted node and verifies all three preserved timestamps match after import.

### tests/test_review.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Updated assertions correctly check `reviewed_at` instead of `metadata["last_reviewed"]`. However, the observation data shows an unchanged reference at `tests/test_review.py:739` (`assert "last_reviewed" in node`) that wasn't modified in this diff. If that assertion is checking `show_node` output (which no longer includes `last_reviewed`), it would fail. If it's checking `list_nodes` output (which still includes `last_reviewed` for backward compat), it passes. This should be verified.

### Migration: last_reviewed → reviewed_at
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The migration is intentionally gradual — `list_nodes` falls back to `metadata.get("last_reviewed")` and the output dict still uses the key `last_reviewed`. New writes go to `reviewed_at` exclusively. Old data is readable. However, the observation data flagged 12 stale references including `cli.py:1416-1417` which reads `node.get("last_reviewed")` from `list_nodes` output — this works but means the old key name persists in the public interface indefinitely. There's no migration path to move existing `metadata["last_reviewed"]` values into `reviewed_at` during database load/schema migration. Old nodes will have empty `reviewed_at` and populated `metadata["last_reviewed"]` permanently until individually re-reviewed.

### Self-Review
**Limitations:** - Could not see `_propagate` method body to verify whether cascade-affected nodes get timestamp updates — flagged as concern based on the retract/assert method bodies alone.
- Could not see full context around `tests/test_review.py:739` to determine whether the unchanged `assert "last_reviewed" in node` would break.
- Did not see `tests/test_pg.py` to assess PG-specific test coverage for the new timestamp columns.
- Could not verify whether `verified_at` is set anywhere outside this diff (it's defined but never automatically populated by any code path shown).

### Feature Requests
- Include the body of `_propagate` (or any method called by modified methods) when reviewing changes to retract/assert — cascade behavior is a critical correctness concern that can't be assessed from the diff alone.
- For migration-style changes (renaming a field), show unchanged references in test files with surrounding context (5 lines) so I can determine whether they'll break.
- When observations include `stale_references`, automatically include the surrounding context (function name, what dict is being accessed) to determine if they're actual bugs or false positives.

## gemini [PASS]

### reasons_lib/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Added five lifecycle timestamp fields (`created_at`, `updated_at`, `reviewed_at`, `verified_at`, `retracted_at`) as class attributes on the dataclass `Node` with empty string default values. Placing them at the end of the dataclass declaration avoids default/non-default ordering initialization bugs.

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly integrates timestamps into `show_node`, `update_node`, `set_metadata`, `export_network`, `import_json`, `list_nodes`, and `review_beliefs`. Uses timezone-aware UTC datetimes (`timezone.utc`) to prevent naive/aware comparisons. Restores exact timestamps in `import_json` *after* any truth value assertion/retraction transitions to avoid overwriting them. Backward compatibility for legacy metadata `last_reviewed` is preserved in both node listing and retrieval.

### reasons_lib/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Modifies `cmd_show` to dynamically extract non-empty timestamp keys, formats their names cleanly (e.g., `created_at` -> `Created`), and displays them at the end of the command output.

### reasons_lib/network.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly initializes `created_at` and `updated_at` automatically on `add_node` when not provided. Automatically updates `updated_at` and `retracted_at` on node retraction, and resets `retracted_at` to empty string while updating `updated_at` on node assertion.

### reasons_lib/pg.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Excellent implementation for the PostgreSQL database adapter. Columns are correctly added using native `TIMESTAMPTZ` datatype. Dynamic schema migration in `init_db` automatically alters existing `rms_nodes` tables if columns are missing. All query parameter mappings, variable unpackings, list/metadata updates, and JSON import roundtrips match perfectly and correctly.

### reasons_lib/storage.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly supports SQLite database adapter. The `nodes` schema is updated with columns using `TEXT DEFAULT ''` to cleanly store ISO8601 strings. Handles backwards-compatible dynamic database column existence detection in `_init_schema` and `load`, allowing old databases without timestamps to load and save flawlessly.

### tests/test_api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Added a robust new test suite (`TestLifecycleTimestamps`) checking and validating timestamp creation, propagation, update actions, retraction/assertion handling, CLI showing, and exported/imported JSON lossless roundtrips.

### tests/test_review.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Updates corresponding mock review assertions in tests from checking legacy `metadata["last_reviewed"]` to the new first-class `reviewed_at` attribute, keeping test-coverage aligned with the new schema.

### Self-Review
**Limitations:** None. Fully mapped class structures and database layers using `read_file` to ensure absolute correctness and complete backward-compatibility coverage.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#211
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons_lib/__init__.py — Node dataclass
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Five new `str` fields with empty-string defaults. Backward-compatible — existing callers that don't pass them get empty strings. Consistent with the existing `date` and `source_hash` patterns.

### reasons_lib/network.py:add_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Auto-generates `created_at` and `updated_at` when not provided (new nodes), passes caller-supplied values through for imports. Uses UTC throughout. The `if not created_at` guard correctly avoids overwriting imported timestamps.

### reasons_lib/network.py:retract
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`retract()` sets `retracted_at` on the explicitly-retracted node, but `_propagate()` (line 883–914) does NOT set `retracted_at` on dependents that go OUT via cascade. The test `test_cascade_does_not_set_retracted_at_on_dependents` asserts this is intentional — cascade-OUT dependents have `retracted_at == ""`. This is a defensible semantic choice (cascade != explicit retraction), but it creates an asymmetry: a node can be `truth_value == "OUT"` with `retracted_at == ""`, which means `retracted_at` is not a reliable indicator of OUT status. The intent should be documented or the field renamed to something like `explicitly_retracted_at` to avoid confusion. Minor concern, not a blocker.

### reasons_lib/network.py:assert_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clears `retracted_at` and bumps `updated_at` on assertion. Correctly mirrors the retract semantics. Test `test_assert_clears_retracted_at` validates this.

### reasons_lib/api.py:show_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All five timestamp fields added to the returned dict. Consumed by CLI `cmd_show`.

### reasons_lib/api.py:update_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Only bumps `updated_at` when `updated` list is non-empty — no spurious timestamp changes on no-op calls. Correct.

### reasons_lib/api.py:set_metadata
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Bumps `updated_at` unconditionally, which is correct since `set_metadata` always writes.

### reasons_lib/api.py:export_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All five timestamps included in export. Round-trip tested in `test_import_roundtrips_timestamps`.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Smart approach — passes `created_at`/`updated_at` to `add_node`, then restores `reviewed_at`, `verified_at`, `retracted_at`, and `updated_at` AFTER `retract()`/`assert_node()` calls that would overwrite them. The duplicate logic for the "no progress" fallback path is correct and mirrors the primary path. `updated_at` is restored last, which is the right order since retract/assert bump it.

### reasons_lib/api.py:list_nodes (reviewed_at / last_reviewed migration)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The backward-compat fallback `node.reviewed_at or node.metadata.get("last_reviewed")` is correct for reading. However, `list_nodes` still outputs the key as `"last_reviewed"` in its return dict (line 2350). This is an incomplete migration — it should output `"reviewed_at"` for consistency with `show_node`. The test at `tests/test_review.py:739` still asserts `"last_reviewed" in node` against `list_nodes` output, confirming this inconsistency persists. Similarly, `cli.py:1416-1417` reads `node.get("last_reviewed")` from the list output, which works but perpetuates the old name. This isn't broken, but it's a half-migration that will confuse callers.

### reasons_lib/api.py:review_beliefs
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Migrated from `metadata["last_reviewed"]` to `node.reviewed_at`. Also sets `updated_at`. Uses UTC. Tests updated to match (`test_review.py:631, 635, 649`).

### reasons_lib/storage.py — schema + save + load
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Schema migration via `ALTER TABLE` for older DBs is correct. `save()` writes all five columns. `load()` handles both old (no timestamp columns) and new schemas with the `has_timestamps` branch. The `or ""` fallback in load guards against SQLite NULLs. Solid backward compatibility. However, `test_storage.py` doesn't appear to have specific timestamp round-trip tests — coverage comes indirectly through `test_api.py`.

### reasons_lib/pg.py — Postgres backend
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

The Postgres backend is updated consistently with the SQLite path in most places, but `pg.py:add_node` (line 265) calls `_add_node_raw` WITHOUT passing `created_at` or `updated_at`. This means nodes created through the PG `add_node` path get their timestamps from `_add_node_raw`'s default logic, which does set them — so it works, but it's inconsistent with the pattern in `import_json` which explicitly passes them. More concerning: `import_beliefs` (line 2484) and `import_agent` (lines 2580, 2634) and `sync_agent` (lines 2710, 2895) all call `_add_node_raw` without `created_at`/`updated_at`, which means they get server-time defaults rather than any source timestamps. Also, the PG `import_json` doesn't restore `reviewed_at`/`verified_at`/`retracted_at`/`updated_at` after truth-value fixup in the same careful way — wait, checking... it does at line 2355-2370. Correction: the PG path looks correct for `import_json`. The `add_node` omission is the real issue but is functionally fine since `_add_node_raw` generates defaults.

### reasons_lib/cli.py:cmd_show
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Clean formatting: replaces `_at` suffix, title-cases, only prints non-empty timestamps. The label transform `"Created At" -> "Created"` is clear. No CLI tests for this display, but it's presentation-only.

### tests/test_api.py:TestLifecycleTimestamps
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

11 well-structured tests covering: creation, update, set_metadata, retraction, assertion (clear retracted_at), cascade behavior, show_node fields, export, import roundtrip, and verified_at roundtrip. The cascade test is particularly valuable — it validates the design choice that dependent nodes don't get `retracted_at` set. The `verified_at` test reaches into `Storage` directly to set the field, which is a pragmatic approach since there's no API to set it yet.

### tests/test_review.py — review_beliefs metadata tests
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Updated to assert on `node["reviewed_at"]` instead of `node["metadata"]["last_reviewed"]`. Dry-run test updated to `assert not node["reviewed_at"]`. Consistent with the migration.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass. Did not have access to the full `pg.py` test suite to confirm Postgres-path coverage. The `build/lib/` stale references were noted but not fully audited — they appear to be a stale build artifact and likely don't affect runtime. Could not see the `review_premises` function body to verify `datetime` import removal was safe (though the diff shows it removing an import that's now at module level).

## gemini [PASS]

### reasons_lib/__init__.py:Node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Dataclass fields `created_at`, `updated_at`, `reviewed_at`, `verified_at`, and `retracted_at` are successfully added to the `Node` structure with default empty string initializers to ensure backwards-compatible instantiation.

### reasons_lib/api.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Functional wrapper methods (`show_node`, `update_node`, `set_metadata`, `export_network`, `import_json`, `list_nodes`, `review_beliefs`) are successfully updated to handle the new first-class timestamps. In `import_json`, the original exported timestamps are restored immediately after `retract`/`assert_node` fixups, preventing the round-tripped timestamps from being overwritten with current times. Furthermore, `list_nodes` maintains backward compatibility with the legacy `last_reviewed` metadata field and handles timezone-aware cutoff comparisons correctly.

### reasons_lib/cli.py:cmd_show
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The CLI command `cmd_show` successfully prints non-empty timestamps using a clean, programmatic formatting loop (`ts_key.replace("_", " ").title().replace(" At", "")`) without breaking existing assertions in test files which expect specific command structures.

### reasons_lib/network.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Truth maintenance propagation correctly populates timestamps. `add_node` defaults the `created_at` and `updated_at` properties to current UTC timestamp on node creation. `retract` records the exact UTC retraction time on `retracted_at` and `updated_at`, and `assert_node` correctly clears `retracted_at` and updates `updated_at` upon node restoration.

### reasons_lib/pg.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

PostgreSQL-native database backend is fully updated. The `init_db` method performs incremental ALTER statement checks to add the new timestamp columns to `rms_nodes` if missing. SQL-based mutations (`_add_node_raw`, `retract_node`, `assert_node`, `update_node`, `set_metadata`, `import_json`) are cleanly adjusted to manipulate the columns directly in the DB.

### reasons_lib/storage.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

SQLite storage logic successfully executes schema migrations. It queries database schema structures dynamically via `PRAGMA table_info` and gracefully falls back to using empty strings when loading columns from legacy/unmigrated SQLite databases, ensuring robust backwards compatibility.

### tests/test_api.py:TestLifecycleTimestamps
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Detailed lifecycle testing class is added covering node addition, updates, metadata alterations, retractions, timestamp inclusion in exports, round-tripping through exports/imports, assertions, and verification that cascading retractions do not incorrectly taint dependents' direct `retracted_at` timestamps.

### tests/test_review.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Existing test suite assertions are seamlessly updated to assert the new first-class `reviewed_at` dataclass field instead of the legacy `last_reviewed` metadata field.

### Self-Review
**Limitations:** - Could not run live integration tests against a PostgreSQL database server since a configured Postgres service is unavailable in this test environment. However, SQL queries, column types (`TIMESTAMPTZ`), schema migrations, and column existence queries were rigorously verified via static code analysis.

PR #214, opened 2026-06-17
Add verify command to check beliefs against source documents

benthomasson:
## Summary

- New `reasons verify NODE` command reads the actual source file for a belief, sends it with the claim to an LLM, and returns CONFIRMED/STALE/PARTIAL/INCONCLUSIVE verdicts with verbatim quotes
- `--trace` follows derived beliefs to leaf premises and verifies each one
- `--retract` auto-retracts beliefs with STALE verdicts
- `--dry-run` previews what would be verified without LLM cost
- Stamps `verified_at` metadata on CONFIRMED beliefs
- Ported from `redhat-expert/scripts/verify_belief.py` prototype, using existing `resolve_source_path`, `trace_assumptions`, `invoke_model` infrastructure

## Test plan

- [x] `parse_verify_response` — valid JSON, prose-wrapped, malformed, empty, case-insensitive (6 tests)
- [x] `format_belief_for_verify` — with/without source (2 tests)
- [x] `read_source` — reads file, missing file, empty, truncation (4 tests)
- [x] API: premise verified, stale detected, retract stale, dry_run, trace derived, derived without trace, missing node (7 tests)
- [x] CLI: help, dry-run, basic verify (3 tests)
- [x] Full test suite passes (1537 tests)

### DIFF
--- a/reasons/api.py
+++ b/reasons/api.py
@@ -2696,6 +2696,8 @@ def verify_belief(
 
     verified = []
     stale = []
+    retract_failed = []
+    stamp_failed = []
     now = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
 
     for b, _ in beliefs_with_sources:
@@ -2712,7 +2714,7 @@ def verify_belief(
                     reason = verdict.get("reason", "stale per verify")
                     retract_node(bid, reason=f"verify: {reason}", **backend)
                 except Exception:
-                    pass
+                    retract_failed.append(bid)
 
     if not pg_conninfo and verified:
         try:
@@ -2721,10 +2723,12 @@ def verify_belief(
                     if bid in net.nodes:
                         net.nodes[bid].verified_at = now
                         net.nodes[bid].updated_at = now
+                    else:
+                        stamp_failed.append(bid)
         except Exception:
-            pass
+            stamp_failed = list(verified)
 
-    return {
+    result = {
         "results": verdicts,
         "verified": verified,
         "stale": stale,
@@ -2732,6 +2736,11 @@ def verify_belief(
         "is_derived": has_justifications,
         "dry_run": False,
     }
+    if retract_failed:
+        result["retract_failed"] = retract_failed
+    if stamp_failed:
+        result["stamp_failed"] = stamp_failed
+    return result
 
 
 NEGATIVE_BATCH_SIZE = 50
--- a/reasons/api.py
+++ b/reasons/api.py
@@ -2616,6 +2616,119 @@ def report_belief(
     }
 
 
+def verify_belief(
+    node_id: str,
+    trace: bool = False,
+    retract: bool = False,
+    dry_run: bool = False,
+    model: str = "claude",
+    timeout: int = 120,
+    db_path: str = DEFAULT_DB,
+    pg_conninfo=None, project_id=None,
+) -> dict:
+    """Verify a belief against its source documents.
+
+    If trace=True and the belief is derived, traces to leaf premises and
+    verifies each one.  Stamps verified_at on CONFIRMED beliefs.  If
+    retract=True, retracts STALE beliefs.
+
+    Returns: {"results": dict[id, verdict_dict], "verified": list[str],
+              "stale": list[str], "beliefs_checked": list[dict],
+              "is_derived": bool, "dry_run": bool}
+    """
+    from .verify import read_source, verify_beliefs
+    from datetime import datetime, timezone
+
+    backend = dict(db_path=db_path, pg_conninfo=pg_conninfo,
+                   project_id=project_id)
+
+    detail = show_node(node_id, **backend)
+    has_justifications = bool(detail.get("justifications"))
+
+    if has_justifications and trace:
+        trace_data = trace_assumptions(node_id, **backend)
+        premise_ids = trace_data["premises"]
+        beliefs_to_check = []
+        seen = set()
+        for pid in premise_ids:
+            if pid in seen:
+                continue
+            seen.add(pid)
+            try:
+                pd = show_node(pid, **backend)
+                beliefs_to_check.append({
+                    "id": pid,
+                    "text": pd.get("text", ""),
+                    "truth_value": pd.get("truth_value", ""),
+                    "source": pd.get("source", ""),
+                    "source_url": pd.get("source_url", ""),
+                })
+            except (KeyError, PermissionError):
+                beliefs_to_check.append({
+                    "id": pid, "text": "", "truth_value": "UNKNOWN",
+                    "source": "", "source_url": "",
+                })
+    else:
+        beliefs_to_check = [{
+            "id": node_id,
+            "text": detail.get("text", ""),
+            "truth_value": detail.get("truth_value", ""),
+            "source": detail.get("source", ""),
+            "source_url": detail.get("source_url", ""),
+        }]
+
+    beliefs_with_sources = []
+    for b in beliefs_to_check:
+        source_content = read_source(b["source"], db_path=db_path) if b["source"] else None
+        beliefs_with_sources.append((b, source_content))
+
+    if dry_run:
+        return {
+            "results": {},
+            "verified": [],
+            "stale": [],
+            "beliefs_checked": beliefs_to_check,
+            "is_derived": has_justifications,
+            "dry_run": True,
+        }
+
+    verdicts = verify_beliefs(beliefs_with_sources, model=model, timeout=timeout)
+
+    verified = []
+    stale = []
+    now = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+
+    for b, _ in beliefs_with_sources:
+        bid = b["id"]
+        verdict = verdicts.get(bid, {})
+        v = verdict.get("verdict", "INCONCLUSIVE")
+
+        if v == "CONFIRMED":
+            verified.append(bid)
+            try:
+                set_metadata(bid, "verified_at", now, **backend)
+            except Exception:
+                pass
+
+        elif v == "STALE":
+            stale.append(bid)
+            if retract:
+                try:
+                    reason = verdict.get("reason", "stale per verify")
+                    retract_node(bid, reason=f"verify: {reason}", **backend)
+                except Exception:
+                    pass
+
+    return {
+        "results": verdicts,
+        "verified": verified,
+        "stale": stale,
+        "beliefs_checked": beliefs_to_check,
+        "is_derived": has_justifications,
+        "dry_run": False,
+    }
+
+
 NEGATIVE_BATCH_SIZE = 50
 
 NEGATIVE_TERMS = [
--- a/reasons/api.py
+++ b/reasons/api.py
@@ -2705,11 +2705,6 @@ def verify_belief(
 
         if v == "CONFIRMED":
             verified.append(bid)
-            try:
-                set_metadata(bid, "verified_at", now, **backend)
-            except Exception:
-                pass
-
         elif v == "STALE":
             stale.append(bid)
             if retract:
@@ -2718,6 +2713,16 @@ def verify_belief(
                     retract_node(bid, reason=f"verify: {reason}", **backend)
                 except Exception:
                     pass
+
+    if not pg_conninfo and verified:
+        try:
+            with _with_network(db_path, write=True) as net:
+                for bid in verified:
+                    if bid in net.nodes:
+                        net.nodes[bid].verified_at = now
+                        net.nodes[bid].updated_at = now
+        except Exception:
+            pass
 
     return {
         "results": verdicts,
--- a/reasons/cli.py
+++ b/reasons/cli.py
@@ -1571,6 +1571,11 @@ def cmd_verify(args):
               f"{len(stale)} stale, {len(partial)} partial, "
               f"{len(inconclusive)} inconclusive")
 
+    if result.get("retract_failed"):
+        print(f"\n  WARNING: failed to retract: {', '.join(result['retract_failed'])}", file=sys.stderr)
+    if result.get("stamp_failed"):
+        print(f"\n  WARNING: failed to stamp verified_at: {', '.join(result['stamp_failed'])}", file=sys.stderr)
+
     print(f"\nResults: {len(confirmed)} confirmed, {len(stale)} stale, "
           f"{len(partial)} partial, {len(inconclusive)} inconclusive")
 
--- a/reasons/cli.py
+++ b/reasons/cli.py
@@ -1512,6 +1512,73 @@ def cmd_report(args):
             print(f"  {cost}", file=sys.stderr)
 
 
+def cmd_verify(args):
+    from .llm import reset_cost_tracker, format_cost_summary
+    reset_cost_tracker()
+
+    result = api.verify_belief(
+        args.node_id,
+        trace=getattr(args, "trace", False),
+        retract=getattr(args, "retract", False),
+        dry_run=getattr(args, "dry_run", False),
+        model=getattr(args, "model", "claude") or "claude",
+        timeout=args.timeout,
+        **_backend_kwargs(args),
+    )
+
+    beliefs = result["beliefs_checked"]
+
+    if result["dry_run"]:
+        print(f"--dry-run: {len(beliefs)} belief(s) would be verified:")
+        for b in beliefs:
+            has_src = "yes" if b.get("source") else "NO"
+            print(f"  {b['id']} [source: {has_src}]")
+        return
+
+    markers = {
+        "CONFIRMED": "+", "STALE": "-",
+        "PARTIAL": "~", "INCONCLUSIVE": "?",
+    }
+    verdicts = result["results"]
+
+    confirmed, stale, partial, inconclusive = [], [], [], []
+    for b in beliefs:
+        bid = b["id"]
+        v = verdicts.get(bid, {})
+        verdict = v.get("verdict", "INCONCLUSIVE")
+        reason = v.get("reason", "")
+        quote = v.get("quote")
+        marker = markers.get(verdict, "?")
+
+        print(f"\n  [{marker}] {verdict}: {bid}")
+        if reason:
+            print(f"      {reason}")
+        if quote:
+            print(f'      Quote: "{quote}"')
+
+        if verdict == "CONFIRMED":
+            confirmed.append(bid)
+        elif verdict == "STALE":
+            stale.append(bid)
+        elif verdict == "PARTIAL":
+            partial.append(bid)
+        else:
+            inconclusive.append(bid)
+
+    if result["is_derived"] and getattr(args, "trace", False):
+        total = len(beliefs)
+        print(f"\n  Antecedent summary: {len(confirmed)}/{total} confirmed, "
+              f"{len(stale)} stale, {len(partial)} partial, "
+              f"{len(inconclusive)} inconclusive")
+
+    print(f"\nResults: {len(confirmed)} confirmed, {len(stale)} stale, "
+          f"{len(partial)} partial, {len(inconclusive)} inconclusive")
+
+    cost = format_cost_summary()
+    if cost:
+        print(f"  {cost}", file=sys.stderr)
+
+
 def cmd_list_negative(args):
     result = api.list_negative(
         visible_to=_parse_visible_to(args),
@@ -2594,6 +2661,19 @@ def main():
                    help="LLM model for narrative synthesis (default: structured report)")
     p.add_argument("--timeout", type=int, default=300, help="LLM timeout in seconds (default: 300)")
 
+    # verify
+    p = sub.add_parser("verify", help="Verify beliefs against their source documents")
+    p.add_argument("node_id", help="Belief ID to verify")
+    p.add_argument("--trace", action="store_true",
+                   help="Trace derived beliefs to leaf premises and verify each")
+    p.add_argument("--retract", action="store_true",
+                   help="Retract beliefs with STALE verdicts")
+    p.add_argument("--dry-run", action="store_true",
+                   help="Show what would be verified without calling LLM")
+    p.add_argument("-m", "--model", default="claude",
+                   help="LLM model to use (default: claude)")
+    p.add_argument("--timeout", type=int, default=120, help="LLM timeout in seconds (default: 120)")
+
     p = sub.add_parser("list-negative", help="Find IN beliefs describing problems/defects/risks (LLM-classified)")
     p.add_argument("--visible-to", metavar="TAG,TAG", help="Only show nodes whose access_tags are a subset of these tags")
     p.add_argument("-m", "--model", default=None,
@@ -2852,6 +2932,7 @@ def main():
         "list-gated": cmd_list_gated,
         "report-gated": cmd_report_gated,
         "report": cmd_report,
+        "verify": cmd_verify,
         "list-negative": cmd_list_negative,
         "topics": cmd_topics,
         "build-wiki": cmd_build_wiki,
--- a/reasons/verify.py
+++ b/reasons/verify.py
@@ -105,8 +105,11 @@ def parse_verify_response(response):
     for k, v in data.items():
         if not isinstance(v, dict) or "verdict" not in v:
             continue
+        raw_verdict = v["verdict"]
+        if not isinstance(raw_verdict, str):
+            continue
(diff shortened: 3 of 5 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#214
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **reasons/verify.py** [LOW]: claude: PASS, gemini: CONCERN

## claude [CONCERN]

### reasons/verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean module with clear separation of concerns. `read_source` correctly delegates to `resolve_source_path` and handles the `db_path` → `db_dir` conversion. `parse_verify_response` uses a greedy regex `r"\{.*\}"` with `re.DOTALL` which correctly captures the outermost JSON object, and gracefully returns `{}` on any parse failure. The `VERIFY_PROMPT` is well-structured with explicit rules about rigor and verbatim quoting. `format_belief_for_verify` handles missing sources cleanly. All four functions are tested.

### reasons/api.py:verify_belief
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function follows the same structural pattern as `report_belief` (line 2564–2616), which is good. Two concerns:

1. **Silent `except Exception: pass`** on lines 2710 and 2719 around `set_metadata` and `retract_node`. The `set_metadata` swallow is acceptable (non-critical metadata stamp), but the `retract_node` swallow is riskier — if retraction fails (e.g. database locked, node already deleted), the function returns the belief in the `stale` list implying it was retracted when `retract=True`, but it wasn't. At minimum, log a warning. Compare with `retract_node` itself (line 361) which doesn't defend callers this way.

2. **`show_node` raises `KeyError` for the initial lookup** (line 2647) which correctly propagates to the caller (tested in `test_missing_node_raises`). But inside the trace loop (line 2659), `KeyError` from `show_node` is caught and the belief is added with `text=""` and `truth_value="UNKNOWN"`. This matches the `report_belief` pattern, which is fine, but a premise that can't be loaded will be sent to the LLM with empty text — the LLM will return INCONCLUSIVE, which is the right behavior.

### reasons/cli.py:cmd_verify
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Follows the same pattern as `cmd_report` — `reset_cost_tracker` at top, `format_cost_summary` at bottom. The `getattr(args, "trace", False)` calls are unnecessary since argparse sets the attributes, but harmless and consistent with defensive coding elsewhere in the CLI. Output formatting with `[+]`/`[-]`/`[~]`/`[?]` markers is clear. The antecedent summary line (line 1572) correctly gates on `is_derived and trace`. CLI parser (line 2665) correctly registers the subcommand with all expected flags, and the dispatch table (line 2935) is updated.

### reasons/cli.py (parser registration, lines 2664–2676)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Argument parser correctly uses `action="store_true"` for boolean flags, provides sensible defaults (model="claude", timeout=120), and includes the `--dry-run` flag with `store_true`. The dispatch entry at line 2935 maps `"verify"` to `cmd_verify`. No missing args versus what `api.verify_belief` accepts.

### reasons_lib/verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Standard backwards-compatibility shim matching the pattern used by all other `reasons_lib/` modules (confirmed by project structure in CLAUDE.md). The wildcard re-export is appropriate here since the shim exists solely for import compatibility. No dedicated test for the shim, but the pattern is well-established in this project and the `reasons.verify` imports are tested directly.

### tests/test_verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test suite covering all layers — unit tests for `parse_verify_response` (6 cases including edge cases), `format_belief_for_verify` (2 cases), `read_source` (4 cases including truncation), API-level `verify_belief` (7 cases), and CLI-level (3 cases). The mocking strategy (`patch("reasons.llm.subprocess.run")`) is consistent with other test files in the project. The `_mock_result` helper cleanly simulates LLM subprocess output. Test for `test_retract_stale` properly verifies the side effect by checking `show_node` truth_value after retraction. One gap: no test for the `except Exception: pass` paths in `verify_belief` (set_metadata failure, retract_node failure), though these are defensive catches.

### Self-Review
**Limitations:** Could not see the full `show_node` return schema to verify all field names used in the belief dict construction (source, source_url, truth_value). Could not verify whether `invoke_model` might raise exceptions that should be handled at the `verify_belief` level rather than letting them propagate uncontrolled. Could not run the tests to confirm they pass.

### Feature Requests
- Include the return schema of called functions (e.g. `show_node` return dict keys) when they're used to build data structures in the reviewed code
- Show existing test patterns (e.g. from `test_report.py`) alongside new test files to verify consistency

## gemini [CONCERN]

### reasons/api.py:verify_belief
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

There is an inconsistency in how node verification timestamps are recorded. The `Node` dataclass and database schemas in `storage.py` and `pg.py` declare a first-class `verified_at` attribute/column. However, `verify_belief` calls `set_metadata(bid, "verified_at", now, **backend)`, which only writes the timestamp into the JSON dictionary inside the `metadata` column. This leaves the database column `verified_at` and the `node.verified_at` attribute empty, causing `show_node` to return an empty string for `"verified_at"`. To align with how `reviewed_at` is directly set in the review flow, the field should be set directly on the `Node` object within a database transaction.

### reasons/verify.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

In `parse_verify_response()`, if the LLM output contains a verdict field with a non-string value (such as `null` or an integer), calling `v["verdict"].upper()` will raise an `AttributeError` and crash the verification process. The JSON parser is wrapped in `try/except json.JSONDecodeError` but the loop body is not protected. The code should verify that the verdict is a string before calling `.upper()`, or coerce it safely using `str()`. Additionally, using greediness in the regex `r"\{.*\}"` can cause parsing failures if the LLM output includes any trailing text containing brace characters `{}`.

### reasons/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The CLI integration is well-designed. It correctly resets the cost tracker before invoking the API, safely retrieves configuration defaults, and maps results to user-friendly output markers (e.g., `+` for CONFIRMED, `-` for STALE). It prints cost summaries correctly and handles dry runs cleanly. Argument parsing integration matches standard CLI patterns in the codebase.

### reasons_lib/verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

This is a clean and standard wildcard export shim for backward-compatibility, matching the established design pattern of the `reasons_lib` package.

### tests/test_verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The test suite is highly comprehensive, covering edge cases like LLM JSON parsing within surrounding prose, source file truncation, dry runs, premise backtracking (`trace=True`), and CLI error/help flags. It uses standard `unittest.mock.patch` to mock LLM subprocess calls cleanly without requiring network access.

### Self-Review
**Limitations:** No live execution capability was available during the review, but thorough static analysis successfully identified type-safety and schema inconsistencies.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#214
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons/verify.py (new file)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `parse_verify_response` regex `r"\{.*\}"` with `re.DOTALL` is greedy. If the LLM response contains prose after the JSON (e.g., `{...} and also {"note": ...}`), the match extends from the first `{` to the last `}` in the entire response, which may not be valid JSON. This is a known fragility with greedy extraction. The existing `json.JSONDecodeError` catch handles the failure gracefully (returns `{}`), so it's not a crash bug, but it can cause silent data loss — the LLM returns a valid response that gets silently dropped because the greedy match over-captures. Consider using `re.search(r"\{[^{}]*(?:\{[^{}]*\}[^{}]*)*\}", ...)` or parsing from the first `{` with bracket counting. The `read_source` function and `format_belief_for_verify` are clean. The prompt template is well-structured with clear instructions and appropriate rigor level.

### reasons/api.py:verify_belief
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Three concerns:

1. **Source reading uses `db_path` even on pg backend** (line 2682): `read_source(b["source"], db_path=db_path)` is called regardless of whether `pg_conninfo` is set. When using PostgreSQL, `db_path` defaults to `DEFAULT_DB` (a local path), and `resolve_source_path` uses `Path(db_path).parent` as the directory to resolve relative source paths. This could resolve to the wrong directory if the pg-backed belief references source files relative to a different project root. Not a crash, but silently reads wrong files or returns None.

2. **`verified_at` stamping skipped for pg** (line 2717): The `if not pg_conninfo and verified` guard intentionally skips stamping for pg, but there's no TODO or fallback — pg-backed verified beliefs never get a `verified_at` timestamp. The `report_belief` function (same file) has a similar gap pattern, so this appears to be accepted technical debt, but worth noting.

3. **Bare `except Exception: pass`** (lines 2714, 2724): Two broad exception swallows. The retraction one (2714) silently eats failures to retract stale beliefs — the user asked for `--retract` and it may silently fail. The `verified_at` stamping one (2724) similarly swallows. At minimum, these should log or include the failed IDs in the result dict so callers can tell something went wrong.

### reasons/cli.py:cmd_verify
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Clean CLI handler following the same pattern as `cmd_report`. The `getattr(args, "trace", False)` guards are defensive but unnecessary since argparse always sets these attributes — harmless though. The output formatting is clear with markers (`+`, `-`, `~`, `?`). The antecedent summary only prints when `is_derived and trace`, which is correct. CLI test coverage exists for `--help`, `--dry-run`, and basic verify, but there's no CLI test for `--trace` or `--retract` flags (the API tests cover those paths, but integration through the CLI is untested for those flags).

### reasons/cli.py (parser registration)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Parser args are correctly defined. The `verify` command is added to the dispatch dict. The `--dry-run` uses `store_true` correctly. Default timeout of 120 matches the API default. All flags match the `verify_belief` API signature.

### reasons_lib/verify.py (backwards-compat shim)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Follows the established pattern from the `reasons_lib` -> `reasons` rename (every module in `reasons/` has a corresponding shim in `reasons_lib/`). Star import with noqa comments matches the other shims.

### tests/test_verify.py (new file)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Thorough test suite covering: JSON parsing (valid, malformed, empty, missing verdict, case insensitivity), formatting, source reading (file exists, missing, empty, truncation), API-level verification (premise confirmed, stale detected, retract-stale, dry-run, trace-derived, derived-without-trace, missing node), and CLI-level testing (help, dry-run, basic verify). The mocking approach using `subprocess.run` patches is consistent with how `invoke_model` works internally. One minor gap: no test verifies that `verified_at` is actually stamped on the node after a CONFIRMED verdict — the API test `test_premise_verified` checks return values but doesn't read the node back to confirm the timestamp was persisted.

### Self-Review
**Limitations:** Could not run the test suite to verify tests pass (permission not granted). Could not confirm the greedy regex concern with a concrete repro. Did not trace the full `invoke_model` -> `subprocess.run` path to verify the mock setup is accurate for all model types. Did not check whether `pg_conninfo` path for `verify_belief` has any pg dispatch equivalent that could handle `verified_at` stamping server-side.

### Feature Requests
- Include callers of modified/new API functions to verify all integration paths
- Auto-run the test suite as part of the review and include results
- Flag bare `except Exception: pass` patterns automatically as a lint concern

## gemini [CONCERN]

### reasons/verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The core verification logic is robust and correct. It handles path resolution via existing check_stale infrastructure, truncates large files (`MAX_SOURCE_CHARS = 30_000`) to avoid token limits, and gracefully catches IO/decoding errors (e.g. `IsADirectoryError` or encoding issues via `errors="replace"`). The regex fallback in JSON parsing safely extracts the JSON payload from the surrounding LLM markdown code blocks or prose.
*Note/Minor Suggestion:* `verify_beliefs` should check if `beliefs_with_sources` is empty and immediately return `{}` instead of unnecessarily formatting a prompt and invoking the model with an empty list of beliefs.

### reasons/api.py:verify_belief
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

- **PostgreSQL Integration Gap (Critical):** While the method works correctly for SQLite-backed databases, if `pg_conninfo` is set, the confirmed beliefs never have their `verified_at` or `updated_at` fields updated. The timestamp writing block is guarded by `if not pg_conninfo and verified:`, and there is no fallback query executed on the PostgreSQL connection to set these timestamps. This represents a missing feature/integration gap for PostgreSQL backend deployments.
- **Robust Exception Handling:** Tracing premises, fetching nodes, handling missing elements with `UNKNOWN` fallback values, and swallowing exceptions during retraction cascades is correctly implemented and aligned with the patterns in the repository.

### reasons/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The CLI subcommand `verify` is correctly configured and wired up. It provides interactive and readable terminal feedback mapping LLM verdicts to ASCII markers (`[+]`, `[-]`, `[~]`, `[?]`), prints antecedent summaries for traced derived beliefs, and displays LLM execution/cost statistics safely on `stderr`.

### reasons_lib/verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

A clean, standard backwards-compatibility shim that exports all symbols from `reasons.verify`.

### tests/test_verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Exceptional unit test coverage. It comprehensively tests CLI invocation, parser robustness against malformed/varying JSON outputs, API verification behavior, tracing logic, retraction triggers, file truncation, and edge cases with missing source documents. LLM execution is successfully and cleanly mocked out via `subprocess.run` and `shutil.which` patches to guarantee fast and deterministic local execution.
*Note/Minor Suggestion:* No tests verify behavior on PostgreSQL; once PostgreSQL support is completed, testing database dispatcher behavior would be highly recommended.

### Self-Review
**Limitations:** - Did not have access to run `pytest` locally to confirm the exact execution within a real Python environment since the `run_shell_command` tool was not available in this turn.
- Relied on static analysis to verify that the mock LLM outputs match the exact structure expected by `_parse_cli_json` in `reasons/llm.py`.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#214
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons/verify.py (new file)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean module with clear separation of concerns. `read_source` correctly delegates to `resolve_source_path` and handles errors. `parse_verify_response` is defensive — handles missing/malformed JSON, non-dict entries, missing `verdict` keys, and normalizes case. `format_belief_for_verify` produces well-structured prompts. `verify_beliefs` ties them together cleanly. The `MAX_SOURCE_CHARS = 30_000` truncation is a reasonable guard. One minor note: the regex `r"\{.*\}"` with `re.DOTALL` grabs from the first `{` to the last `}` — this works for single-object responses but would fail if the LLM returned two separate JSON blocks. Given the prompt asks for a single object, this is acceptable.

### reasons/api.py:verify_belief
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Several concerns:

1. **PostgreSQL `verified_at` silently skipped** (line 2719): The guard `if not pg_conninfo and verified:` means PostgreSQL-backed beliefs never get their `verified_at` stamped after verification. The retraction path (line 2715) correctly passes `**backend` which includes `pg_conninfo`, so retraction works for PG — but the timestamp stamping is SQLite-only. This is a silent data loss for PG users. Should either implement PG stamping or raise/log/return a warning.

2. **Two separate `_with_network` calls for the same DB** — `show_node` at line 2648 opens and closes the network, then the `verified_at` stamping at line 2721 opens it again. The retraction at line 2715 opens it a third time. Each call does a full load/save cycle. This is correct but inefficient for a premise with many nodes to verify. For a single belief it's fine.

3. **Broad `except Exception` silencing** (lines 2716, 2728): Both the retraction and stamping blocks catch all exceptions silently, only recording the failure in the return dict. If the DB is corrupt or the network is in a bad state, this masks it. The caller (CLI) doesn't check `retract_failed` or `stamp_failed` at all — `cmd_verify` never prints these fields, so the user gets no indication of failure.

4. **`source` field used as path** — `read_source(b["source"], db_path=db_path)` passes the raw `source` string from `show_node`. Looking at `resolve_source_path`, this works when `source` is a relative path resolvable from `db_dir`. But the tests pass absolute paths to `source` (e.g., `str(source_file)`) which `resolve_source_path` resolves because `Path(source).exists()` catches absolute paths at the single-component fallback on line 51 of `check_stale.py`. This works but is worth noting — the tests exercise a different resolution path than production likely will.

### reasons/cli.py:cmd_verify
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

1. **`retract_failed` and `stamp_failed` not surfaced** — The API returns these keys on failure, but `cmd_verify` never checks or prints them. A user whose beliefs failed to retract or stamp gets no indication. This is the more significant half of the concern raised above.

2. **Redundant categorization** — `cmd_verify` re-categorizes beliefs into `confirmed/stale/partial/inconclusive` lists (lines 1551–1560) even though `result["verified"]` and `result["stale"]` already provide the confirmed and stale lists. The local re-categorization is for display, which is fine, but the `partial` and `inconclusive` lists from the CLI side are never compared against anything from the API — they only exist for the summary line. This is harmless but slightly duplicative.

3. **`getattr` usage** — `getattr(args, "trace", False)` is used throughout even though the argparse parser guarantees these attributes exist. Not a bug, just unnecessary defensive coding (probably cargo-culted from other `cmd_*` functions).

### reasons/cli.py (argparse wiring)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Parser is correctly defined with all flags (`--trace`, `--retract`, `--dry-run`, `-m/--model`, `--timeout`). Command is registered in the dispatch dict at line 2935. The `--dry-run` flag uses `store_true` and `dest` defaults to `dry_run` (argparse converts hyphens). Help text is clear.

### reasons_lib/verify.py (backwards-compat shim)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Follows the existing pattern from `reasons_lib/__init__.py` which does the same `from reasons import *` re-export. No test needed for a one-line shim.

### tests/test_verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test coverage — 16 tests covering: parse_verify_response (valid, prose-wrapped, malformed, empty, missing verdict, case normalization), format_belief_for_verify (with/without source), read_source (hit, miss, empty, truncation), API-level verify (premise, verified_at stamp, stale detection, retract-on-stale, dry_run, trace-derived, derived-without-trace, missing node), and CLI-level (help, dry-run, basic verify). The mocking strategy (`_mock_result` + patching `subprocess.run` and `shutil.which`) matches the existing test patterns in the codebase. One gap: no test for the `retract_failed` or `stamp_failed` error paths, which compounds the concern above about those being silently swallowed.

### Self-Review
**Limitations:** Could not run the test suite to verify all tests pass (permission denied). Did not see the full `_pg_dispatch` implementation to know if PG stamping could be trivially added. Did not check if other CLI commands in the codebase similarly ignore error fields from the API, which would indicate this is an accepted pattern.

### Feature Requests
- Include the output of running the test suite in the observations, so the reviewer can confirm tests pass
- Show other `cmd_*` functions' error-handling patterns to assess whether ignoring error fields is a codebase norm or an oversight
- Flag when API return fields are never read by any caller (dead data detection)

## gemini [CONCERN]

### reasons/api.py:verify_belief
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

The function correctly implements belief verification with support for tracing to premises, dry runs, retraction of stale nodes, and stamping verified nodes with `verified_at` and `updated_at`. However, there are two key concerns:
1. **PostgreSQL/pg backend limitation**: If `pg_conninfo` is set, the verified/updated stamps (`verified_at` and `updated_at`) are never written to the PostgreSQL database because postgres-based updates for these fields are not implemented. The updates are silently skipped, and `stamp_failed` is not populated to reflect this.
2. **Redundant LLM Call**: If `trace=True` but the node has no trace premises (e.g. empty tracing output), `beliefs_with_sources` will be empty, leading to a redundant LLM call with a blank belief block. It should return an empty dict immediately without calling `verify_beliefs`.

### reasons/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new `verify` sub-command and its handler `cmd_verify` are correctly registered in the argparse CLI and properly integrated with cost tracking, backend argument resolution, and verdict markers matching existing CLI output patterns.

### reasons/verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The file contains clean and modular helper functions for handling source document resolution/reading, LLM prompting, regex-based JSON response parsing, and prompt formatting. The truncation handling (`MAX_SOURCE_CHARS` limit) is robust, and the JSON parser uses defensive fallback checks to handle model noise and markdown block formatting gracefully.

### reasons_lib/verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Standard backwards-compatibility shim matching the pattern used elsewhere in the `reasons_lib/` directory.

### tests/test_verify.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

High-quality unit and integration tests covering parser behavior (JSON edge cases), source reading (missing file, file path resolution, truncation), API-level verification (tracing, retraction, stamps), and CLI-level execution. Mocking of LLM responses is well-designed.

### Self-Review
**Limitations:** None. All modified functions and new files were examined in their entirety using standard search tools and direct reads.

PR #222, opened 2026-06-24
Add --parallel option to review-justifications

benthomasson:
## Summary
- Extracts per-batch LLM work into `_process_batch()` helper
- Adds `--parallel N` option using `ThreadPoolExecutor` + `as_completed`, matching the existing pattern in `review-premises`, `repair-premises`, and `build-wiki`
- Default `0` = sequential (no behavior change)

## Test plan
- [x] New `test_parallel` test with `batch_size=1, parallel=2` verifies concurrent execution produces correct results
- [x] All 18 tests pass

### DIFF
--- a/reasons/api.py
+++ b/reasons/api.py
@@ -3020,6 +3020,7 @@ def review_justifications(
     timeout: int = 300,
     min_antecedents: int = 2,
     on_batch: Callable | None = None,
+    parallel: int = 0,
     db_path: str = DEFAULT_DB,
 ) -> dict:
     """Review SL justifications for ALL vs ANY misclassification.
@@ -3049,7 +3050,8 @@ def review_justifications(
     total_candidates = len(candidates)
 
     results = _review(nodes, belief_ids=belief_ids, model=model, timeout=timeout,
-                      min_antecedents=min_antecedents, on_batch=on_batch)
+                      min_antecedents=min_antecedents, on_batch=on_batch,
+                      parallel=parallel)
 
     convert_any = sum(1 for r in results if r.get("classification") == "ANY")
     convert_mixed = sum(1 for r in results if r.get("classification") == "MIXED")
--- a/reasons/cli.py
+++ b/reasons/cli.py
@@ -1793,6 +1793,7 @@ def cmd_review_justifications(args):
         model=model,
         timeout=args.timeout,
         min_antecedents=args.min_antecedents,
+        parallel=args.parallel,
         db_path=args.db,
     )
 
@@ -2810,6 +2811,8 @@ def main():
                    help="Model to use (default: claude)")
     p.add_argument("--timeout", type=int, default=600,
                    help="LLM timeout in seconds (default: 600)")
+    p.add_argument("--parallel", type=int, default=0,
+                   help="Number of concurrent LLM workers (default: 0 = sequential)")
     p.add_argument("-o", "--output", default=None,
                    help="Write findings to markdown file")
 
--- a/reasons/review_justifications.py
+++ b/reasons/review_justifications.py
@@ -95,9 +95,22 @@ def parse_justification_review(response):
     return []
 
 
+def _process_batch(batch, nodes, model, timeout):
+    """Process a single batch of beliefs through the LLM."""
+    beliefs_text = "\n\n".join(
+        format_belief_for_review(nid, nodes) for nid in batch
+    )
+    prompt = REVIEW_JUSTIFICATIONS_PROMPT.format(beliefs=beliefs_text)
+    response = invoke_model(prompt, model=model, timeout=timeout)
+    results = parse_justification_review(response)
+    if not results and batch:
+        print(f"  WARN: batch returned no parseable results", file=sys.stderr)
+    return results
+
+
 def review_justifications(nodes, belief_ids=None, model="claude", timeout=300,
                           batch_size=REVIEW_BATCH_SIZE, min_antecedents=2,
-                          on_batch=None):
+                          on_batch=None, parallel=0):
     """Review SL justifications for ALL vs ANY classification.
 
     Args:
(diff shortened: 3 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#222
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons/review_justifications.py:_process_batch
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Clean extraction of the batch-processing logic into a standalone function. Follows the exact same pattern as `_process_batch` in `review_premises.py:114`. The warning message loses the batch number (prints "batch returned no parseable results" instead of "batch N returned..."), but this is minor — the parallel path prints batch info on completion anyway.

### reasons/review_justifications.py:review_justifications
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The parallel implementation is structurally correct and mirrors `review_premises.py` exactly. However, there is a **thread-safety concern with `on_batch`**: in the parallel path, `all_results.extend(results)` and the `on_batch(all_results)` call happen in the main thread's `as_completed` loop, so there's no data race on `all_results` itself — this is safe. The `on_batch` callback receives a snapshot of cumulative results in non-deterministic order, which is acceptable given the parallel semantics. The `parallel` parameter is not documented in the docstring (the docstring still omits it, matching `review_premises` which also omits it — consistent but imperfect). Overall, the implementation is correct and follows the established pattern.

### reasons/api.py:review_justifications
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `parallel` parameter is correctly threaded from the public API down to the internal `_review` call. The signature addition is clean and the default of `0` preserves backward compatibility.

### reasons/cli.py:cmd_review_justifications + argparse
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The `--parallel` argument is correctly added to argparse and threaded through `cmd_review_justifications` to `api.review_justifications`. The default of `0` means sequential, matching the help text. Consistent with how `--parallel` is likely handled in other review commands. No CLI-level test exists but this is a thin pass-through.

### tests/test_review_justifications.py:test_parallel
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The test verifies that parallel execution produces the right results from both batches. It uses `batch_size=1` to force two batches with `parallel=2`, which is good. However, the test doesn't verify that threads were actually used — it would pass identically with sequential execution (if the `parallel` flag were ignored). A more robust test would assert that `invoke_model` was called concurrently (e.g., via a threading barrier or checking call timing). Additionally, there's no test for: (1) error handling in the parallel path (a batch that raises), (2) `on_batch` callback behavior under parallelism, (3) `parallel=1` (single-worker thread pool, which exercises the threading path but with no concurrency). These are moderate gaps.

### Self-Review
**Limitations:** Could not run the test suite to confirm all tests pass (permission denied on `uv run`). The observation results failed to resolve `_review` and `invoke_model` — I read the files directly to compensate. I did not see the full `review_premises.py` parallel tests to compare coverage patterns.

### Feature Requests
- Include the full function body of all modified functions in the observation results, not just caller locations — the `_review` and `invoke_model` resolution failures were unhelpful but reading files directly worked around it.
- When a new feature mirrors an existing pattern (e.g. parallel was already in `review_premises.py`), automatically include the reference implementation for comparison.

## gemini [PASS]

### reasons/api.py:review_justifications
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly introduces the optional `parallel` parameter (defaulting to 0) to the function signature and forwards it to the underlying `_review` call. This maintains clean backwards compatibility with existing API callers while enabling concurrent execution.

### reasons/cli.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Successfully exposes `--parallel` via `argparse` on the `review-justifications` subparser, specifying `type=int` with a default of `0`. This argument is cleanly fetched from `args` and passed directly to the `api.review_justifications` call within `cmd_review_justifications`.

### reasons/review_justifications.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Modularizes single batch processing into a dedicated `_process_batch` helper. In `review_justifications`, it sets up a `ThreadPoolExecutor` and uses `as_completed` when `parallel > 0`. 
- Since LLM calls are highly I/O bound (network/subprocess calls), this is the optimal concurrency mechanism.
- Accumulating batch results and invoking the `on_batch` callback inside the main thread (during future resolution) guarantees thread-safe lists operations and prevents callback race conditions.
- Cost-tracking logic in `reasons/llm.py` is safely wrapped with `_cost_lock` locks, preventing concurrent logging races.
- The design matches existing concurrent patterns in the codebase (such as `review_premises.py`).

### tests/test_review_justifications.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Introduces `test_parallel`, which sets a batch size of 1 over 2 distinct derived beliefs to generate exactly 2 batches. The test mocks `invoke_model` to verify that each concurrent worker receives the correct batch prompt context and successfully aggregates the results, making the assertion robust and deterministic.

### Self-Review
**Limitations:** - Shell tool execution was not available in this workspace context, so tests were reviewed statically/heuristically without direct execution, though the static validity of the code changes is extremely clear.

PR #235, opened 2026-07-09
Add designated supporting justification tracking

### LINKED ISSUE #227: Track designated supporting justification per Doyle's TMS
## Problem

When a node has multiple valid justifications, ftl-reasons treats them all equally -- `_compute_truth` returns IN if any justification is valid, with no preference or designation. This diverges from Doyle's 1979 TMS, which picks exactly one as the **designated supporting justification** that currently accounts for the node being IN.

This causes several issues:

### 1. Ambiguous explain output

`explain` can legitimately cite any valid justification, including one the maintainer has already declared wrong (as happened with `propagation-is-crash-free`, which carried a wrong justification alongside the correct replacement until manually removed).

### 2. Inconsistent depth

Depth is computed in the wiki generator as max over all antecedents across all justifications. If a node has a depth-1 justification (all premises) and a depth-3 justification (derived antecedents), depth shows 3 -- but if the system were tracking designated support, depth would reflect the actual support chain, not the worst case.

### 3. Retraction over-propagation

`_propagate` re-evaluates all dependents regardless of which justification they are actually using. With designated support, only nodes whose designated justification is affected need re-evaluation.

## Doyle's approach

In the 1979 TMS, each IN node has a single designated supporting justification. When that justification becomes invalid, the system attempts to find an alternative before marking the node OUT. The designated support creates a well-defined dependency tree that makes explain, depth, and retraction all unambiguous.

## Proposal

Add a `supporting_justification` field to nodes that tracks which justification (by index) is currently designated as the active support. Update:

- `_compute_truth` -- set the designated justification when a node goes IN
- `explain` -- follow the designated support chain
- Wiki generator -- show which justification is the active support, compute depth from it
- Optionally: `_propagate` could use the designation for more targeted re-evaluation

## Wiki rendering

When a belief page shows multiple justifications, mark the designated one (e.g. with a tag or visual indicator) so readers know which argument the system is actually relying on.

## Context

Discovered when `propagation-is-crash-free` had two nearly identical justifications differing only in the third antecedent -- one was wrong but still valid and equally weighted. The repair added a new justification without removing the old one, and without designated support there was no way to tell which one the system was using.

benthomasson:
## Summary

- Adds `supporting_justification` field to `Node` that tracks which justification (by index) is the active support for each IN node
- `_compute_truth` now iterates newest-first, so the most recently added valid justification is preferred
- `explain` follows the designated justification instead of re-scanning all justifications
- Depth computation uses only the designated justification's antecedents
- Wiki generator shows "Depends on (active)" vs "Depends on (other)" when multiple justifications exist
- SQLite schema updated with backward-compatible migration

Closes #227

## Test plan

- [x] All 1564 existing tests pass
- [ ] Create a node with two justifications, verify `supporting_justification` points to the newest valid one
- [ ] Defeat the newest justification, verify fallback to the older one
- [ ] Verify `explain` follows the designated justification
- [ ] Save and load, verify the field round-trips through SQLite
- [ ] Regenerate wiki, verify designated justification is marked

### DIFF
--- a/reasons/__init__.py
+++ b/reasons/__init__.py
@@ -33,6 +33,7 @@ class Node:
     text: str
     truth_value: str = "IN"  # IN or OUT
     justifications: list[Justification] = field(default_factory=list)
+    supporting_justification: int | None = None  # index into justifications list
     dependents: set[str] = field(default_factory=set)  # reverse index
     source: str = ""
     source_url: str = ""
--- a/reasons/api.py
+++ b/reasons/api.py
@@ -635,6 +635,7 @@ def show_node(node_id: str, visible_to: list[str] | None = None, db_path: str = 
             "id": node.id,
             "text": node.text,
             "truth_value": node.truth_value,
+            "supporting_justification": node.supporting_justification,
             "source": node.source,
             "source_url": node.source_url,
             "source_hash": node.source_hash,
@@ -1235,6 +1236,7 @@ def export_network(visible_to: list[str] | None = None, db_path: str = DEFAULT_D
             nid: {
                 "text": n.text,
                 "truth_value": n.truth_value,
+                "supporting_justification": n.supporting_justification,
                 "justifications": [
                     {"type": j.type, "antecedents": j.antecedents, "outlist": j.outlist, "label": j.label}
                     for j in n.justifications
@@ -2528,8 +2530,13 @@ def _node_depth(nid, net, memo=None):
         memo[nid] = 0
         return 0
     memo[nid] = 0  # cycle guard
+    si = node.supporting_justification
+    if si is not None and 0 <= si < len(node.justifications):
+        justifications = [node.justifications[si]]
+    else:
+        justifications = node.justifications
     max_d = 0
-    for j in node.justifications:
+    for j in justifications:
         for a in j.antecedents:
             max_d = max(max_d, _node_depth(a, net, memo))
     memo[nid] = max_d + 1
--- a/reasons/build_wiki.py
+++ b/reasons/build_wiki.py
@@ -103,10 +103,23 @@ def _format_node(node_id, node_detail, node_to_page, all_details=None):
     lines.append(node_detail["text"])
     lines.append("")
 
-    antecedents = set()
-    for j in node_detail.get("justifications", []):
-        for a in j.get("antecedents", []):
-            antecedents.add(a)
+    justifications = node_detail.get("justifications", [])
+    si = node_detail.get("supporting_justification")
+    if si is not None and 0 <= si < len(justifications):
+        designated = justifications[si]
+        antecedents = set(designated.get("antecedents", []))
+        other_antecedents = set()
+        for idx, j in enumerate(justifications):
+            if idx != si:
+                for a in j.get("antecedents", []):
+                    if a not in antecedents:
+                        other_antecedents.add(a)
+    else:
+        antecedents = set()
+        other_antecedents = set()
+        for j in justifications:
+            for a in j.get("antecedents", []):
+                antecedents.add(a)
 
     if antecedents:
         links = []
@@ -116,7 +129,17 @@ def _format_node(node_id, node_detail, node_to_page, all_details=None):
                 links.append(f"[{a}]({page}#{a})")
             else:
                 links.append(a)
-        lines.append(f"**Depends on:** {', '.join(links)}")
+        label = "**Depends on (active):**" if other_antecedents else "**Depends on:**"
+        lines.append(f"{label} {', '.join(links)}")
+    if other_antecedents:
+        links = []
+        for a in sorted(other_antecedents):
+            page = node_to_page.get(a)
+            if page:
+                links.append(f"[{a}]({page}#{a})")
+            else:
+                links.append(a)
+        lines.append(f"**Depends on (other):** {', '.join(links)}")
 
     dependents = node_detail.get("dependents", [])
     if dependents:
@@ -171,10 +194,15 @@ def _format_beliefs_for_prompt(node_ids, node_details):
         lines.append(f"Status: {detail['truth_value']}")
         lines.append(f"Text: {detail['text']}")
 
-        antecedents = set()
-        for j in detail.get("justifications", []):
-            for a in j.get("antecedents", []):
-                antecedents.add(a)
+        justifications = detail.get("justifications", [])
+        si = detail.get("supporting_justification")
+        if si is not None and 0 <= si < len(justifications):
+            antecedents = set(justifications[si].get("antecedents", []))
+        else:
+            antecedents = set()
+            for j in justifications:
+                for a in j.get("antecedents", []):
+                    antecedents.add(a)
         if antecedents:
             lines.append(f"Depends on: {', '.join(sorted(antecedents))}")
 
--- a/reasons/derive.py
+++ b/reasons/derive.py
@@ -121,8 +121,15 @@ def _get_depth(node_id, nodes, derived, memo=None):
         memo[node_id] = 0
         return 0
     memo[node_id] = 0  # cycle guard
+    node_data = derived[node_id]
+    all_justifications = node_data.get("justifications", [])
+    si = node_data.get("supporting_justification")
+    if si is not None and 0 <= si < len(all_justifications):
+        justifications = [all_justifications[si]]
+    else:
+        justifications = all_justifications
     max_d = 0
-    for j in derived[node_id].get("justifications", []):
+    for j in justifications:
         for a in j.get("antecedents", []):
             max_d = max(max_d, _get_depth(a, nodes, derived, memo))
     memo[node_id] = max_d + 1
--- a/reasons/export_card.py
+++ b/reasons/export_card.py
@@ -14,8 +14,13 @@ def _node_depth(nid, nodes, memo=None):
         memo[nid] = 0
         return 0
     memo[nid] = 0
+    si = node.supporting_justification
+    if si is not None and 0 <= si < len(node.justifications):
+        justifications = [node.justifications[si]]
+    else:
+        justifications = node.justifications
     max_d = 0
-    for j in node.justifications:
+    for j in justifications:
         for a in j.antecedents:
             max_d = max(max_d, _node_depth(a, nodes, memo))
     memo[nid] = max_d + 1
--- a/reasons/network.py
+++ b/reasons/network.py
@@ -648,6 +648,7 @@ class Network:
                     self.nodes[out_id].dependents.discard(node_id)
 
         node.justifications = []
+        node.supporting_justification = None
 
         # A premise is IN by default
         changed = []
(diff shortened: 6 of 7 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#235
**Models:** claude
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons/__init__.py:Node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Core data structure change adding `supporting_justification: int | None = None` field to Node. Type annotation is appropriate, default is sensible. This directly implements the issue's proposal to track which justification is designated.

### reasons/api.py:show_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Exposes the new field in the API response dict at line 638. Simple passthrough of the field value - correct and necessary for API consumers.

### reasons/api.py:export_network
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

Adds `supporting_justification` to JSON export at line 1239. Export is correct, but I cannot see the corresponding import_json code in this diff to verify round-trip serialization works. If import_json doesn't handle this field, exported networks won't reload correctly.

### reasons/api.py:_node_depth
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Lines 2533-2539 change depth calculation to only consider the designated justification when available, falling back to all justifications otherwise. This directly addresses the issue's complaint about "inconsistent depth" where depth showed worst-case across all justifications rather than the actual support chain. Bounds checking is correct (0 <= si < len).

### reasons/build_wiki.py:_format_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Lines 106-142 split antecedents into "active" (from designated) and "other" (from non-designated justifications), labeling them differently in wiki output. This implements the issue's proposal for wiki rendering: "mark the designated one so readers know which argument the system is actually relying on." The logic correctly extracts antecedents from the designated justification first, then collects others while excluding duplicates.

### reasons/build_wiki.py:_format_beliefs_for_prompt
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Lines 197-204 apply the same designated-justification-first logic to prompt formatting. Consistent with _format_node changes.

### reasons/derive.py:_get_depth
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Lines 124-135 mirror the api.py:_node_depth changes. Same bounds checking, same fallback logic. Maintains consistency between the two depth calculation sites.

### reasons/network.py:convert_to_premise
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Line 651 sets `node.supporting_justification = None` when clearing justifications during premise conversion. Proper cleanup - prevents dangling index into an empty justifications list.

### reasons/network.py:explain
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Lines 817-840 make explain prefer the designated supporting justification, with fallback to any valid justification (preferring newest via reversed iteration). This directly addresses the issue's "ambiguous explain output" problem. The fallback is defensive programming - _compute_truth should already ensure designated justification is valid, but if state is inconsistent, explain won't crash. Bounds checking is correct.

### reasons/network.py:_compute_truth
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

Lines 935-939 implement the core designation logic - iterate backwards through justifications (newest first), set supporting_justification index when valid one is found. The "prefer newest" policy is documented in the docstring and reasonable (newer justification likely supersedes older).

**CONCERN 1**: When a node has multiple valid justifications and one becomes invalid, _compute_truth will re-scan and potentially designate a different one. But what about code that adds/removes individual justifications outside of _compute_truth? I don't see such code in this diff, but if it exists, those sites need to update supporting_justification or set it to None. The only mutation visible is convert_to_premise, which correctly clears it.

**CONCERN 2**: The issue mentions "retraction over-propagation" - that _propagate could use designated support for more targeted re-evaluation. The proposal marks this as "optional", and it's not implemented. This means _propagate still re-evaluates all dependents regardless of which justification they're using. Not a blocker, but leaves one issue unresolved.

### reasons/storage.py:schema and migration
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Lines 22, 94-95 add `supporting_justification INTEGER DEFAULT NULL` column with proper migration for existing databases. The ALTER TABLE check (`if "supporting_justification" not in cols`) ensures idempotent migration. Default NULL is correct for old nodes.

### reasons/storage.py:save
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Lines 126-138 add supporting_justification to INSERT statement. Correct persistence of the field.

### reasons/storage.py:load
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Lines 203-260 handle backwards compatibility - check if column exists, SELECT NULL if it doesn't. This allows loading old databases, which will have supporting_justification=None until next propagation sets it. The nested conditionals (has_timestamps, has_supporting) are verbose but correct. The variable `supporting_j` is passed to Node constructor at line 258.

### Self-Review
**Limitations:** - **No test files in diff** - Cannot verify that existing tests still pass, or that new behavior (designated support, depth calculation, explain preference, wiki rendering) is tested. This is the biggest gap.
- **No import_json code visible** - export_network adds the field to JSON, but I can't verify the corresponding import handles it. If import_json is in api.py or another file, it's not in this diff.
- **Cannot see full Network class** - Can't verify whether there are other methods that mutate justifications lists and might need to update supporting_justification. The diff only shows convert_to_premise doing this correctly.
- **Cannot see _propagate implementation** - The issue mentions optional optimization to _propagate using designated support for targeted re-evaluation. Since it's not in the diff, I assume it's not implemented, but can't verify current _propagate behavior.

### Feature Requests
- Include test files that cover the changed functionality, or explicitly note "no test changes" if existing tests are sufficient
- For storage changes, include test coverage showing migration works (old DB -> add column -> load/save)
- For API changes like export_network, show the corresponding import code or note where it is
- When reviewing a feature that spans multiple files, provide a summary of which aspects are/aren't implemented (e.g., "retraction optimization: deferred")

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl-reasons#235
**Models:** claude
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### reasons/__init__.py:Node
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: ADDRESSES
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: UNTESTED (tests not in diff)
INTEGRATION: WIRED
REASONING: Adds `supporting_justification: int | None = None` field to track which justification (by index) currently supports the node being IN. This directly implements the core requirement from issue #227. The field defaults to None appropriately.

---

### reasons/api.py:show_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Exposes `supporting_justification` in the API response. Necessary for external consumers to see which justification is designated. Simple addition, no logic errors.

### reasons/api.py:export_network
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Includes `supporting_justification` in JSON export. Required for persistence and debugging. Consistent with show_node changes.

### reasons/api.py:_node_depth
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Uses designated justification for depth calculation when available, otherwise falls back to all justifications. This directly addresses issue point #2 "Inconsistent depth". Bounds checking is correct: `if si is not None and 0 <= si < len(node.justifications)`. Fallback gracefully handles missing/invalid designation.

### reasons/build_wiki.py:_format_node
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Implements wiki rendering requirement from the issue. Separates dependencies into "Depends on (active):" (from designated justification) and "Depends on (other):" (from non-designated justifications). The logic correctly identifies the designated justification's antecedents and distinguishes them from other antecedents. Clean visual indicator for users.

### reasons/build_wiki.py:_format_beliefs_for_prompt
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Uses designated justification for dependency extraction in prompt context. Consistent with _format_node. Correct bounds checking and fallback.

### reasons/derive.py:_get_depth
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Mirrors the pattern from api.py:_node_depth for consistency. Uses designated justification when available, falls back to all justifications. Correct implementation for derive context.

### reasons/export_card.py:_node_depth
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Third instance of the same _node_depth pattern. All three implementations (api.py, derive.py, export_card.py) now consistently use designated justification for depth calculation. Addresses issue #2 across all depth calculation sites.

### reasons/network.py:convert_to_premise
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly clears `supporting_justification = None` when converting a node to a premise (removing all justifications). A premise has no justifications, so no designated support makes sense.

### reasons/network.py:explain
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Implements the "follow designated support chain" requirement from issue point #3. The logic prefers the designated justification, then validates it, and falls back to searching other justifications in reverse order if the designated one is invalid. 

**Concern**: If the designated justification is invalid but the node is still IN, this suggests stale state. The fallback is defensive and allows explain to work, but doesn't update the designation or trigger re-propagation. This could mask inconsistencies where `supporting_justification` points to an invalid justification. However, since `explain` is read-only, not updating state is probably correct. A comment explaining this defensive behavior would help future maintainers.

### reasons/network.py:_compute_truth
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Sets the designated justification when computing truth. Iterates backward through justifications (newest first) and designates the first valid one found. If none are valid, clears the designation and returns OUT.

**Concern**: The preference for newest justification is a reasonable heuristic but undocumented. The docstring says "A node is IN if ANY justification is valid. Prefers the newest (last-added) valid justification as the designated support." This is good, but the rationale should be documented - is this Doyle's original policy, or an implementation choice? The issue doesn't specify which justification to prefer when multiple are valid.

**Minor issue**: The docstring was updated to mention the preference, which is good. However, it's unclear if "newest" is the right term - it depends on whether justifications are always appended (making newest == last). If justifications can be inserted at arbitrary positions, the policy is really "last in the list" not "newest by timestamp".

### reasons/storage.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Adds schema migration for `supporting_justification` column. The migration logic correctly checks for column existence and adds it if missing. Save/load logic properly handles the new field with appropriate fallbacks for old schemas (selects NULL where the column doesn't exist). The schema migration is backward-compatible - old databases will get the new column added automatically, and load logic handles both old and new formats.

### Self-Review
**Limitations:** - Test files not included in diff - cannot verify that changes are covered by tests or that existing tests were updated for new field
- Cannot see full codebase to verify no other code modifies justifications list (which could invalidate the index stored in supporting_justification)
- Cannot verify if there are existing tests that might break due to the new field in serialization formats
- Cannot see CLAUDE.md or other project beliefs to check for architectural constraints
- Cannot verify the historical behavior of justification ordering (whether newest == last) without seeing add_justification implementation

### Feature Requests
- Include related test files in review context, especially when new fields are added to core data structures
- Show all callsites of modified methods to verify integration completeness (e.g., all places that serialize Node to check if they handle the new field)
- Include full method bodies for modified functions, not just diff hunks, to better understand control flow
- Flag if a schema migration is present but no corresponding test for the migration path

